PR #20, opened 2015-03-17
minecraft-server: fixing quotes so volume will create correctly

justinclayton:
Single quotes don't work for VOLUME, but double quotes do.


itzg:
Thanks, your change is the correct way. It's strange it used to work...they must have gotten stricter at some point.

itzg:
Built and pushed image ID 1446815ba012

PR #32, opened 2015-06-10
Add difficulty knob.

timvisher:
Allows setting difficulty level for the server via an environment variable.


itzg:
Oh yeah, that's a great knob to have. Thanks.

Do you mind also adding a little snippet in README?

timvisher:
There you go. :)

itzg:
Sorry, I hadn't noticed before -- how about a default of `normal` or `easy`? It looks like the stock default is easy, which I think is the same for the client.

itzg:
Sorry for the delay on merging. It's built and pushed to Docker hub -- image ID 2bca89e353c5.

PR #34, opened 2015-06-13
Fix for new Forge URLs from 1.7.10-10.13.2.1300

nichivo:
Forge didn't work for me on the latest master - this seems to fix it!

The download URL format changed only for 1.7.10 (not 1.8) for some reason?!? From 1.7.10-10.13.2.1300 the URL is:

$norm-$FORGE_VERSION-$norm
e.g. 1.7.10-10.13.2.1300-1.7.10

instead of:

$norm-$FORGE_VERSION
1.7.10-10.13.2.1300


itzg:
Awesome! Thanks for also updating the docs for the new `FORGEVERSION`.

itzg:
Image 6dd16d8099b1 built and pushed on Hub.

PR #43, opened 2015-10-09
Consistent use of name "mc" everywhere

everett-toews:


itzg:
Thanks for your help.

itzg:
...built and published to Hub.

everett-toews:
:+1: 

Thanks for the top notch images!

PR #52, opened 2016-01-01
Add Spigot support, and non-zip URLs

### LINKED ISSUE #50: [mc] Allow given WORLD URLs to not end in .zip
@sshipway I was testing out my addition of `WORLD` to my container-deployer web app and noticed that download links from http://www.minecraftmaps.com/ are exposed as something like http://www.minecraftmaps.com/parkour-maps/aspect-parkour/download but then do a temp redirect to the actual zip's URL.

Since `wget` can follow redirects, perhaps we should let it download and then verify the final the filename/content.

sshipway:
- Allow world URLs to end in something other than .zip (fixes #50 )
- Some whitespace fixes (tabs->4space)
- Allow TYPE=SPIGOT and TYPE=BUKKIT.  Note that only version 1.8.8 is currently available.
- Rename directories for dimensions if TYPE=SPIGOT/BUKKIT
- Some text changes to reflect new features
- Allow type options to be lower-case
- Add documentation
- Add /plugins mount for external bukkit plugins (like /mods for forge mods)

### DIFF
--- a/minecraft-server/start-minecraft.sh
+++ b/minecraft-server/start-minecraft.sh
@@ -37,6 +37,26 @@ echo "Checking type information."
 case "$TYPE" in
   *BUKKIT|*bukkit|SPIGOT|spigot)
     TYPE=SPIGOT
+    case "$TYPE" in
+      *BUKKIT|*bukkit)
+        echo "Downloading latest CraftBukkit 1.8 server ..."
+        SERVER=craftbukkit_server.jar
+        ;;
+      *)
+        echo "Downloading latest Spigot 1.8 server ..."
+        SERVER=spigot_server.jar
+        ;;
+    esac
+    case $VANILLA_VERSION in
+      1.8*)
+        URL=/spigot18/$SERVER
+        ;;
+      *)
+        echo "That version of $SERVER is not available."
+        exit 1
+      ;;
+    esac
+    wget -q https://getspigot.org$URL
     ;;
 
   FORGE|forge)
--- a/minecraft-server/start-minecraft.sh
+++ b/minecraft-server/start-minecraft.sh
@@ -33,19 +33,15 @@ esac
 
 cd /data
 
-echo "Checking minecraft / forge type information."
+echo "Checking type information."
 case "$TYPE" in
-  VANILLA)
-    SERVER="minecraft_server.$VANILLA_VERSION.jar"
+  *BUKKIT|*bukkit|SPIGOT|spigot)
+    TYPE=SPIGOT
+    ;;
 
-    if [ ! -e $SERVER ]; then
-      echo "Downloading $SERVER ..."
-      wget -q https://s3.amazonaws.com/Minecraft.Download/versions/$VANILLA_VERSION/$SERVER
-    fi
-  ;;
-
-  FORGE)
+  FORGE|forge)
     # norm := the official Minecraft version as Forge is tracking it. dropped the third part starting with 1.8
+    TYPE=FORGE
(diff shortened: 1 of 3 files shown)


minecraft-server/start-minecraft.sh
```diff
@@ -33,19 +33,35 @@ esac
...
-  VANILLA)
-    SERVER="minecraft_server.$VANILLA_VERSION.jar"
-
-    if [ ! -e $SERVER ]; then
-      echo "Downloading $SERVER ..."
-      wget -q https://s3.amazonaws.com/Minecraft.Download/versions/$VANILLA_VERSION/$SERVER
-    fi
-  ;;
+  *BUKKIT|*bukkit|SPIGOT|spigot)
+    TYPE=SPIGOT
```
itzg: It seems like this line would effectively coerce BUKKIT type into SPIGOT.

minecraft-server/start-minecraft.sh
```diff
@@ -33,19 +33,35 @@ esac
...
+    case $VANILLA_VERSION in
+      1.8*)
+        URL=/spigot18/$SERVER
+        ;;
+      *)
+        echo "That version of $SERVER is not available."
+        exit 1
+      ;;
+    esac
+    wget -q https://getspigot.org$URL
```
itzg: Awesome! I had only come across their build-it-yourself instructions before. I'm glad you knew about them having a pre-built jar available.

minecraft-server/start-minecraft.sh
sshipway: Actually, Bukkit is currently only available as a 100% compatible compile from the Spigot people (see how $SERVER is changed) due to DMCA issues.  The SPigot API is an extension of the Bukkit one.

sshipway: Both come from the same place and require the same modifications to the world structure, so are handled together.

sshipway:
I'm using the TYPE=SPIGOT to later ensure the correct handling of plugins and world structure; TYPE=BUKKIT is handled here to download the correct craftbukkit-server.jar file and it is set to TYPE=SPIGOT to ensure the correct handling later.

minecraft-server/start-minecraft.sh
itzg: Got it. I remember there was controversy with Bukkit...and then I hadn't researched much of Spigot. Thanks.

minecraft-server/start-minecraft.sh
sshipway: The only pity is that there is no simple way to get Version $X or a list of available versions, though you can get Latest 1.8.  As it is, only v1.8, 1.8.3 and 1.8.8 are available via the Spigot Jenkins and are all compatible.

minecraft-server/start-minecraft.sh
```diff
@@ -108,6 +139,12 @@ case "X$WORLD" in
         fi
       done
     fi
+    if [ "$TYPE" = "SPIGOT" ]; then
```
sshipway: This catches BUKKIT types as well as we set TYPE=SPIGOT for both cases

minecraft-server/start-minecraft.sh
```diff
@@ -252,6 +289,11 @@ do
     cp -rf "$c" /data/config
   fi
 done
-
+if [ "$TYPE" = "SPIGOT" ]; then
```
sshipway: Again, this catches BUKKIT types as well.  We use a separate volume for plugins as they are not compatible with mods though they look like them

PR #65, opened 2016-03-15
[mc] resolved issues with downloading new server executable

jtcressy:
Using this code, wget will download and overwrite spigot_server.jar if and ONLY if it can contact the server and download a copy. otherwise, it will leave the file alone. If -O was used, wget would overwrite the file with 0 bytes, effectively deleting it. In combination with -N and double pipe, bash will skip to the next statement if the first returns false or an error.

### DIFF
--- a/minecraft-server/start-minecraft.sh
+++ b/minecraft-server/start-minecraft.sh
@@ -62,7 +62,8 @@ case "$TYPE" in
       ;;
     esac
 
-    rm -f $SERVER
+    #BAD Idea
+    #rm -f $SERVER
     wget -q https://getspigot.org$URL
 
     ;;
--- a/minecraft-server/start-minecraft.sh
+++ b/minecraft-server/start-minecraft.sh
@@ -62,9 +62,8 @@ case "$TYPE" in
       ;;
     esac
 
-    #BAD Idea
-    #rm -f $SERVER
-    wget -q https://getspigot.org$URL
+    #attempt https, and if it fails, fallback to http and download that way. Display error if neither works.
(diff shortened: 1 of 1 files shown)


itzg:
Awesome, thanks. I'll take a look at this tonight and get it merged down.

jtcressy:
It works great, built it on my separate branch and already have it running. `docker pull jtcressy/minecraft-server` gets you the image I compiled from this commit.

minecraft-server/start-minecraft.sh
```diff
@@ -62,8 +62,8 @@ case "$TYPE" in
       ;;
     esac
 
-    rm -f $SERVER
-    wget -q https://getspigot.org$URL
+    #attempt https, and if it fails, fallback to http and download that way. Display error if neither works.
+    wget -q -N spigot_server.jar https://getspigot.org$URL || (echo "Falling back to http, unable to contact server using https..." && wget -q -N spigot_server.jar http://getspigot.org$URL) || echo "Unable to download new copy of spigot server"
```
itzg: To keep the code better to maintain, can you change two things here:
* use `$SERVER` instead of "hardcoding" `spigot_server.jar`
* put a backslash and line break after each `||` and `&&` to break this long line into four such as

```
wget -q -N spigot_server.jar https://getspigot.org$URL || \
  (echo "Falling back to http, unable to contact server using https..." && \
  wget -q -N spigot_server.jar http://getspigot.org$URL) || \
  echo "Unable to download new copy of spigot server"
```

itzg:
Sorry for the delay @jtcressy ...somehow I missed the conversation notification.

I have kicked off the build on Hub...

itzg:
...and built: sha256:e1e91d0ed57c6bd12d80da63f74b99f119395c1ead68e6eabf31fe1b0ead8df0

PR #67, opened 2016-04-08
Build spigot using buildtools

brantje:
This pull request adds the following things:

- Spigot is now using buildtools to build the server jar, because of this, first time startup takes longer
- Added the following ENV variables
ENABLE_RCON
RCON_PASSWORD
ENABLE_QUERY
MAX_PLAYERS
MAX_WORLD_SIZE
Their descriptions can be found in the readme.md which i edited.

More server settings will follow


minecraft-server/start.sh
```diff
@@ -11,5 +11,4 @@ while lsof -- /start-minecraft; do
   echo -n "."
   sleep 1
 done
-echo "...switching to user 'minecraft'"
-exec sudo -E -u minecraft /start-minecraft
```
itzg: The `sudo` is needed to avoid files in the attached volume getting created as root. I'm thinking you ran into issues with `/root` since `$HOME` might have still been referring to root's. Adding an

    export HOME=/data

to the top of `start-minecraft.sh` might resolve that.

minecraft-server/start-minecraft.sh
```diff
@@ -39,34 +39,24 @@ echo "Checking type information."
...
+        find * -maxdepth 0 ! -name '*.jar' -exec rm -rf {} \; && \
+        chown minecraft:minecraft spigot-*.jar && \
+        chown minecraft:minecraft craftbukkit-*.jar && \
+        mv spigot-*.jar /data/spigot_server.jar && \
+        mv craftbukkit-*.jar /data/craftbukkit.jar
+        echo "Cleaning up"
+        rm -rf /data/temp
+        cd /data
+    fi
+    SERVER='spigot_server.jar'
```
itzg: It looks like the choice for Bukkit gets lost here.

itzg:
I really like the new environment variable options, like RCON since that'll be beneficial for my [MCCY](https://github.com/moorkop/mccy-engine) project, especially features [like this one](https://github.com/moorkop/mccy-engine/issues/28).

Overall, I'm having second thoughts about building Spigot/Bukkit during startup. Using [getspigot](http://getspigot.org/) still seems to be the best of both worlds. They're building as recommended and they're pre-building...so the only container startup cost is downloading the appropriate version.

So...I hate to ask this, but can you revert the build related changes or if it's easier submit a new PR with just the env variable type changes?

itzg:
Oh, since the [RCON protocol specifies 27015 as the default](https://developer.valvesoftware.com/wiki/Source_RCON_Protocol#Using_Source_RCON), can you add an `EXPOSE` to the `Dockerfile` for that?

brantje:
> Overall, I'm having second thoughts about building Spigot/Bukkit during startup. Using getspigot still seems to be the best of both worlds. They're building as recommended and they're pre-building...so the only container startup cost is downloading the appropriate version.

False, currently their builds crashes when you have boots with frost walker.
I assume it will be fixed, but i needed my server, so building from scratch seemed the best option.
Maybe add an BUILD_FROM_SOURCE flag?

And i took the default server settings from the http://minecraft.gamepedia.com/Server.properties page.

itzg:
Ah, providing an option to build or not is a very interesting idea. Can you proceed with that?

As for Bukkit vs Spigot, it's hard to find a definitive answer especially since I don't personally use either that much. They are definitely plugin compatible, but can't tell if they still have differences with commands and config.

brantje:
Afaik there are no difference.
Also, CraftBukkit stopped the development (see here)[https://bukkit.org/threads/is-bukkit-still-going.401513/].

Will add the remaining server properties, and a build from source option.

itzg:
Sorry it's taken me a while to get back to this. I'm looking at the newest round of changes now.

minecraft-server/README.md
```diff
@@ -267,28 +269,111 @@ downloaded, scaled, and converted from any other image format:
     docker run -d -e ICON=http://..../some/image.png ...
 
 ### Rcon
+
 To use rcon use the `ENABLE_RCON` and `RCON_PASSORD` variables.
+By default rcon port will be `25575` but can easily be changed with the `RCON_PORT` variable.
```
itzg: I wonder if it's better to recommend they do a port mapping rather adjust the config...but then you already have the code for that so doesn't need to change.

minecraft-server/README.md
```diff
@@ -267,28 +269,111 @@ downloaded, scaled, and converted from any other image format:
...
 
     docker run -d -e ENABLE_RCON=true -e RCON_PASSWORD=testing
     
 ### Query
+
 Enabling this will enable the gamespy query protocol.
-By default the query port will be `25565` (UDP) but can easily be chnged with the `QUERY_PORT` variable.
+By default the query port will be `25565` (UDP) but can easily be changed with the `QUERY_PORT` variable.
```
itzg: Similar comment as above.

minecraft-server/README.md
```diff
@@ -267,28 +269,111 @@ downloaded, scaled, and converted from any other image format:
...
+
+Allows server to announce when a player gets an achievement.
+
+    docker run -d -e ANNOUNCE_PLAYER_ACHIEVEMENTS=true   
+
+### Enable  Command Block
+
+Enables command blocks
+
+     docker run -d -e ENABLE_COMMAND_BLOCK=true
```
itzg: Yay!

minecraft-server/start-minecraft.sh
```diff
@@ -187,6 +225,68 @@ if [ ! -e server.properties ]; then
     sed -i "/motd\s*=/ c motd=$MOTD" /data/server.properties
   fi  
 
+  if [ -n "$ALLOW_NETHER" ]; then
```
itzg: Wow, you knocked out a whole bunch of options here. A helper shell function would be good to avoid the repetition, but I can refactor that later if you don't want to mess with that.

itzg:
Hmm, GitHub thinks I commented on outdated diffs...I'll try this a different way.

minecraft-server/start-minecraft.sh
```diff
@@ -178,31 +211,135 @@ esac
...
+  echo "Creating server.properties"
   cp /tmp/server.properties .
 
   if [ -n "$WHITELIST" ]; then
+    echo "Creating whitelist"
     sed -i "/whitelist\s*=/ c whitelist=true" /data/server.properties
     sed -i "/white-list\s*=/ c white-list=true" /data/server.properties
   fi
 
   if [ -n "$MOTD" ]; then
```
itzg: It looks like we need a helper function now for all the repeated sed blocks. I can refactor that later if you didn't want to do that.

brantje: Yes please =), i don't have mucht experience with bash.

itzg:
Sorry again for the delay and thanks so much for the contribution.

I'll follow up later on and refactor that properties-set bits.

itzg:
...and pushed to hub at sha256:33b498a59d496369f2949bd843957e19f9224c5dd4178d376e9c4b46bb1582ca

PR #90, opened 2016-07-07
Update start-minecraft.sh

larrywright:
Each time this script is run it downloads the modpack if one is specified. When it tries to unzip the file, if the files already exist in the plugins directory, the script will hang indefinitely because unzip asks for confirmation that you want to overwrite the existing files. Adding -o to the unzip command will tell it to overwrite the files. I suspect the world download has the same issue, but I don't use that and so can't really test it.


itzg:
What's "funny" is that I just now ran into this same problem, so thanks for the timely solution :)

itzg:
Pushed to Hub at sha256:43cfa8bacc2b2ad07380aab41d4a0a979985cd9a2f412099404fcaf03d5f830d

PR #94, opened 2016-07-10
fixed env inheritance issue, re-ordered OPTS

jaychris:
- modified the su CMD so that root's environment, which includes passed container variables, is passed to the elasticsearch user.
- modified the ordering of OPTS a bit so that --path.plugins is always last


itzg:
Thanks for your help @jaychris . If you hadn't already noticed, the update was pushed at sha256:99f265f5a774e3120d1a311e1bb42983137a180b283ce0a90a6b379eb04caa8f

jaychris:
no problem.  I did notice.  Have pulled the latest and verified that both issues are resolved.  Thanks for the quick merge!

PR #101, opened 2016-08-15
minecraft-server: EULA var needs to be a string (docker-compose.yml)

parthipanramesh:
docker-compose refuses to build a container without the quotation marks.


itzg:
Thanks for catching and fixing that...it used to work, but apparently broke somewhere along the way.

PR #106, opened 2016-09-22
Added CONSOLE envvar to control console output

kstrauser:
By default, CONSOLE is true. When set to false, start-minecraft.sh
adds "--noconsole" to java. Use this to keep Spigot from burning CPU.


kstrauser:
CI seems to be wonky. I'll revisit this tonight.

itzg:
I'll look later too since I bet its one of the other images (usually Cassandra) that's simply failing due to an old download.

kstrauser:
Cool. I'd forgotten to update both branches of the with/without bootstrap code anyway.

itzg:
OK, if you `git pull` from master branch of https://github.com/itzg/dockerfiles then I have pushed the Cassandra fix that's failing the build. With that you can re-push and the build should go green.

If that becomes a hassle, then I'm cool with merging as it is since the changes look totally fine to me.

itzg:
FYI, here's the clean build of master:

https://circleci.com/gh/itzg/dockerfiles/167

kstrauser:
I've rebased off your master and force pushed. Crossing my fingers...

kstrauser:
\o/

PR #149, opened 2017-05-11
Fix comment

florian-besser:


elasticsearch/start
```diff
@@ -149,7 +149,7 @@ mkdir -p /conf/scripts
 echo "Starting Elasticsearch with the options $OPTS"
 CMD="$ES_HOME/bin/elasticsearch $OPTS"
 if [ `id -u` = 0 ]; then
-  echo "Running as non-root..."
+  echo "Running as root..."
```
itzg: I had to stop and think about this one, but I'm pretty sure the original comment is correct. Since the `DEFAULT_ES_USER` is "elasticsearch" [here](https://github.com/itzg/dockerfiles/blob/master/elasticsearch/Dockerfile#L19) then this block is switching from the container's root user to "elasticsearch".

florian-besser:
Hm, we are both not right it seems:

This seems OK so far:

docker run -d -p 9200:9200 -p 9300:9300 itzg/elasticsearch
Finding IPs. found! 172.17.0.3
Starting Elasticsearch with the options    -E path.conf=/conf   -E path.data=/data   -E path.logs=/data   -E transport.tcp.port=9300   -E http.port=9200 -E network.host=172.17.0.3
Running as non-root...

Lets run this as root:

docker run -u root itzg/elasticsearch
Finding IPs. found! 172.17.0.3
Starting Elasticsearch with the options    -E path.conf=/conf   -E path.data=/data   -E path.logs=/data   -E transport.tcp.port=9300   -E http.port=9200 -E network.host=172.17.0.3
Running as non-root...

Seems the idea to check for UID is not quite what I thought it was. Maybe I'll come up with a better PR.

itzg:
@florian-besser thanks for your continued thoughts on this. UIDs (rather than usernames) seem to be the only universal truth spanning the outer host and inner container, which was why I checked for that. I'm hoping that datapoint helps to narrow improvements you want to make there.

PR #150, opened 2017-05-12
Use cURL insecurely when downloading

jackwilsdon:
This fixes an error caused by the expired certificate on http://ci.mcadmin.net.

I know it's a sub-optimal solution but I can't see another solution until http://ci.mcadmin.net sorts out their certificate.

### DIFF
--- a/minecraft-server/start-minecraft.sh
+++ b/minecraft-server/start-minecraft.sh
@@ -75,7 +75,7 @@ function downloadSpigot {
   fi
 
   echo "Downloading $match"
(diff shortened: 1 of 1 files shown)


itzg:
(I guess I didn't click to save my comment.) Thanks for the change and I agree that it's sub-optimal, but necessary.

Your change is now pushed to Hub 
https://hub.docker.com/r/itzg/minecraft-server/builds/bg2qk9bdvfyfgynww22qxzr/

Issue #139, opened 2017-04-06
Biome value validation needs to be relaxed for modded biome 


itzg:
From Hub comments:
> stelk
13 days ago
I am playing a FTB modpack and want to enable BIOMES O PLENTY. Normally I would update the LEVEL_TYPE to BIOMESOP. However, this does not pass your validation code and throws an error. Is there a way to enable this mod using your docker script?

PR #156, opened 2017-06-05
minecraft-server: Fixed FORGEVERSION case, add support for BIOMESOP

dirkcjelli:
First of all: Thanks for your minecraft-server image, I really like it 👍 

* Fixed FORGEVERSION in case statement to working FORGE_VERSION
   * Removed default case because `FORGE_VERSION=$FORGE_VERSION` doesn't make a lot of sense if nothing needs to be changed.
* Add support for BIOMESOP level-type

### DIFF
--- a/minecraft-server/start-minecraft.sh
+++ b/minecraft-server/start-minecraft.sh
@@ -454,7 +454,7 @@ if [ ! -e server.properties ]; then
     echo "Setting level type to $LEVEL_TYPE"
     # check for valid values and only then set
     case $LEVEL_TYPE in
-      DEFAULT|FLAT|LARGEBIOMES|AMPLIFIED|CUSTOMIZED)
+      DEFAULT|FLAT|LARGEBIOMES|AMPLIFIED|CUSTOMIZED|BIOMESOP)
         sed -i "/level-type\s*=/ c level-type=$LEVEL_TYPE" /data/server.properties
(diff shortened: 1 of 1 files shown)


itzg:
Thanks! That `FORGEVERSION` has been bugging me lately since I've been in and around that code. The only challenge in changing that is that [the docs](https://github.com/itzg/dockerfiles/blob/master/minecraft-server/README.md#running-a-forge-server) reference `FORGEVERSION`. As a migration, can you make a few more changes:

1. Update the README to reference `FORGE_VERSION`
2. Update the default env var name [here](https://github.com/itzg/dockerfiles/blob/master/minecraft-server/Dockerfile#L48)
3. Above the case block, for anyone using the old variable, put a `if [[ -n $FORGEVERSION ]]; then FORGE_VERSION=$FORGEVERSION; fi` for backfilling old into current

itzg:
I just noticed that this will fix #139 , so if you don't mind I'm going to revert just the `FORGEVERSION` change. That way I can at least merge part of your change for now.

dirkcjelli:
Hi,

sorry for the delay, I'll be able to have a look at this today evening. :)

itzg:
I totally understand. Since I went ahead and merged the level type change, I hope you don't mind opening a new PR for the remainder. Thanks.

PR #166, opened 2017-06-29
Fix typos in config file generation

macman31:
Env variable `SPAWN_ANIMALS` was misspelled as `SPAWN_ANIMAILS`, thus rendering the modification of this option in the config file from env variables non effective if the documentation is correctly followed.
Also, `spawn-npcs` was repeated two times.
Also, `force-gamemode` was never taken into account because of a regex issue.


itzg:
...and [built on Hub](https://hub.docker.com/r/itzg/minecraft-server/builds/bioza5nxeouzws32yp96vut/)

PR #167, opened 2017-07-01
Added option REMOVE_OLD_MODS

lukascernydis:
When the option `MODPACK` is specifed you can also instruct script to delete old mods/plugins prior to installing new ones.
This behaviour is desirable in case you want to upgrade or replace mods/plugins from downloaded zip file.


itzg:
And built
https://hub.docker.com/r/itzg/minecraft-server/builds/bt7htzzgygprfyzrvbajj3f/

PR #169, opened 2017-07-12
Update Dockerfile

manojsb:
Updated to v5.5.0


manojsb:
Its okay. We all are on same goal.

PR #173, opened 2017-07-15
Set UID for ES_DEFAULT_USER to 1100, avoid conflicts with debian host…

opHASnoNAME:
When you mount a volume from your host system on a Debian host, you can face conflicts with systemd users (uid 100,101..).

### DIFF
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -26,7 +26,7 @@ ENV ES_HOME=/usr/share/elasticsearch-$ES_VERSION \
     DEFAULT_ES_USER_UID=$DEFAULT_ES_USER_UID \
     ES_JAVA_OPTS="-Xms1g -Xmx1g"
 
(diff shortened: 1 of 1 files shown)


elasticsearch/Dockerfile
```diff
@@ -21,6 +23,7 @@ HEALTHCHECK --timeout=5s CMD wget -q -O - http://$HOSTNAME:9200/_cat/health
 
 ENV ES_HOME=/usr/share/elasticsearch-$ES_VERSION \
     DEFAULT_ES_USER=elasticsearch \
+    DEFAULT_ES_USER_UID=$DEFAULT_ES_USER_UID \
     ES_JAVA_OPTS="-Xms1g -Xmx1g"
 
 RUN adduser -S -s /bin/sh $DEFAULT_ES_USER
```
itzg: It looks like this line needs a `-u $DEFAULT_ES_USER_UID` to actually created with 1100

itzg:
This change has now been pushed to Hub.

PR #183, opened 2017-10-29
Code refactorization

s0rcerer88:
Just refactor work.

I'm now working on Sponge integration.


itzg:
@s0rcerer88 I just wanted to let you know that I started reviewing this. So far it's looking good. 

At some point I'm thinking it'll be handy to make use of [the wiki](https://github.com/itzg/dockerfiles/wiki) and capture a "design document" with the script-to-script flow.

s0rcerer88:
Ok. No matter. I will start working on it. I'm working for now on Sponge Integration, but I think that I can use some time for documentation. I've seen that Wiki is empty, so if you want to prepare a page for the documentation, I can complete it.

FYI, I've tested deployment of every single type before commiting, so all should work as before refactor (No issues correction done).

itzg:
Building on what you suggested, I'll create that wiki page and document the script flow while I'm reviewing the PR :)

So, you keep on moving as you were. Thanks for all your help and attention to detail on this.

minecraft-server/start-configuration.sh
```diff
@@ -0,0 +1,82 @@
...
+  FORGE|forge)
+    su-exec minecraft /start-deployForge $@
+  ;;
+
+  FTB|ftb)
+    su-exec minecraft /start-deployFTB $@
+  ;;
+
+  VANILLA|vanilla)
+    su-exec minecraft /start-deployVanilla $@
```
itzg: When I tried this out it doesn't seem to like this "nested" call to `su-exec`. It seems like the `su-exec` at the end of `start` is the only one that would be needed anyway. Here is what I got:

```
> docker run -it --rm -e EULA=TRUE mc
Switching to user 'minecraft'
Checking version information.
Checking type information.
su-exec: setgroups: Operation not permitted
```

minecraft-server/Dockerfile
```diff
@@ -32,7 +32,17 @@ EXPOSE 25565 25575
 ADD https://github.com/itzg/restify/releases/download/1.0.4/restify_linux_amd64 /usr/local/bin/restify
 ADD https://github.com/itzg/rcon-cli/releases/download/1.3/rcon-cli_linux_amd64 /usr/local/bin/rcon-cli
 COPY start.sh /start
-COPY start-minecraft.sh /start-minecraft
+COPY start-configuration.sh /start-configuration
```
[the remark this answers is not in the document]
s0rcerer88: Done, but I've used "COPY start* /" instead due to the first file.

minecraft-server/start-configuration.sh
s0rcerer88: I've started modifying it. I'm changing 'su-exec' commands with 'exec' comands instead.

s0rcerer88: Also done. Changes commited

s0rcerer88:
I'm ready to PR changes for sponge integration (Vanilla side only at this moment).
Ready on my own fork if you need to check it. I'm going to wait to PR till merge is done to avoid any kind of problem.

PR #190, opened 2018-01-27
making use of all the server properties magic in ftb packages ..

scones:
cheers!


itzg:
...and it's pushed to hub

https://hub.docker.com/r/itzg/minecraft-server/builds/btxngfbgp3ru7zm3zrqjutb/

PR #197, opened 2018-02-12
Fixes #196 makes it possible to add server port as a env.

### LINKED ISSUE #196: Set minecraft server port from environment variable
Is it possible to set the server port from a -e variable?

Thanks in advance.

TheLeftMoose:
I hope that this is the only place that has to be altered in order to pass server port and change it in the server.properties.


itzg:
What scenario did you encounter that needed this? If it is indeed needed you could established the default value along with the other environment variables:

https://github.com/itzg/dockerfiles/blob/master/minecraft-server/Dockerfile#L44-L49

TheLeftMoose:
The problem is when you have a bungeecord proxy setup. Here each minecraft server has to have awareness about the actual port it is running on (as i understand)... So building a docker-compose file with multible minecraft servers, on the same host... then it is a pain to start them, and then configure server.properties afterwards.
I'll change the default env's as well. didn't notice that.

PR #216, opened 2018-05-19
[mc] Add snooper enabled toggle

pcdevil:
Hello there,
I started to use the minecraft server docker file but one of the server properties I use is missing. With this modifications it can be configured via environment variable.

Please review it!

### DIFF
--- a/minecraft-server/README.md
+++ b/minecraft-server/README.md
@@ -553,7 +553,7 @@ If set to true, players will be set to spectator mode if they die.
 
 ### Snooper
 
-If set to true, server will send snoop data to the minecraft sever.
+If set to false, the server will not send data to snoop.minecraft.net server.
 
     docker run -d -e SNOOPER_ENABLED=false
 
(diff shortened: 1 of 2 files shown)


minecraft-server/README.md
```diff
@@ -551,6 +551,12 @@ If set to true, players will be set to spectator mode if they die.
 
     docker run -d -e HARDCORE=false
 
+### Snooper
+
+If set to true, server will send snoop data to the minecraft sever.
```
itzg: Since minecraft itself defaults this to true, perhaps the wording should be flipped. It would also better match your example. So, how about something like

> If set to false, the server will not send data to snoop.minecraft.net

pcdevil: Very good point! Modified in the following commit!

itzg:
Your change is now pushed to Hub.

pcdevil:
thank you very much!

PR #217, opened 2018-06-02
Allow existing world volume mount for FTB

Lemmons:
The core of this pr allows a user to backup and restore only their world save rather than the whole minecraft data folder.


Lemmons:
No problem.  Thanks for making the excellent image in the first place!

altano:
@Lemmons how would one go about using this? :)

itzg:
@altano I haven't used the exact combination that @Lemmons addressed with the PR, but it looks like you would be able to attach volumes at the respective world save directories under the container's `/data/FeedTheBeast` directory. This PR ensures that applying the FTB mod pack won't overwrite the world saves.

PR #223, opened 2018-07-05
Update MC Dockerfile

SConaway:


minecraft-server/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN apk add -U \
 
 RUN pip install mcstatus
 
-HEALTHCHECK CMD mcstatus localhost ping
+HEALTHCHECK CMD mcstatus localhost:$SERVER_PORT ping
```
itzg: Does Docker delay the variable substitution here until the actual invocation?

SConaway:
I think so, but I'll try it later today.

itzg:
I was curious and was going to test it out later also :) .

SConaway:
I just built this and it seems to work. I'm making a image to control a server also and using yours as guidance. I have one question though: How does `mc-server-runner` work? There's no readme, but it seems like something useful.

PR #235, opened 2018-07-24
Fix typos in config file generation for max-tick-time setting.

sumomoneko:
Env variable `MAX_TICK_TIME` didn't work because of this typos in property name.


sumomoneko:
Thanks for merging!

PR #243, opened 2018-08-24
unzip fails if a file already exists in the zip file.

mattiaskagstrom:
Unzipping world
replace banned-ips.json? [y]es, [n]o, [A]ll, [N]one, [r]ename:

unzip: can't read standard input


itzg:
This is now pushed to Hub.

PR #249, opened 2018-09-18
mc: run FTB installer from correct location

duaneg:
We have pushd'ed into the directory so do not prefix the shell script
with the directory name.


duaneg:
I was having problems starting up a new server -- this seems to fix it. Thanks for your work on this docker image!

itzg:
This is now built and available on Docker Hub.

PR #253, opened 2018-09-27
mc: set umask to 0002 and chmod g+w the /data folder

proudlytm:
Do this to allow write access for groups that the minecraft user is in.

This commit also fixes an issue if a group is specified with $GID and it has
no write access in the mounted host volume.

### DIFF
--- a/minecraft-server/start
+++ b/minecraft-server/start
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+umask 0002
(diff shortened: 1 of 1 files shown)


itzg:
This is now available on Hub.

proudlytm:
No problem! Credits to you for actually giving the solution to the problem!

PR #255, opened 2018-10-07
only download world if there is no preexisting world

mattiaskagstrom:
My world just got overwritten when i re-created the container. this should hopefully solve it..


mattiaskagstrom:
it's all fine, i have off-site backups :)

probably my fault tho, i made the edit to the dockerfile that the unzip should overwrite, after i had problem with the unzip failing if the zip file had a settings file etc.

itzg:
Ah :). This is now available on Docker Hub.

PR #268, opened 2018-12-14
feat(minecraft-server): update spigot plugins using /plugins if newer

### LINKED ISSUE #218: How to create ready to start image including plugins and configs
I am trying to continously build a docker image on our jenkins packaging the latest builds of our custom plugins and provided plugins like WorldEdit. I also want to download the latest versions of our Plugin configs where every Plugin Config has its own git repository.

I then want to package all that into a single docker image that I can deploy and start on my server or localhost for development.

Any help or hint in the right direction would be very appreciated :).

Silthus:
Deep copy any *.jar files in /plugins and update the corresponding files in the /data/plugins dir.

The goal of this PR is to allow updating of plugins through child Docker images.

If wanted this can also be implemented for the /mods dir. May workaround implementing #267 and  #218

### DIFF
--- a/minecraft-server/start-minecraftFinalSetup
+++ b/minecraft-server/start-minecraftFinalSetup
@@ -36,13 +36,11 @@ done
 
 # If any modules have been provided, copy them over
 mkdir -p /data/mods
-for m in /mods/*.{jar,zip}
-do
-  if [ -f "$m" -a ! -f "/data/mods/$m" ]; then
-    echo Copying mod `basename "$m"`
-    cp "$m" /data/mods
-  fi
-done
+if [ -d /mods ]; then
+  echo "Copying any mods over..."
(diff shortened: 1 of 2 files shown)


minecraft-server/start-minecraftFinalSetup
```diff
@@ -54,8 +54,12 @@ done
 
 if [ "$TYPE" = "SPIGOT" ]; then
   if [ -d /plugins ]; then
+    # Copy all of the plugins directory first to include initial configs
     echo Copying any Bukkit plugins over
     cp -r /plugins /data
```
itzg: Shouldn't this `cp` be removed then?

Silthus: Yes you are right. I added copying the mods with rsync as well.

PR #270, opened 2018-12-17
feat(minecraft-server): add version argument for buildtools

Silthus:
For future compatibility and to avoid breaking backwards compat I added a build argument with for the version of the build tools download.


itzg:
This is now published on Docker Hub.

PR #291, opened 2019-02-15
Fix for custom curseforge modpacks that do not have mods dir

dudymas:
This is sort of a combo fix.

For one, it appears some modpacks have been generated without a 'mods' dir, so running the manifest download will fail without a 'mkdir'. This appears to just be a thing that happens with the newest modpacks.

There's also an issue with the jq expression where slurping and stdin syntax differ (the dollar sign shouldn't be used).


dudymas:
I started testing more servers by deleting the whole dir except for settings and ops, and saw that mod dir was broken and the jq path was spotty. Hopefully that's the only things I'll find. But if I see anything else, I'll be sure to open another PR. Apologies I didn't see this sooner.

itzg:
This is now pushed to Hub.

PR #293, opened 2019-02-24
Updated PAPERMC CI URL and job name for older versions

LoudSoftware:
There were some small changes done to PaperMc's CI urls for older versions of minecraft.


itzg:

PR #299, opened 2019-03-23
feat(minecraft-server): replace environment variables in configs

### LINKED ISSUE #298: [Feat]: dynamic replacement of variables with environment variables
I recently finalized my ready to start docker image (#218) and came accross one more issue. I have a lot of config files that use database connection strings and such. Actually I think most minecraft server owners have at least one file that contains database informations.

Now we could include the connection string when building the image but then sensitve information would be out there in the wild. Also it is Docker best practice to use environment variables or even better secrets to provide sensitve information to docker containers.

I then had the idea of going thru all of the config files and replacing predefined variables with environment variables. Lets take the following example.

We have a `config.yml` file somewhere burried deep inside the `/data/plugins/` directory.

```yml
...
database:
    host: ${CFG_DB_HOST}
    name: ${CFG_DB_NAME}
    password: ${CFG_DB_PASSWORD}
```

All of these variables will be replaced automatically by environment variables defined in your container.

```docker-compose
version: '3'
# Other docker-compose examples in /examples

services:
  minecraft:
    image: itzg/minecraft-server
    ports:
      - "25565:25565"
    volumes:
      - "mc:/data"
    environment:
      EULA: "TRUE"
      CONSOLE: "false"
      ENABLE_RCON: "true"
      RCON_PASSWORD: "testing"
      RCON_PORT: 28016
      # enable env variable replacement
      REPLACE_ENV_VARIABLES: "TRUE"
      # define an optional prefix for your env variables you want to replace
      ENV_VARIABLE_PREFIX: "CFG_"
      # and here are the actual variables
      CFG_DB_HOST: "http://localhost:3306"
      CFG_DB_NAME: "minecraft"
      CFG_DB_PASSWORD: "ug23u3bg39o-ogADSs"
    restart: always
  rcon:
    image: itzg/rcon
    ports:
      - "4326:4326"
      - "4327:4327"
    volumes:
      - "rcon:/opt/rcon-web-admin/db"

volumes:
  mc:
  rcon:
```

After your server started all configs that contain the prefix variables wrapped inside `${...}` will be replaced with the given values.

Silthus:
Add additional setup script that replaces environment variables inside
config files.

Closes #298


Silthus:
The code and docs are ready, but I did not have a chance to fully test this with a docker image. I only tested the script itself for now. That's why it is still WIP.

Silthus:
Okay my tests were successful and you can pull this now 👍

itzg:
Cool, I’ll try review this later today.

minecraft-server/start-finalSetup05EnvVariables
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+if [ "$REPLACE_ENV_VARIABLES" = "TRUE" ]; then
+  echo "Replacing env variables in configs that match the prefix $ENV_VARIABLE_PREFIX..."
+  while IFS='=' read -r name value ; do
+    # check if name of env variable matches the prefix
+    # sanity check environment variables to avoid code injections
+    if [[ "$name" = $ENV_VARIABLE_PREFIX* ]] && [[ $value =~ ^[0-9a-zA-Z_:/=?.+\-]*$ ]] && [[ $name =~ ^[0-9a-zA-Z_\-]*$ ]]; then
+        echo "$name = $value"
+        find /data/ -type f -exec sed -i 's#${'"$name"'}#'"$value"'#g' {} \;
```
itzg: It looks this `find` needs to include only common filename patterns or exclude jar (and other binary files, like `.db`):

When restarting the container with an already populated data volume, I got:
```
Replacing env variables in configs that match the prefix CFG_...
CFG_MINE = testing 
sed: can't move '/data/minecraft_server.1.13.2.jarDaJAED' to '/data/minecraft_server.1.13.2.jar': No such file or directory 
```

Silthus: Yes you are right. Do you think the following file types would be enough:

- `.yml`
- `.yaml`
- `.properties`
- `.cfg`
- `.txt`

itzg: Yep, that seems to cover it.

itzg:
This is now published to Docker Hub. Thanks again for the great feature.

abifrwork:
Hi can i get the steps to run minecraft server in ec2 linux instance

itzg:
@abifrwork I should probably create a YouTube video to answer that, but it mainly boils down to:
- installing Docker on the ec2 instance
- setting up a security group to allow TCP port 25565 inbound to the ec2

The remainder of the minecraft server specifics will be covered generically by [the README](https://github.com/itzg/dockerfiles/blob/master/minecraft-server/README.md).

Issue #204, opened 2018-03-10
updating .zip file does not update the mods


samara8609:
Was running an FTB server, i followed the instructions for stopping the server, removing the container, updating the zip file, then starting a new container.  It doesn't seem to want to unzip the modpack over the existing files.

TheLeftMoose:
Just my few cents... i think this question leads to a more general question, as this is also the same behavior for all server.properties env's that is set. There is simply no bash to remove, ones unpacked or server.properties is placed in the app folder.
When using docker-compose what you descripe is somehow automated, so when you run docker-compose down it will remove the container, but of course maintain your volume on the disk.

I have been trying to think for at better solution, and maybe there should be a OVERWRITE_MODS or OVERWRITE_SERVER_PROP that could be set to true (but false as default)
But, if the OVERWRITE_MODS flag is set, would you still like to persist plugin config files inside the folders? I would... :) 
with the server.properties this is different, as you properly would have defined your configs as ENVs already.

Other thoghts was to find a way not to persist the specific files on the volume, as this will automatically lead to the config files or mods to be redownloaded when starting the container via docker-compose up -d. This approch just gives overhead as e.g.:
* there is no way to configure e.g. spigot where to put config or data, meaning you can't really volumen out just the config, or just the persistable data. You could spend time volume out specific files, but that would be a nightmare in the long run.
* The container will need to spend time downloading content upon every startup, which really is against best practices working with containers.

itzg:
Excellent food for thought. It did make me think that maybe situating the working directory separate from the location of downloaded and expanded jars might reveal  some cleaner volume attachment points.

itzg:
I need to prove this out, but you can also create arbitrary volume attachments such as at /data/FeedTheBeast/world in order to retain your world data across container restarts but leave the jars and mods to be updated fresh each time.

fabianvf:
Would you be in support of an OVERWRITE_SERVER_PROP setting? For the operator I'd like the Kubernetes configuration to have precedence over manual edits in the volume.

itzg:
So, when that is enabled then always update the server.properties with the latest environment variable values? If so, that sounds like a great idea.

fabianvf:
Awesome, I'll get that patch in as soon as I get a moment

PR #302, opened 2019-04-03
Allow user to specify that server properties be overridden on start

fabianvf:
As discussed in #204 

This was motivated by the game-server-operator minecraft deployment logic, where we map a custom Kubernetes resource to a Minecraft instance. Ideally users would be able to change the Minecraft resource and have the deployed server configuration change to reflect this.

This only handles `server.properties` related environment variables, and doesn't resolve the mods-related variables (because I don't understand how minecraft modding works well enough to do that with confidence). It also doesn't handle ops/whitelist variables, I wasn't sure if I should make a broader `OVERRIDE_CONFIGURATION` and get rid of `OVERRIDE_SERVER_PROPERTIES`, or if it would be better to have more granular `OVERRIDE_OPS` + `OVERRIDE_WHITELIST` (maybe in addition to a general `OVERRIDE_CONFIGURATION`?). I lean towards one `OVERRIDE_CONFIGURATION` variable, but was curious what everyone else thought.

### DIFF
--- a/minecraft-server/start-finalSetup04ServerProperties
+++ b/minecraft-server/start-finalSetup04ServerProperties
@@ -17,17 +17,7 @@ function setServerProp {
   fi
 }
 
-# Deploy server.properties file
-if [ ! -e $SERVER_PROPERTIES ]; then
-  echo "Creating server.properties in ${SERVER_PROPERTIES}"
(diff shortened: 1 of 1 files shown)


itzg:

PR #315, opened 2019-05-03
Make sure, the java version is always latest

JuniorJPDJ:


md5:
Sorry to come late to the game, but I wanted to point out that the Alpine OpenJDK package currently being used was discontinued in May 2019, very soon after the last comments were left here. See https://github.com/docker-library/openjdk/pull/322

I doesn't look like there are any widely replacement images based on the early access builds for musl, but there are quite a few Dockerfiles around that are downloading the builds as tarballs: https://github.com/search?l=Dockerfile&q=openjdk+early_access+musl&type=Code

itzg:
Yes, openjdk seems to have abandoned that version stream. I am thinking of making the `multiarch` branch mainline/master/latest at some point soon. Since it is based on Ubuntu I am able to support the newest JDK 8 versions and, as the name implies, provide a multi-processor architecture image with that. The downside is the increased image size, but seems a reasonable trade-off all in all.

md5:
@itzg is there a reason you want to stay on Java 8? The Mojang page doesn't seem to specify a preferred version, so perhaps a newer JRE version would have some advantages.

itzg:
Forge compatibility 

https://codakid.com/guide-to-minecraft-modding-with-java/

md5:
That's a bummer. Not trying to tell you things you already know, since I'm not at all dialed-in to the Minecraft community when you clearly are, but it looks like Forge 31.2 Minecraft 1.15.2, announced in May, has “preliminary support for Java 13+”: https://forums.minecraftforge.net/topic/85620-forge-312-minecraft-1152/

Appreciate your time answering my question

itzg:
Thanks for linking that. I’m glad to hear they are working on Java 13 support. I’d personally prefer to have the image based on 13 or 14.

itzg:
Just to clarify this image is already available for Java 13 by using the `adopt13` label.

MarcDoughty:
@JuniorJPDJ, the definition of 'latest' in Java is a world of complication. There are two long-term support branches of Java, 8 & 11, and one current branch which is at 15. Each distro maintains their own packages too, so the Java 8 in Alpine might lag behind what is actually current.

I'm doing some work in the 'multiarch-latest' branch of this project to get Java 15 up and running, and to use of some of its features. In that branch, the container is from AdoptOpenJDK, which does a great job maintaining 8, 11, and 15 at current patchlevels.

PR #335, opened 2019-06-12
add option for force re-download of server file

### LINKED ISSUE #198: [minecraft] feature request: Handle upgrade of server
So, upgrading a server can be some kind of a challenge since this image is so diverse as it is.

I see in previous issues that the only really way to do an upgrade, is to delete the server jar file, and then boot the container. This should trigger either a download or run of spigot buildtools.

Though, this approach is kinda painful, since you will need to do something by hand; It would be more nice if it was possible to pass some kind of argument that triggers an upgrade.

Any thoughts about this suggestion?

billimek:
This may partially (or fully?) resolve #198 

In some scenarios, it is not ideal to manually delete the server .jar file in order to force minecraft-server to download a fresh copy.  For situations when a new version of minecraft is released and there is steady incremental fixes made to the server type, there are new versions of the server created multiple times a day (e.g. [the 81 builds of the paper server since 1.14 was releases](https://papermc.io/downloads#Paper-1.14)).  

The option added in this PR gives server operators the option to have minecraft-server automatically download a fresh, (most likely) updated copy of the server for that particular release of minecraft.

### DIFF
--- a/minecraft-server/README.md
+++ b/minecraft-server/README.md
@@ -537,6 +537,20 @@ If it is a URL, it will only be downloaded into the `/data` directory if it wasn
 such, if you need to upgrade or re-download the JAR, then you will need to stop the container,
 remove the file from the container's `/data` directory, and start again. 
 
+## Force re-download of the server file
+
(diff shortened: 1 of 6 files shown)


itzg:
This change is now pushed to Docker Hub.

PR #337, opened 2019-06-13
Change to 8u212

sfxworks:
To have the java 8 properly take account for available memory usage available to the container.

https://blog.softwaremill.com/docker-support-in-new-java-8-finally-fd595df0ca54

### DIFF
--- a/minecraft-server/Dockerfile
+++ b/minecraft-server/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8u201-jre-alpine
+FROM openjdk:8u212-jre-slim
 
 LABEL maintainer "itzg"
 
--- a/minecraft-server/Dockerfile
+++ b/minecraft-server/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8u212-jre-slim
+FROM openjdk:8u212-jre-alpine
 
 LABEL maintainer "itzg"
 


sfxworks:
Need to change apk / packages or await for an alpine variant. Or build from source.

itzg:
Switching from Alpine to Debian is not trivial -- as you mentioned in your follow up comment, the apk packages need to be replaced with all of the equivalent debian packages. It's certainly doable, but it has been a methodical process of confirming the packages installed by the current image.

The good news is that I see there is now a `13-alpine` image, so I'm going to give that a try.

sfxworks:
Awesome. Glad to hear. Was pretty short on time this weekend. I'll change it in this branch here too so I can see the results in the pipeline.

itzg:
Oh, there's a `8u212-jre-alpine` tag? That's even better as the initial change.

sfxworks:
Yeah I just saw that one. Honestly keeping it at java 8 is good too since SpongeVanilla + SpongeForge doesn't support anything higher at this time

itzg:
Cool. I'm good to go with this PR with that change then.

itzg:
This is now available on Docker Hub. Thanks for the contribution.

sfxworks:
Thanks for accepting the PR! More to come as I move forward!

PR #346, opened 2019-07-07
Support HTTP redirects in modpack URLS

### LINKED ISSUE #345: Support HTTP redirects to jar files.
As it is, there is no support for downloading plugins that do not follow the http*jar pattern. This makes it harder for people to get plugins as they need to figure out the absolute URL of the plugin. Most websites out there don't supply this information, so you end up on CI boxes, which may or may not fail for random reasons like SSL certificate validation (totally the problem I'm having).

I think the url_effective property should be used to try and find the .jar URL behind the redirects. The following could achieve this:
```
URL_EFFECTIVE=$(curl -Ls -o /dev/null -w %{url_effective} $URL)
```

lfshr:
This allows for links such as https://dev.bukkit.org/projects/multiverse-core/files/2729939/download to work in the MOD fields.

I've been able to test the MODS implementation but not MODPACK and Manifest. If somebody could fire a test in there that would be great.

This closes #345

### DIFF
--- a/minecraft-server/start-finalSetup02Modpack
+++ b/minecraft-server/start-finalSetup02Modpack
@@ -14,24 +14,25 @@ fi
 
 # If supplied with a URL for a modpack (simple zip of jars), download it and unpack
 if [[ "$MODPACK" ]]; then
-case "X$MODPACK" in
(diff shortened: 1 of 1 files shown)


itzg:
...I don't have a MODPACK URL to test at the moment, but the code change looks good.

itzg:

PR #348, opened 2019-07-10
Add support for Fabric server

crazymanjinn:
Fabric is another modding API for minecraft that needs an installer for the server (http://fabricmc.net/). Made changes to Dockerfile and added new deploy script based on the Forge script since it appears to be similar kind of installation steps.


minecraft-server/start-deployFabric
```diff
@@ -0,0 +1,84 @@
...
+export TYPE=FABRIC
+
+FABRIC_INSTALLER=${FABRIC_INSTALLER:-}
+FABRIC_INSTALLER_URL=${FABRIC_INSTALLER_URL:-}
+FABRICVERSION=${FABRICVERSION:-LATEST}
+if [[ -z $FABRIC_INSTALLER && -z $FABRIC_INSTALLER_URL ]]; then
+  echo "Checking Fabric version information."
+  case $FABRICVERSION in
+    LATEST)
+      FABRIC_VERSION=$(python - << 'EOF'
```
itzg: I was considering removing python from the image, but the alternatives, such as

https://github.com/jeffbr13/xq

would require python based tools :). So you've given me a good reason to simply keep python included.

minecraft-server/README.md
```diff
@@ -529,6 +529,72 @@ Just change it with `SPONGEBRANCH`, such as:
...
+
+Enable Fabric server mode by adding a `-e TYPE=FABRIC` to your command-line.
+By default the container will run the latest version of [Fabric server](http://fabricmc.net/use/)
+but you can also choose to run a specific version with `-e FABRICVERSION=0.5.0.32`.
+
+    $ docker run -d -v /path/on/host:/data -e VERSION=1.14.3 \
+        -e TYPE=FABRIC -e FABRICVERSION=0.5.0.32 \
+        -p 25565:25565 -e EULA=TRUE --name mc itzg/minecraft-server
+
+To use a pre-downloaded Forge installer, place it in the attached `/data` directory and
```
itzg: Should this be
```suggestion
To use a pre-downloaded Fabric installer, place it in the attached `/data` directory and
```

minecraft-server/README.md
```diff
@@ -529,6 +529,72 @@ Just change it with `SPONGEBRANCH`, such as:
...
+specify the name of the installer file with `FABRIC_INSTALLER`, such as:
+
+    $ docker run -d -v /path/on/host:/data ... \
+        -e FABRIC_INSTALLER=fabric-installer-0.5.0.32.jar ...
+
+To download a Forge installer from a custom location, such as your own file repository, specify
```
itzg: ...and
```suggestion
To download a Fabric installer from a custom location, such as your own file repository, specify
```

minecraft-server/start-deployFabric
crazymanjinn: I'm not a huge fan of this whole "inline python script into bash" mess, but it also seemed to be the cleanest way to parse XML without having to introduce anything new into the base image.

If you're okay with adding xq into the base image, then I'll rewrite the deploy script to take advantage of it so it's a little easier to read (and also follows the same pattern in the forge deploy scripts of `curl FOO | jq BAR`)

minecraft-server/README.md
crazymanjinn: Yes it should; looks like I missed a couple places with my find-replace.

minecraft-server/start-deployFabric
itzg: Yeah, give the xq approach a try especially since like you said, it follows the pattern of the jq usage.

crazymanjinn:
I've used the proposed changes to deploy a working fabric minecraft server, and was able to join it with a fabric minecraft client. I haven't tested loading it up with mods on client and server side, but since mods are just dropping .jar files into a mods/ folder, I can't imagine anything with the deploy scripts really interfering with that.

If you don't want to merge this in before specifically testing that, that's fine with me. I'll let you know once I've confirmed that this is all working with a "fully loaded" server.

crazymanjinn:
Just changed the deploy script to use {{xq}} instead of inline'd python.

Also just spun up a modded server with it, and can confirm that this works with a fabric modded server and client.

itzg:

Issue #351, opened 2019-07-21
Using Fabric with something other than the latest version of MC?


altano:
I'm trying to use TYPE=FABRIC (recently added by @crazymanjinn, thanks!).

Unfortunately specifying:

```
TYPE=FABRIC
VERSION=1.14.2
```

Doesn't work. It's still trying to use MC 1.14.4 for some reason. These are snippets of the container startup logs:

```
Installing Fabric 0.5.0.33 using /tmp/fabric-installer-0.5.0.33.jar
Loading Fabric Installer: 0.5.0.33
Using latest loader version
Using latest game version
...

[05:56:29] [main/INFO]: Loading for game Minecraft 1.14.4
```

Since my /data volume and world already exist for MC 1.14.2 this is causing startup to crash.

Is this line using the wrong value for `-version` or something?: https://github.com/itzg/docker-minecraft-server/blob/539f7ada706724fb38becb7d527ac73ae198ed13/start-deployFabric#L53

altano:
FWIW the flag is actually documented as `mcversion` and not `version`.

altano:
Yeah I think that was the problem. Manually installing with `-mcversion` fixed it for me.

itzg:
Fixed by #352

PR #352, opened 2019-07-21
Fix -version => -mcversion

altano:
Fabric's installer uses `-mcversion` to specify the Minecraft version (https://fabricmc.net/wiki/install). The way the command is written right now it will always use the latest version of MC and ignore the `VERSION` specification.

See https://github.com/itzg/docker-minecraft-server/issues/351 for more discussion.


altano:
@crazymanjinn if you could take a look at this and give your stamp of approval that might help @itzg. Thanks!

itzg:

Issue #364, opened 2019-07-31
Typo in sed command 


iamtpage:
https://github.com/itzg/docker-minecraft-server/blob/539f7ada706724fb38becb7d527ac73ae198ed13/start-minecraftFinalSetup#L108

if `$MAX_MEMORY` is 8G, then
`sed -i "/MAX_RAM=.*/ c MAX_RAM=${MAX_MEMORY};"`
translates to:
`MAX_RAM=8G;` in `settings.cfg`
and when this is strung together in the JVM command, `java -Xmx 8G; ...` results in an invalid heap size error from the JVM.

Removing the `;` seems to resolve this issue.

itzg:
Fixed by #365

itzg:
Re-opening since we found in #366 that some (or most?) CurseForge/FTB startup scripts expect to strip the trailing character.

iamtpage:
I must be confused with how the container runs, let me explain my use case.
First a few assumptions:
- `MEMORY` by default is `1G`
- You can override `Xms` and `Xmx` by setting `INIT_MEMORY` and `MAX_MEMORY` respectively
- The docker container modifies `/data/FeedTheBeast/settings.cfg` and replaces values with sed according to environmental variables specified when the container starts up (including the ones above)

I run a docker-compose setup for this container as such:
```
  minecraft-server:
    image: hub.tpage.io/tyler/minecraft-server:latest
    ports:
      - 25565:25565
    volumes:
      - /opt/docker-local/minecraft-server:/data
    environment:
      EULA: "TRUE"
      TYPE: "CURSEFORGE"
      MANIFEST: "/data/FeedTheBeast/manifest.json"
      VERSION: "1.12.2"
      CF_SERVER_MOD: "https://www.curseforge.com/minecraft/modpacks/dmt-survival/download/2752193"
      ENABLE_QUERY: "true"
      INIT_MEMORY: "4G"
      MAX_MEMORY: "8G"
      SNOOPER_ENABLED: "false"
      MAX_BUILD_HEIGHT: "512"
      GENERATE_STRUCTURES: "true"
      ENABLE_RCON: "false"
```

When I run the container, it modified my `settings.cfg` script according to the variables above, when it gets to `MAX_RAM`, it replaces it with `8G;`. Then when it's finished, it attempts to run the server with the `java` command and it crashes. Taking a look at `settings.cfg` shows this:

```
;SETTINGS FOR SERVERSTART SCRIPTS
[...]

PR #365, opened 2019-07-31
Update start-minecraftFinalSetup to remove typo in sed command

iamtpage:
See #364


itzg:

PR #371, opened 2019-08-02
fix duplicate projects subpath

iamtpage:
`CURSE_URL_BASE` already had `/projects` appended to it, missed it by mistake. Originally wrote `redirect_url` assuming `CURSE_URL_BASE` was `https://minecraft.curseforge.com`


iamtpage:
It worked, but the files downloaded were still HTML pages not JAR files

-------- Original Message --------

itzg:
Aha. Thanks for the quick fix.

PR #379, opened 2019-08-21
Added ability to add resource-pack from environment vars

docquantum:
Added ability to change/add resource-pack link from environment variables.

Both resource-pack and resource-pack-sha1 can be set by RESOURCE_PACK and RESOURCE_PACK_SHA1


docquantum:
I use a "clear glass" resource pack on my servers, and being able to set it via environment vars would be very useful and convenient.

Maybe allowing all options to be configurable would be useful. [wiki](https://minecraft.gamepedia.com/Server.properties)

docquantum:
Check it out, I tried to model it after the others.

itzg:

PR #386, opened 2019-09-08
Add spawn protection setting

Nautigsam:
Allow to set spawn-protection setting with an env variable.


itzg:

PR #387, opened 2019-09-13
Only write EULA when it's set to `TRUE`

x3rAx:
### Issue

When you for example set the `EULA` environment variable to `true` (instead of uppercase `TRUE`) the `eula.txt` is written with a wrong value and the server won't start. After that, it is not possible to change the value using the `EULA` environment variable and you would have to change the file manually, which might be a little complicated for some users when they use docker volumes instead of bind mounts.

### Fix

This pull request fixes this by checking the value to be `TRUE` (uppercase) before writing the `eula.txt` file. For any other value, the request to set the value using `-e EULA=TRUE` is displayed.

### DIFF
--- a/start-configuration
+++ b/start-configuration
@@ -6,14 +6,7 @@ shopt -s nullglob
 export HOME=/data
 
 if [ ! -e /data/eula.txt ]; then
-  if [ "$EULA" != "" ]; then
-    echo "# Generated via Docker on $(date)" > eula.txt
-    echo "eula=$EULA" >> eula.txt
-    if [ $? != 0 ]; then
-      echo "ERROR: unable to write eula to /data. Please make sure attached directory is writable by uid=${UID}"
-      exit 2
-    fi
-  else
+  if [ "$EULA" != "TRUE" ]; then
     echo ""
     echo "Please accept the Minecraft EULA at"
     echo "  https://account.mojang.com/documents/minecraft_eula"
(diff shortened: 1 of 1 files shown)


start-configuration
```diff
@@ -6,14 +6,7 @@ shopt -s nullglob
...
 if [ ! -e /data/eula.txt ]; then
-  if [ "$EULA" != "" ]; then
-    echo "# Generated via Docker on $(date)" > eula.txt
-    echo "eula=$EULA" >> eula.txt
-    if [ $? != 0 ]; then
-      echo "ERROR: unable to write eula to /data. Please make sure attached directory is writable by uid=${UID}"
-      exit 2
-    fi
-  else
+  if [ "$EULA" != "TRUE" ]; then
```
itzg: A bash trick I have learned since this original code is the case modification expansion `${parameter^^pattern}` ([described here](http://manpages.ubuntu.com/manpages/bionic/en/man1/bash.1.html#expansion)). So, this could be even more robust by changing this to:

```suggestion
  if [ "${EULA^^}" != "TRUE" ]; then
```

start-configuration
```diff
@@ -22,8 +15,16 @@ if [ ! -e /data/eula.txt ]; then
     echo ""
     exit 1
   fi
+
+  echo "# Generated via Docker on $(date)" > eula.txt
+  echo "eula=$EULA" >> eula.txt
```
itzg: ...and this to
```suggestion
  echo "eula=${EULA^^}" >> eula.txt
```

x3rAx: Thanks for the hint! I also thought about allowing mixed case (or even `yes`) and just hard coding the `TRUE` like so:

    echo "eula=TRUE" >> eula.txt

But I didn't know if this was legally allowed or if there is any requirement, that the user has to type *exactly* the text that has to be written in the eula file. Therefore I decided to not change this and just use the value the user typed, but disallow anything that would not work.

If you don't think there is any problem in doing it the way you proposed, I'm totally fine with it and I would also prefer your solution. 😉

itzg: I think just sticking to variations of “true” is sufficient flexibility. The combination of the extra check you have added and the uppercase’ing will guard and simplify giving Minecraft the right value.

x3rAx: I've added another commit in which the value is transformed to upper case once before it is used so it does not have to be done every time the value is used.

itzg:

PR #397, opened 2019-10-11
(improvement) Offsets healthcheck to T+1m

denzuko:
Changes:
   - Added the `--start-period=5m` option to the docker healthcheck which offsets the mcstatus commands by 5 minutes to allow modpacks to complete downloading.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,7 @@ RUN apk add --no-cache -U \
 
 RUN pip install mcstatus yq
 
-ARG HEALTHCHECK_DEFAULT_START=5m
-HEALTHCHECK --start-period=${HEALTHCHECK_DEFAULT_START} CMD mcstatus localhost:$SERVER_PORT ping
+HEALTHCHECK --start-period=5m CMD mcstatus localhost:$SERVER_PORT ping
 
 RUN addgroup -g 1000 minecraft \
   && adduser -Ss /bin/false -u 1000 -G minecraft -h /home/minecraft minecraft \
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN apk add --no-cache -U \
 
 RUN pip install mcstatus yq
 
-HEALTHCHECK CMD mcstatus localhost:$SERVER_PORT ping
+HEALTHCHECK --start-period=5m CMD mcstatus localhost:$SERVER_PORT ping
 
 RUN addgroup -g 1000 minecraft \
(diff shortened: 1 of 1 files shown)


itzg:
I'm now worried this change from the default of 0 seconds to 5 minutes might be too disruptive for anyone with a system running vanilla containers and expecting containers to report readiness sooner than 5 minutes.

How about a default of 1 minute and then you can use the override approaches to increase the start period to 5 minutes for FTB, etc containers?

denzuko:
Been testing and finding that the larger packs like FTB (ie 75-125Mb+) take about 25 minutes to start otherwise the healthcheck kills the container before the zip completes.

Dockerfile
```diff
@@ -20,7 +20,8 @@ RUN apk add --no-cache -U \
 
 RUN pip install mcstatus yq
 
-HEALTHCHECK CMD mcstatus localhost:$SERVER_PORT ping
+ARG HEALTHCHECK_DEFAULT_START=5m
```
itzg: Are you good with changing this to `1m` as a compromise for vanilla containers?

denzuko: that should work.

itzg: Cool. When you change that I'll be good to merge this in.

Dockerfile
```diff
@@ -20,8 +20,7 @@ RUN apk add --no-cache -U \
 
 RUN pip install mcstatus yq
 
-ARG HEALTHCHECK_DEFAULT_START=5m
```
itzg: I liked the build arg, I was just wanting a value of `1m` instead of `5m`

denzuko: I liked it too but had a hell of a type trying to get it to work so I just revered to a know good state.

itzg: Oh, I wonder if build args don't carry over to the health check's execution environment.

Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN apk add --no-cache -U \
 
 RUN pip install mcstatus yq
 
-HEALTHCHECK CMD mcstatus localhost:$SERVER_PORT ping
+HEALTHCHECK --start-period=5m CMD mcstatus localhost:$SERVER_PORT ping
```
itzg: I know it conflicts with the PR title :) , but can you change this to 1m for the default:
```suggestion
HEALTHCHECK --start-period=1m CMD mcstatus localhost:$SERVER_PORT ping
```

denzuko: Done.

Issue #392, opened 2019-09-26
CurseForge modpack not pulled from URL


pixtim:
Hi Itzg. Thanks for the awesome docker image!

I'm trying to spin up a container from a CurseForge URL. I've followed the steps in the README but it looks like the `TYPE` and `CF_SERVER_MOD` environment variables are being ignored. A vanilla server is deployed instead of a modded one. I'm hoping that I've missed something small.

Steps to reproduce:

1.  Pull the image
```
docker pull itzg/minecraft-server
```

2. Create a CurseForge server container
```
docker run -d -it -e EULA=TRUE -p 25565:25565 --name mc itzg/minecraft-server \
-e VERSION=1.12.2 \
-e TYPE=CURSEFORGE \
-e CF_SERVER_MOD=https://www.curseforge.com/minecraft/modpacks/rlcraft
```

3. Start the container
```
docker container start mc
```
See [logs.txt](https://github.com/itzg/docker-minecraft-server/files/3660118/logs.txt)

itzg:
Ah, the `-e` arguments need to come before the image name, so try:

```
docker run -d -it -e EULA=TRUE -p 25565:25565 --name mc \
-e VERSION=1.12.2 \
-e TYPE=CURSEFORGE \
-e CF_SERVER_MOD=https://www.curseforge.com/minecraft/modpacks/rlcraft \
itzg/minecraft-server
```

alessandro-massarenti:
> Ah, the `-e` arguments need to come before the image name, so try:
> 
> ```
> docker run -d -it -e EULA=TRUE -p 25565:25565 --name mc \
> -e VERSION=1.12.2 \
> -e TYPE=CURSEFORGE \
> -e CF_SERVER_MOD=https://www.curseforge.com/minecraft/modpacks/rlcraft \
> itzg/minecraft-server
> ```

Hi, I've been trying for days to start a Curseforge server in vain.

I tried first by following the directions and composing a docker string for a custom RLCraft server, but it never started because it said that probably there is no server executable in the link I gave it.

So I decided to try any example of docker-compose contained in the examples folder of your repo; also there is nothing to do but with a different error:

```
Downloading FTB modpack...
  https://www.curseforge.com/minecraft/modpacks/rlcraft -> /data/modpacks.zip
Unpacking FTB server modpack /data/modpacks.zip ...
.unzip: short read
```

[...]

PR #398, opened 2019-10-11
(bugfix) download pack only when checksum file does not exist

denzuko:
Changes:
   - added test for /data/.generic_pack.sum to exist before downloading the modpack (allows other methods to update the pack including removing this checksum file to update the modpack)
   - added removing /data/.generic_pack.sum automatically when Environment Variable `REMOVE_OLD_MODS` is set to 'TRUE'

### DIFF
--- a/start-finalSetup02Modpack
+++ b/start-finalSetup02Modpack
@@ -14,6 +14,7 @@ if [ "$REMOVE_OLD_MODS" = "TRUE" ]; then
   else
     rm -rf /data/mods/*
   fi
+    rm -rf /data/.generic_pack.sum
 fi
 
 # If supplied with a URL for a modpack (simple zip of jars), download it and unpack
@@ -111,7 +112,7 @@ esac
 fi
 
 if [[ "${GENERIC_PACK}" ]]; then
-  if isURL "${GENERIC_PACK}"; then
+  if [ -f /data/.generic_pack.sum && isURL "${GENERIC_PACK}"]; then
     generic_pack_url=${GENERIC_PACK}
     GENERIC_PACK=/tmp/$(basename ${generic_pack_url})
     echo "Downloading generic pack from ${generic_pack_url} ..."


start-finalSetup02Modpack
```diff
@@ -111,7 +112,7 @@ esac
 fi
 
 if [[ "${GENERIC_PACK}" ]]; then
-  if isURL "${GENERIC_PACK}"; then
+  if [ -f /data/.generic_pack.sum && isURL "${GENERIC_PACK}"]; then
```
itzg: I should have included some comments in the code here. I'll explain my logic, but I can also see the need for the change you're proposing here:

My theory was that the generic pack at the given URL might have changed since last downloaded/installed. So, at the expense of downloading the generic pack archive on every startup, it guarantees that the latest is always in use by comparing the checksum of the downloaded archive with the `.generic_pack.sum`.

For large modpacks that download expense could be quite significant, so I'd propose that you add a check for a new environment variable, such as `SKIP_GENERIC_PACK_CHECKSUM` to have it skip the download when the checksum file is present.

start-finalSetup02Modpack
```diff
@@ -14,6 +14,7 @@ if [ "$REMOVE_OLD_MODS" = "TRUE" ]; then
   else
     rm -rf /data/mods/*
   fi
+    rm -rf /data/.generic_pack.sum
```
itzg: Good idea.

itzg:
And I reviewed this PR before seeing your comment

https://github.com/itzg/docker-minecraft-server/issues/392#issuecomment-540883658

As you can gather

> the shasum is out of sync

was the need for downloading the generic pack every time. Otherwise there's not a consistent way to determine the checksum of the archive prior downloading. There are some techniques to query for the last modified time of remote file without downloading, but that is not totally reliable depending on the web server and configuration.

start-finalSetup02Modpack
itzg: I was just thinking more about this change and was thinking it would be nice if there was a web service out there that given a URL would return a checksum for the file at that URL. I did a quick search and couldn't find anything, so I might have to implement one if only for the fun of it.

denzuko: oh you mean something like https://developers.virustotal.com/reference#file-report? Where one can post a url and get back a hash in either md5, sha256, sha1, plus the scan date for that hash.

itzg: Very helpful. It looks like `/url/report` (https://developers.virustotal.com/reference#url-report) might be exactly what I was thinking since it could be given the remote pack URL to analyze.

tomlarkworthy: hey the featureyou want already exists. It is the etag mechanism and is used extensively on the web for intermediate caching.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag
Check the headers!

PR #400, opened 2019-10-14
Fixed JVM_XX_OPTS for FTB servers

jmyoung:
Creation of settings-local.sh for FTB instances does not propagate JVM_XX_OPTS and JVM_OPTS into the file, leading to those instances ignoring those parameters.  This fix adds those to the exported JAVA_PARAMETERS env var, which should have the desired effect.


itzg:

jmyoung:
Thanks much, that was my first ever pull request :)

Issue #360, opened 2019-07-29
Consider implementing alternative JVM


DJDANNY123:
Alternative JVM's such as openJ9 ( https://www.eclipse.org/openj9/ or https://hub.docker.com/r/adoptopenjdk/openjdk8-openj9) can provide benefits such as higher performance/ lower memory usage ect. 
Have you considered using this as the base image, or making this an option?

Enf0:
I second this. In the meantime, you can easily make the change yourself by cloning the repo and changing the first line in the Dockerfile from

`FROM openjdk:8u212-jre-alpine`
to
`FROM adoptopenjdk/openjdk9-openj9:alpine`
and running docker build.

itzg:
_Sorry @DJDANNY123  and @Enf0 for the delay in responding -- this got lost in my notification list._

I have admit I haven't considered alternatives beyond the stock openjdk builds/images. I'm definitely willing to give this a try especially since I have a lot of faith the Eclipse foundation has created a good solution there.

Since the startup time of 1.14.x has gotten pretty horrible, it seems like running a few timings of each base image reaching world generated and ready state would be an easy case study.

I'll give that a try, but any contributions you can provide of measurement ideas and/or data would be fantastic.

Enf0:
Here's a very interesting post about running Minecraft with OpenJ9:
https://steinborn.me/posts/tuning-minecraft-openj9/

itzg:
Thanks. I'll add `-Xtune:virtualized` into the mix. So far, the startup time is not improved with j9, but memory usage is roughly cut in half.

itzg:
Docker Hub is currently building the image based on OpenJ9. I have documented the two new opt-in tuning options in the JVM section of the README

https://github.com/itzg/docker-minecraft-server/#jvm-options

itzg:
This is now available on Docker Hub.

itzg:
There were reports in #374  of the container not starting, so I have reverted the openj9 base image change and moved instead to a branch and image tag of `openj9`.

itzg:
@DJDANNY123 and @Enf0 , can you test out `itzg/minecraft-server:openj9` to see if it works ok for you also?

Enf0:
Just pulled the openj9 build, seems to work just fine. Haven't had the chance to compare performance yet, but the server is running!

Here are my java flags, as per the article above:
[...]

PR #408, opened 2019-10-29
docker-versions-create init

wlp7s0:
I've created a small sh script to simplify docker container creation with version.
Hope it will speedup [#360 ](https://github.com/itzg/docker-minecraft-server/issues/360) issue resolution. 
I've tested it to the point of actual pushes. As I don't know the environment configuration of your workstation - I assume the script to be changed to work in your environment. 
Upon completion you need to do some manual steps:
1. Merge github branch that the script created
2. Update docker hub description and add new version with link to a dockerfile that was created by this script.

### DIFF
--- a/docker-versions-create.sh
+++ b/docker-versions-create.sh
@@ -1,119 +1,77 @@
 #!/bin/bash
 #set -x
-# Use this array to build different versions of the docker container
-version_array=('openjdk:8u212-jre-alpine' \
-  'adoptopenjdk/openjdk8-openj9:alpine' \
-  'adoptopenjdk/openjdk8:alpine-jre' \
-  'adoptopenjdk/openjdk8:alpine-jre-nightly')
-# This variable will determine what build would be the default one (without version)
-defailt_version='openjdk:8u212-jre-alpine'
+# Use this variable to indicate a list of branches that docker hub is watching
+branches_list=('openj9' 'another_name')
 
-function ShowHelp {
-  cat <<EOF
-This script will create and push new docker images to docker hub.
-It will also create a new git branch on github. Please consider merging new brunch with master if you are satisfied with changes.
-To add or remove different version from the build - edit this file and add\remove version from 'version_array'
-By default, it will skip recreation of image if Dockerfile for this image was found in ./versions dir
-
-Usage:
-  docker-versions-create.sh [--help, --ask]
-  --help: shows this massage
-  --ask:  don't skip image recreation and ask if recreation is required for each version.
-EOF
-  exit 0
+function TrapExit {
+  echo "Checking out back in master"
+  git checkout master
 }
 
-function DockerBuild {
-  local container_version docker_file tag_version
-  container_version="$1"
-  docker_file="$2"
-  tag_version=$(echo "$container_version" | tr '/:' '-')
-  cp ./Dockerfile ./versions/"$docker_file"
-  sed -e "s|^FROM.*\$|FROM $container_version|" -i ./versions/"$docker_file"
-  if [[ "$container_version" == "$defailt_version" ]]; then
-    # skipping version tag creation
-    docker build -f ./versions/"$docker_file" -t itzg/minecraft-server .
+trap TrapExit EXIT SIGTERM
+
+test -d ./.git || { echo ".git folder was not found. Please start this script from root directory of the project!";
+  exit 1; }
+
+# Making sure we are in master
+git checkout master
+#Updating git branch list. Cause the branch may be created from diff place
+# Are tags in use? if not - remove --tags
+git fetch --all --tags || { \
+  echo "Can't fetch changes from the remote. Permissions?"; \
+  exit 1; }
+# Updating current master from remote. Cause repo could be changed
+git merge origin/master || { echo "Can't update local master from remote repo!"; \
+  exit 1; }
+
+git_branches=$(git branch -a)
+
+for branch in "${branches_list[@]}"; do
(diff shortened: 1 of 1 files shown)


itzg:
I should have mentioned earlier that I already have a branch-per-tag strategy going, such as

https://github.com/itzg/docker-minecraft-server/tree/openj9

and was mostly needing to automate the master->branch merging of any updates made in master. I also use Docker Hub currently for builds, so any push to a branch that it is watching will take care of the docker build/push automatically. 

With that I was thinking the automation script would work against a list of branches, checking out each, and performing a `git merge` from master into the branch. If the merge is successful, it could git commit and push the change. If the merge fails, it would "pause" to allow for manually resolving conflicts, etc. The quirky part is that the script would be initiated while on master, but check out over itself while traversing the branches. That might require some research and/or creative solutions.

I know what I describe there is a fairly large difference from what you implemented, but would be willing to at least add a command line option to the run the behavior I described?

itzg:
> Generally, what you need is to update a list of branches to the latest master and create a push if update encountered no errors. Got it.

You summed it up perfectly :)

wlp7s0:
Sorry for a delay on this task.
Can you look into this again? I did change it to use actual branches. I did test it to some extend, but can't test it completely.

docker-versions-create.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+#set -x
+# Use this variable to indicate a list of branches that docker hub is watching
+branches_list=('openj9' 'another_name')
```
itzg: Looking back at your initial findings, it sounded like `openj9-nightly` would be another branch, so I have created and pushed that one too.

docker-versions-create.sh
```diff
@@ -0,0 +1,77 @@
...
+# Making sure we are in master
+git checkout master
+#Updating git branch list. Cause the branch may be created from diff place
+# Are tags in use? if not - remove --tags
+git fetch --all --tags || { \
+  echo "Can't fetch changes from the remote. Permissions?"; \
+  exit 1; }
+# Updating current master from remote. Cause repo could be changed
+git merge origin/master || { echo "Can't update local master from remote repo!"; \
+  exit 1; }
```
itzg: I think these two could be combined into a `git pull` since it's a good assumption that the local `master` branch is already associated with an upstream.

docker-versions-create.sh
```diff
@@ -0,0 +1,77 @@
...
+
+The following commands may encounter an error!
+This is completely fine if the changes were made locally and remote branch doesn't know about them.
+
+EOL
+          # Updating branch from remote before trying again
+          git checkout master
+          git fetch --all
+          git pull -a
+          git checkout "$branch"
```
itzg: I'm trying to decide if we a "y" here should check with `git status` that the merge conflicts (the likely cause of a merge failure) were fixed and committed by the user. Then after answering "y" this code could resume with the push.  If they answer "n" then could do a `git checkout -f master` to throw away branch-local merge failure and go back to master.

wlp7s0: Sorry, I don't understand what do you mean. 
The idea here is that if there is an error on the merge of the branch with master - this error will be fixed manually and git status will be checked after a manual commit. 

There is a trap, that will start a function that will perform checkout to master without throwing away any changes. There is no need to checkout in master after pressing 'n'. The only difference is in forcing.

itzg: Thinking through the scripts logic here I think it turns out my suggestion is already covered.

wlp7s0:
Yeah, sure. I mean.. you will use this. If you feel that that's OK for you to use. 
I'm still using modified first version as I need to build it on the server itself. 
Also I would like to suggest to take a look into those versions of the java VM for new branches (I tried them all. Depends on the java application one is using inside a container - they have diff benefits.):
adoptopenjdk/openjdk8:alpine-jre
adoptopenjdk/openjdk8:alpine-slim
adoptopenjdk/openjdk8:alpine-jre-nightly
mcr.microsoft.com/java/jre:8u232-zulu-alpine (Oracle jdk. Found that it to be better optimized for CPU)

Also, I think it will be a good idea to include 9, 10 and 11 versions of java too. As an example, Forge for MC 1.14 on java 8 works about 20% slower than on java 10 and this is a huge benefit on heavy-moded servers. 

Your scripts that are already in a repository work awesome with any java version. I didn't change anything else, except "FROM" line in Dockerfile, to build the list above plus different java versions.

itzg:
What base images of Java provide 9, 10, and 11 for alpine? I don't see those tags available for `openjdk` [here](https://github.com/docker-library/docs/blob/master/openjdk/README.md#supported-tags-and-respective-dockerfile-links).

itzg:
...sorry, I think I answered by own question since I found ones like `adoptopenjdk/openjdk11:alpine-jre`

itzg:
FYI I have now added an `adopt11` image tag and added batch mode support https://github.com/itzg/docker-minecraft-server/commit/382336d39d2b48a7b4e4e7c419d9624f1e951149 to prepare for me using [github actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions) to automate the script invocation on push to master.

itzg:
@wlp7s0 I just wanted to take a moment to thank you again for this script. I have used it several times now and it has made maintaining the alternate image tags so much easier.

wlp7s0:
@itzg it is good to hear! I glad it helped and thank you for a kind message!

Issue #413, opened 2019-11-19
Install of CurseForge modpack silently terminates


hannah98:
> I didn't find any obvious exit points that would cause the quiet container termination, so I thought I would try recreating your scenario. Can you provide the `manifest.json` that you're using?

_Originally posted by @itzg in https://github.com/itzg/docker-minecraft-server/issues/404#issuecomment-545915612_

---

@obsessionwithfire do you mind opening a new issue? Yours is definitely in the general realm of FTB/CurseForge challenges, but seems to be different since the startup logic is quietly failing at the point:

```
Looking for Feed-The-Beast / CurseForge server modpack.
```

I probably have some missing error checking around there.

_Originally posted by @itzg in https://github.com/itzg/docker-minecraft-server/issues/392#issuecomment-545444716_

---

I have done some debugging and [this is the line](https://github.com/itzg/docker-minecraft-server/blob/581d5fde1d504b99d84be8612f1e11a501f19fe2/start-finalSetup02Modpack#L81) where it is silently failing:

```
curl -Ls -o /dev/null -w %{url_effective} $MANIFEST
```

My MANIFEST variable is equal to ```/data/FeedTheBeast/manifest.json``` but when I run that curl command inside your image/container, it exits with an error code 3.

To reproduce, run this command at the shell inside your container:
```
curl -Ls -o /dev/null -w %{url_effective} /data/FeedTheBeast/manifest.json
echo $?
```

My output is:

```
/data/FeedTheBeast/manifest.json
3
```

And it is failing at that point because ```set -e``` [is turned on](https://github.com/itzg/docker-minecraft-server/blob/581d5fde1d504b99d84be8612f1e11a501f19fe2/start-finalSetup02Modpack#L3) in the script.

I thought about submitting a pull request to fix but I wasn't sure how you would want to handle it.  I've seen you reference this in a couple of other tickets (listed above) so I thought I'd submit an issue with the research.

itzg:
Great research @hannah98 . A PR would be fantastic. As you might have seen, there's logic nearby that makes use of the `isURL` utility function and I could see a solution using that to conditionally skip the downloading of the manifest.json.

PR #414, opened 2019-11-20
Better handle CurseForge manifest.json

hannah98:
Pull request to address issue #413 

Feel free to push back if you don't like the changes. 

* In my changes, I renamed the MANIFEST_URL variable to be MANIFEST_FILE for the block of code that handles the manifest file.
* I first check if the supplied ENV variable MANIFEST is an existing file, and use that.
* Next check is if the supplied MANIFEST is a url.  If so, use curl to download remote manifest.json file and save to /tmp, and use that file for MANIFEST_FILE.
* Last "check" is the else - to catch if it wasn't an existing file and not a URL.  I display a message but chose not to exit/fail because I didn't know if that would break anything else.  It could probably use an exit between lines 87-88.

I also fixed what I believe is a bug in the start-utils isURL function since my changes are using that function.

### DIFF
--- a/start-finalSetup02Modpack
+++ b/start-finalSetup02Modpack
@@ -78,10 +78,18 @@ done
 fi
 
 if [[ "$MANIFEST" ]]; then
-EFFECTIVE_MANIFEST_URL=$(curl -Ls -o /dev/null -w %{url_effective} $MANIFEST)
-case "X$EFFECTIVE_MANIFEST_URL" in
+    if [[ -e "$MANIFEST" ]]; then
+        EFFECTIVE_MANIFEST_FILE=$MANIFEST
+    elif isURL "$MANIFEST"; then
+        EFFECTIVE_MANIFEST_FILE=/tmp/manifest.json
+        curl -Ls -o $EFFECTIVE_MANIFEST_FILE "$MANIFEST"
+    else
+        echo "MANIFEST='$MANIFEST' is not a valid manifest url or location"
+    fi
+
+case "X$EFFECTIVE_MANIFEST_FILE" in
   X*.json)
-    if [ -f "${EFFECTIVE_MANIFEST_URL}" ]; then
+    if [ -f "${EFFECTIVE_MANIFEST_FILE}" ]; then
       MOD_DIR=${FTB_BASE_DIR:-/data}/mods
       if [ ! -d "$MOD_DIR" ]
       then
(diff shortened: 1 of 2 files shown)


start-finalSetup02Modpack
```diff
@@ -78,18 +78,26 @@ done
 fi
 
 if [[ "$MANIFEST" ]]; then
-EFFECTIVE_MANIFEST_URL=$(curl -Ls -o /dev/null -w %{url_effective} $MANIFEST)
-case "X$EFFECTIVE_MANIFEST_URL" in
+    if [[ -e "$MANIFEST" ]]; then
+        EFFECTIVE_MANIFEST_FILE=$MANIFEST
+    elif isURL "$MANIFEST"; then
+        EFFECTIVE_MANIFEST_FILE=/tmp/manifest.json
+        curl -Ls -o $EFFECTIVE_MANIFEST_FILE "$MANIFEST"
```
itzg: A previous contribution (https://github.com/itzg/docker-minecraft-server/commit/df9909a8e4c499f5d6b9c0bc327a7bba6ea52a9a) discovered that the weird-looking-two-step process was needed to resolve the effective URL and then download the file from that.

I think this combination of the previous line and yours would do it:
```suggestion
        EFFECTIVE_MANIFEST_URL=$(curl -Ls -o /dev/null -w %{url_effective} $MANIFEST)
        curl -Ls -o $EFFECTIVE_MANIFEST_FILE "$EFFECTIVE_MANIFEST_URL"
```

start-utils
```diff
@@ -3,7 +3,7 @@
 function isURL {
   local value=$1
 
-  if [[ ${value:0:8} == "https://" || ${value:0:7} = "http://" ]]; then
+  if [[ ${value:0:8} == "https://" || ${value:0:7} == "http://" ]]; then
```
itzg: I wonder if this quietly worked since most sites are https now.

start-finalSetup02Modpack
```diff
@@ -78,18 +78,26 @@ done
...
+    else
+        echo "MANIFEST='$MANIFEST' is not a valid manifest url or location"
```
itzg: Yeah, I think an exit in this condition makes sense to me too.

start-utils
hannah98: I ran some tests on it and the old way worked too... I [looked it up](https://unix.stackexchange.com/a/72041/1842) and apparently both ways work - I learned something today.

Probably good to be consistent I guess.

start-finalSetup02Modpack
hannah98: Made these changes and pushed in new commit

start-finalSetup02Modpack
hannah98: Added exit in new commit

start-utils
itzg: That's good actually, because that was the same thing I was understanding from my man-page reading.

Issue #442, opened 2020-03-02
Log prefix


scriptcoded:
I'm building a Minecraft server management platform, and I've decided to use your Docker image as a base for running the actual servers.

One thing that's a bit cumbersome right now is differentiating between the logs from the wrapper, and the actual Minecraft server. Would it be a fair idea to prefix the logs with something like `[wrapper]` and `[minecraft]`? That would make it way easier to separate logs correctly. It would also make the output more readable IMO.

I could create a PR for a suggested implementation if it's something that sounds interesting.

Also, great work with this! At first I was using my own image, but this just makes everything 20 times easier!

itzg:
That sounds like a great use case!

I'm definitely open to a PR. Since specifically this logging transfer would be annoying to intercept

https://github.com/itzg/mc-server-runner/blob/master/main.go#L75

I prefer your suggestion of prepending the wrapper's logs with something like `[wrapper]` or `[mc-server-runner]`.

scriptcoded:
Would you prefer having it implemented in the runner instead of the shell scripts in this repo? I added a crude implementation for my own personal usage. It instead replaces the `echo` call and also pipes the output of `mc-server-runner` to `sed`. That effectively shifts the responsibility of labeling the logs to the log aggregator, in this case the actual docker image.

itzg:
Ah, that could work. Can you submit a PR for that then and I'll see how that looks? If we decide to go with an mc-server-runner change, then I can make that change pretty easily too.

scriptcoded:
Now implemented in #444

itzg:
Keep me posted on your management platform. I developed a now-defunct one a long time ago https://github.com/moorkop/mccy-engine , so am always interested in these types of solutions.

itzg:
And apparently a "long time ago" was only 3 years, but that feels like forever in Docker-evolution-time. :)

scriptcoded:
Will do! I'll warn you though. To start of it will only be a personal project. I'm also doing it partially to learn Go, so only the server wrapper will use it. The actual web server will be implemented in Node.js until I feel that I'm confident enough with Go. So it'll be a mess ;)

I'll have a look at your project as well. Perhaps there's some good takeaways.

itzg:
> I'll have a look at your project as well. Perhaps there's some good takeaways.

Go for it!

PR #444, opened 2020-03-03
Added log prefixes

scriptcoded:
In response to #442

Quirks that might be unfavorable:
- Requires importing `start-utils` in every file
- Requires the use of `log` instead` of `echo`
- `[wrapper]` tag is configurable in the `start-utils` file, but `[minecraft]` is repeated twice in start-minecraftFinalSetup and has no config.

I had some thought about configuration, but my own opinion is that logger names shouldn't have to be configurable from the user side. I can't see a case where the user would benefit from changing the logger names, as long as they are kept unique.

These changes have really just served as a workaround for personal use, so please, rip them all out if need be :)


wlp7s0:
@scriptcoded this is tremendous work. And it will be a problem to change it in the feature. I suggest using bash alias to do this job. It will require 2 lines added. What do you think? smth like this:
1. add global alias for bash for next scrips to use it
`echo 'alias echo="echo -e \"\033[0;34m[wrapper]\033[0m $*\""' >> ~/.bashrc`
2. actually use this alias in this script too, because .bashrc will not be executed here anymore. 
alias echo="echo -e \"\033[0;34m[wrapper]\033[0m $*\""
at the beginning of the 
[start](https://github.com/itzg/docker-minecraft-server/blob/master/start) file.

start-minecraftFinalSetup
```diff
@@ -111,21 +111,21 @@ EOF
...
   fi
 
-  echo "Starting the Minecraft server..."
+  log "Starting the Minecraft server..."
   JVM_OPTS="-Xms${INIT_MEMORY} -Xmx${MAX_MEMORY} ${JVM_OPTS}"
   if isTrue ${DEBUG_EXEC}; then
     set -x
   fi
-  exec mc-server-runner ${bootstrapArgs} ${mcServerRunnerArgs} java $JVM_XX_OPTS $JVM_OPTS $expandedDOpts -jar $SERVER "$@" $EXTRA_ARGS
+  exec mc-server-runner ${bootstrapArgs} ${mcServerRunnerArgs} java $JVM_XX_OPTS $JVM_OPTS $expandedDOpts -jar $SERVER "$@" $EXTRA_ARGS | sed "s/^/[minecraft] /"
```
itzg: Since it's an `exec` I'm not sure if piping to `sed` actually works...or worse, it might default the point of the `exec` and insert a shell process as a parent of `mc-server-runner` process. Can you provide a `ps -ef` of a container running with these changes so I can confirm the process tree?

start-user
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
itzg: I don't see where this file gets used

start-utils
```diff
@@ -37,6 +37,10 @@ function isDebugging {
 
 function debug {
   if isDebugging; then
-    echo "DEBUG: $*"
+    log "DEBUG: $*"
   fi
 }
+
+function log {
+  echo "[wrapper] $*"
```
itzg: I'm good with hardcoding this; however, "wrapper" still doesn't feel right to me. How about "[startup]" or "[init]"?

scriptcoded:
> @scriptcoded this is tremendous work. And it will be a problem to change it in the feature. I suggest using bash alias to do this job. It will require 2 lines added. What do you think? smth like this:
> 
> 1. add global alias for bash for next scrips to use it
>    `echo 'alias echo="echo -e \"\033[0;34m[wrapper]\033[0m $*\""' >> ~/.bashrc`
> 2. actually use this alias in this script too, because .bashrc will not be executed here anymore.
>    `alias echo="echo -e \"\033[0;34m[wrapper]\033[0m $*\""`
>    at the beginning of the
>    [start](https://github.com/itzg/docker-minecraft-server/blob/master/start) file.

I personally feel very strongly about modifying native functionality. However, I understand that large changes like my proposal might be unwanted. I guess it boils down to personal preference, and after all this is @itzg's project.

@itzg As for escape codes they make parsing the output very cumbersome. Perhaps adding a flag to enable colorized output would be a solution?

start-user
scriptcoded: Somehow I missed removing this file. It's nothing but remains from another solution I had.

start-utils
scriptcoded: `[init]` sound good in my ears. Even better would be to perhaps have something more descriptive. I can't figure out a good name though, since `[minecraft-server]` would indicate the actual Minecraft server.

start-minecraftFinalSetup
scriptcoded: It works for sure. At least my output looks perfectly fine:
```
2020-03-03T16:30:15.739032751Z  [wrapper] Copying any mods over...
2020-03-03T16:30:15.742490670Z  [wrapper] Setting initial memory to 1G and max to 1G
2020-03-03T16:30:15.742728666Z  [wrapper] Starting the Minecraft server...
2020-03-03T16:30:18.490584942Z  [minecraft] [16:30:18] [main/INFO]: Loading for game Minecraft 1.15.2
2020-03-03T16:30:18.563249509Z  [minecraft] [16:30:18] [main/INFO]: [FabricLoader] Loading 2 mods: minecraft@1.15.2, fabricloader@0.7.8+build.184
```

As for how it actually runs:
```
$ docker exec 8234897ee6ee ps -ef
PID   USER     TIME  COMMAND
    1 minecraf  0:00 {start-minecraft} /bin/bash /start-minecraftFinalSetup
   38 minecraf  0:00 mc-server-runner --stop-duration 60s java -XX:+UseG1GC -Xms1G -Xmx1G -jar fabric-server-launch.jar
   39 minecraf  0:00 sed s/^/[minecraft] /
   46 minecraf  1:24 java -XX:+UseG1GC -Xms1G -Xmx1G -jar fabric-server-launch.jar
  106 root      0:00 ps -ef
```

I'll be perfectly honest and say that I don't quite follow what issues might arise from doing it the way I'm doing it. Are you thinking about performance?

wlp7s0:
@itzg oh, yeah. You are correct, no need to colourize the output. Didn't think about no-tty deploys. 

@scriptcoded I would really like to understand what do you mean by 
> modifying native functionality
I would like to know it from my personal perspective. I don't see any problem with modifying bashrc inside the container. May be my approach is bad? Can you elaborate please? Because for me it is more like configuring the environment rather than changing functionality.

scriptcoded:
@wlp7s0 I like to be careful changing behaviors that are standard. For example, `echo` outputs a raw line of text with line break at the end. Let's say that another script uses something like the following: 
```bash
echo "foo" >> bar.txt
```
You would expect `bar.txt` to contain `foo`, but instead it will contain `[init] foo`. It might not be a problem in this case. It's a rather small project after all, and @itzg will probably be there checking all PRs, but I still find it to be a bad practice changing behaviors that are taken for granted.

wlp7s0:
@scriptcoded I see. This is true, but after modifying global alias, this behaviour is expected. Thank you for the explanation.

itzg:
Yeah, all in all I prefer the `log` usage even if it did mean a lot of little changes. If you'll tweak the log prefix I'm good to approve and merge this.

scriptcoded:
@itzg I'll do that. Did we settle on "init"?

itzg:
“init” works for me

start
```diff
@@ -30,12 +32,12 @@ if [ $(id -u) = 0 ]; then
...
 
   if [[ $(stat -c "%u" /data) != $UID ]]; then
-    echo "Changing ownership of /data to $UID ..."
+    log "Changing ownership of /data to $UID ..."
     chown -R ${runAsUser}:${runAsGroup} /data
   fi
 
   if [[ ${SKIP_NSSWITCH_CONF^^} != TRUE ]]; then
-    echo 'hosts: files dns' > /etc/nsswitch.conf
+    log 'hosts: files dns' > /etc/nsswitch.conf
```
itzg: This needs to remain an `echo` ...that was a good call not to do the alias'ing.

start-configuration
```diff
@@ -8,28 +10,28 @@ export HOME=/data
...
+    log "by adding the following immediately after 'docker run':"
+    log "  -e EULA=TRUE"
+    log ""
     exit 1
   fi
 
-  echo "# Generated via Docker on $(date)" > eula.txt
-  echo "eula=$EULA" >> eula.txt
+  log "# Generated via Docker on $(date)" > eula.txt
+  log "eula=$EULA" >> eula.txt
```
itzg: These need to remain an `echo`

start-deployFabric
```diff
@@ -63,12 +63,12 @@ if [[ ! -e $installMarker ]]; then
...
   if (($tries < 0)); then
-    echo "Fabric failed to install after several tries." >&2
+    log "Fabric failed to install after several tries." >&2
     exit 10
   fi
   export SERVER=fabric-server-launch.jar
-  echo "Using server $SERVER"
-  echo $SERVER > $installMarker
+  log "Using server $SERVER"
+  log $SERVER > $installMarker
```
itzg: This needs to remain an `echo`

start-deployForge
```diff
@@ -88,23 +90,23 @@ if [ ! -e $installMarker ]; then
...
-    echo "Unable to derive server jar for Forge"
+    log "Unable to derive server jar for Forge"
     exit 2
   fi
 
   export SERVER=$latest
-  echo "Using server $SERVER"
-  echo $SERVER > $installMarker
+  log "Using server $SERVER"
+  log $SERVER > $installMarker
```
itzg: This needs to remain an `echo`

start-finalSetup04ServerProperties
```diff
@@ -89,16 +91,16 @@ function customizeServerProps {
...
         ;;
     esac
     setServerProp "difficulty" "$DIFFICULTY"
   fi
 
   if [ -n "$MODE" ]; then
-    echo "Setting mode"
-    MODE_LC=$( echo $MODE | tr '[:upper:]' '[:lower:]' )
+    log "Setting mode"
+    MODE_LC=$( log $MODE | tr '[:upper:]' '[:lower:]' )
```
itzg: This needs to remain an `echo`

start-minecraftFinalSetup
```diff
@@ -3,61 +3,61 @@
 . /start-utils
 
 if [ -n "$OPS" ]; then
-  echo "Setting/adding ops"
+  log "Setting/adding ops"
   rm -rf ops.txt.converted
-  echo $OPS | awk -v RS=, '{print}' > ops.txt
+  log $OPS | awk -v RS=, '{print}' > ops.txt
```
itzg: This needs to remain an `echo`

start-minecraftFinalSetup
```diff
@@ -3,61 +3,61 @@
...
 fi
 
 if [ -n "$WHITELIST" ]; then
-  echo "Setting whitelist"
+  log "Setting whitelist"
   rm -rf white-list.txt.converted
-  echo $WHITELIST | awk -v RS=, '{print}' > white-list.txt
+  log $WHITELIST | awk -v RS=, '{print}' > white-list.txt
```
itzg: This needs to remain an `echo`

start-minecraftFinalSetup
```diff
@@ -3,61 +3,61 @@
...
 # Make sure files exist and are valid JSON (for pre-1.12 to 1.12 upgrades)
-echo "Checking for JSON files."
+log "Checking for JSON files."
 JSON_FILES=$(find . -maxdepth 1 -name '*.json')
 for j in $JSON_FILES; do
   if [[ $(cat $j | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//') == "" ]]; then
-    echo "Fixing JSON $j"
-    echo '[]' > $j
+    log "Fixing JSON $j"
+    log '[]' > $j
```
itzg: This needs to remain an `echo`

start-minecraftFinalSetup
itzg: Sorry, I didn't realize `ps` by default on alpine doesn't show the parent PID of each process. Please re-run with

```
ps -ef -o pid,ppid,args
```

The problem I'm worried about is signal propagation, specifically the SIGTERM when the container is stopped. The main reason I created mc-server-runner was to intercept the SIGTERM and send a `stop` command to gracefully stop the minecraft server since it itself doesn't properly handle SIGTERM. [This article](https://medium.com/@gchudnov/trapping-signals-in-docker-containers-7a57fdda7d86) explains it pretty nicely.

For comparison, here's what the process tree looks like on master:
```
PID   PPID  COMMAND
    1     0 mc-server-runner --stop-duration 60s java -XX:+UseG1GC -Xms1G -Xmx1G -jar minecraft_server.1.15.2.jar
   56     1 java -XX:+UseG1GC -Xms1G -Xmx1G -jar minecraft_server.1.15.2.jar
```

where mc-server-runner must become PID 1 to intercept the signals.

So...long story, short, I think it'll take an enhancement in mc-server-runner to properly prepend those logs. FWIW, I have been wanting to enrich those logs with a date, in addition to time, so that would be a reasonable modification.

start
scriptcoded: Good thing you catched this. Silly mistake from my part, and god thank PRs. I'll make a fix for it.

start-deployFabric
scriptcoded: I'm assuming you only want to change line 71?

```suggestion
  echo $SERVER > $installMarker
```

start-finalSetup04ServerProperties
scriptcoded: Also here, I'm assuming you only want to change line 103?

```suggestion
    MODE_LC=$( echo $MODE | tr '[:upper:]' '[:lower:]' )
```

start-deployBukkitSpigot
```diff
@@ -1,24 +1,26 @@
...
+  log "Building Spigot $VANILLA_VERSION from source, might take a while, get some coffee"
   rm -rf /data/temp
   mkdir /data/temp
   cd /data/temp
 
   jvmOpts="-Xms${INIT_MEMORY:-$MEMORY} -Xmx${MAX_MEMORY:-$MEMORY}"
 
   curl -sSL -o /data/temp/BuildTools.jar https://hub.spigotmc.org/jenkins/job/BuildTools/lastSuccessfulBuild/artifact/target/BuildTools.jar && \
-    java $jvmOpts -jar /data/temp/BuildTools.jar --rev $VANILLA_VERSION 2>&1 |tee /data/spigot_build.log| while read l; do echo -n .; done; echo "done"
+    java $jvmOpts -jar /data/temp/BuildTools.jar --rev $VANILLA_VERSION 2>&1 |tee /data/spigot_build.log| while read l; do log -n .; done; log "done"
```
scriptcoded: Do we want this first log to be a log or echo? I'm assuming it's only for logging the build process, but I find it a little hard to read.

scriptcoded:
@itzg (couldn't reply to your comment for some reason)

Your suspicion was correct, and the process tree becomes odd:
```
PID   PPID  COMMAND
    1     0 {start-minecraft} /bin/bash /start-minecraftFinalSetup
   50     1 mc-server-runner --stop-duration 60s java -XX:+UseG1GC -Xms1G -Xmx1G -jar minecraft_server.1.15.2.jar
   51     1 sed s/^/[minecraft] /
   58    50 java -XX:+UseG1GC -Xms1G -Xmx1G -jar minecraft_server.1.15.2.jar
```

I'm getting the same output as you if I remove the `sed`.

After having looked up a bit more about signal propagation, I agree that the best place to add it would be in `mc-server-runner`.

When it comes to timestamp, `docker logs -t` will show output with timestamps. I guess the benefit from adding timestamps to `mc-server-runner` would be to have them show in `docker run` and perhaps to remove the potential delay from actual log to timestamp appending. That's a discussion for another thread though.

start-deployFabric
itzg: Thanks for checking. Yes, just line 71

start-finalSetup04ServerProperties
itzg: Yes...and using “suggested change” Would have been good of me :)

start-deployBukkitSpigot
itzg: As you suspected it’s just for showing a dot per line of the verbose build output. I think `-n` will carry over fine which is there to put all the dots on one long line. However that’ll means you should probably start with a `log -n ‘’` and then keep the echo’s as they are. That’ll provide the one logging prefix and append dots to that line.

itzg:
Ah, I didn’t know `docker logs` has that `-t` option. In that case, would filtering away the “init” lines be good enough to pick out the Minecraft server logs for your purposes?

start-deployBukkitSpigot
scriptcoded: I tried a few solutions and landed at 633b482. It requires adding flag parsing to the `log` function. I'm not quite sure if I'm very fond of the solution. What's your thoughts?

scriptcoded:
@itzg Just having `[init]` would be enough for my own purposes, yes. It might be nice for the sake of the project to have consistent logs though. But I think that for any solution that has to parse the logs, just having that `[init]` would be plenty good.

start-deployBukkitSpigot
itzg: Oh yeah, I see what you mean. I'm ok with your current approach. The only other idea I could think of would be to introduce a `logn` variant that does the `echo -n`. I'm not totally liking that either, but at least that would result in two little functions rather than one. more complicated, function.

scriptcoded: I could any. I you are okay with the `logn` variant I think I'd prefer it. Makes it more clear what's going on, especially for someone who isn't very used to bash. And it's probably less error prone.

itzg: Yes, please go with a `logn`. We’re almost there! :)

scriptcoded:
I think we're ready for this feature to be merged now. Sorry for making this take so long!

itzg:
> I think we're ready for this feature to be merged now. Sorry for making this take so long!

No problem at all. Thanks for working with me on this one.

itzg:
The updated image is now on Docker Hub.

PR #454, opened 2020-03-20
Allow to modify server-ip in server.properties

P4sca1:
Needed when your machine has multiple ip adresses assigned and you want to host your server on only one ip instead of allocating all available ips.

### DIFF
--- a/start-finalSetup04ServerProperties
+++ b/start-finalSetup04ServerProperties
@@ -43,6 +43,7 @@ function customizeServerProps {
   fi
 
   setServerProp "server-name" "$SERVER_NAME"
+  setServerProp "server-ip" "$SERVER_IP"
   setServerProp "server-port" "$SERVER_PORT"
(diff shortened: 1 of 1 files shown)


itzg:
Just to clarify, were you needing this when binding to the host network? FWIW, I always rely on the container IP isolation and then fine tune port and interface binding with the port binding.

P4sca1:
Yes, I am binding to the host network currently, because [this](https://github.com/moby/libnetwork/pull/2454) has not landed in Docker yet.
I agree that normally relying on container IP isolation would be sufficient, but supporting multiple ips with host networking doesn't hurt.

P4sca1:
Yes, I agree with that. Thank you for merging! :)

itzg:
Docker Hub is building now...

itzg:
...and it's now available from Docker Hub.

Issue #453, opened 2020-03-20
Log deletion


TimoMeijer:
Does the current image accumulate log files forever, or does it add a deletion policy?
I could not find this feature, but if it indeed isn't supported at the moment, I'll gladly open a PR to add a custom logging file.

itzg:
Very good question. Looking at one of my longer running containers I confirmed that it is at least doing its own log rotation, but not sure how long it's going to be keeping the oldest archive. 

Logging is being handled by the Minecraft server itself, so if you could research if/how to pass a JVM or other arg to the server to configure log rotation, then a PR would be great for that.

```
bash-5.0# ls /data/logs
2019-12-19-1.log.gz  2020-01-01-1.log.gz  2020-01-14-1.log.gz  2020-01-27-1.log.gz  2020-02-07-1.log.gz  2020-02-19-1.log.gz  2020-03-03-1.log.gz  2020-03-15-1.log.gz
2019-12-20-1.log.gz  2020-01-02-1.log.gz  2020-01-15-1.log.gz  2020-01-28-1.log.gz  2020-02-08-1.log.gz  2020-02-20-1.log.gz  2020-03-04-1.log.gz  2020-03-16-1.log.gz
2019-12-21-1.log.gz  2020-01-03-1.log.gz  2020-01-16-1.log.gz  2020-01-29-1.log.gz  2020-02-09-1.log.gz  2020-02-21-1.log.gz  2020-03-05-1.log.gz  2020-03-17-1.log.gz
2019-12-22-1.log.gz  2020-01-04-1.log.gz  2020-01-17-1.log.gz  2020-01-30-1.log.gz  2020-02-10-1.log.gz  2020-02-22-1.log.gz  2020-03-06-1.log.gz  2020-03-18-1.log.gz
2019-12-23-1.log.gz  2020-01-05-1.log.gz  2020-01-18-1.log.gz  2020-01-31-1.log.gz  2020-02-11-1.log.gz  2020-02-23-1.log.gz  2020-03-07-1.log.gz  2020-03-19-1.log.gz
2019-12-24-1.log.gz  2020-01-06-1.log.gz  2020-01-19-1.log.gz  2020-02-01-1.log.gz  2020-02-12-1.log.gz  2020-02-24-1.log.gz  2020-03-07-2.log.gz  latest.log
[...]

PR #455, opened 2020-03-20
Add customized log file with 7 day retention

TimoMeijer:
As discussed in issue #453

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,6 +60,7 @@ COPY mcstatus /usr/local/bin
 
 VOLUME ["/data","/mods","/config"]
 COPY server.properties /tmp/server.properties
+COPY log4j2.xml /tmp/log4j2.xml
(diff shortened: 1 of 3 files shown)


itzg:
Fixes #453

itzg:
Oops, thought github would auto-close with a comment like that :) . Docker Hub build is in-progress BTW.

itzg:
...and image is pushed.

itzg:
@TimoMeijer I wanted to let you know that I have needed to make the rolling log files an opt-in feature since it (along with my mc-server-runner) was [interfering with Forge color console](https://github.com/itzg/docker-minecraft-server/issues/820). It had already been found to cause high CPU issues with Spigot/Bukkit/Paper. so rather than blanket disable for various server types I added a new variable `-e ENABLE_ROLLING_LOGS=true` that opts into the feature.

Issue #471, opened 2020-03-30
Docker Status (Unhealthy)


DaFr0n:
Hi all

Iam having a issue regarding Docker health status, Server is running Forge. There does not appear to be any sever issues apart from the status

Failure count 39
failed to ping localhost:25565 : could not connect to Minecraft server: dial tcp [::1]:25565: connect: connection refused

**Portainer Stack:**

---
version: "2"
services:
  minecraft-server:
    image: itzg/minecraft-server
    container_name: minecraft
    network_mode: host
    environment:
      - PUID=100
      - PGID=1000
      - EULA=TRUE
      - VERSION=LATEST
      - MEMORY=2G
      - ANNOUNCE_PLAYER_ACHIEVEMENTS=true
      - TYPE=FORGE
      #- FORGEVERSION=14.23.5.2838
      - FORGE_INSTALLER=forge-1.12.2-14.23.5.2838-installer.jar
    ports:
      - 25565:25565
    volumes:
      - /srv/dev-disk-by-label-SSD120GB/SSD120GB/Appdata/Minecraft:/data
    restart: unless-stopped

Thanks in advance

wlp7s0:
Hello @DaFr0n ,
As you are using `network_mode: host`, docker container does not listen on localhost, but on server's external IP address. As for now - Docker health status is checking only localhost. Will create a pull request to fix the issue.

itzg:
Good catch @wlp7s0 . @DaFr0n are you sure host network mode was necessary. It is very rare I have found that mode was required.

I'm not familiar with Portainer, so perhaps that is an odd constraint it creates. In which case, you can disable healthchecks entirely in compose using:

```
healthcheck:
  disable: true
```

from https://docs.docker.com/compose/compose-file/#healthcheck

DaFr0n:
> Good catch @wlp7s0 . @DaFr0n are you sure host network mode was necessary. It is very rare I have found that mode was required.
> 
> I'm not familiar with Portainer, so perhaps that is an odd constraint it creates. In which case, you can disable healthchecks entirely in compose using:
> 
> ```
> healthcheck:
>   disable: true
> ```
> 
> from https://docs.docker.com/compose/compose-file/#healthcheck


Hi itzg
Apologies i am new to Docker and not to familiar with the network differences. I believe Portainer set this up automatically as a 'Host' network.

If i change to Minecraft_Default network it gives the below error:
**** FAILED TO BIND TO PORT!
[...]

PR #473, opened 2020-03-31
Checking health of the container using SERVER_IP variable

wlp7s0:
SERVER_IP variable will change docker's HEALTHCHECK command if it is set. #471


itzg:
During testing I found this isn't going to work since defaulting the env variable inside the startup script is too late. Healthchecks now fail with

```
                "Log": [
                    {
                        "Start": "2020-03-31T13:07:26.3312138Z",
                        "End": "2020-03-31T13:07:26.4429726Z",
                        "ExitCode": 1,
                        "Output": "failed to ping --port:25565 : could not connect to Minecraft server: dial tcp: lookup --port: no such host"
                    },
```

Perhaps set the default in the Dockerfile's `ENV` line
https://github.com/itzg/docker-minecraft-server/blob/4ad447ba6c736c250b2478a4dbd884abe6d15f3c/Dockerfile#L68-L74

wlp7s0:
I see, should I remove variable check from `start-finalSetup04ServerProperties` file?

start-minecraftFinalSetup
```diff
@@ -96,6 +96,10 @@ if [ -n "$JVM_DD_OPTS" ]; then
       done
 fi
 
+if [ -z "$SERVER_IP" ]; then
```
itzg, line 99: ...yes, go ahead and remove this block since it’s too late in the container lifecycle to be used by healthcheck

itzg:
I had to revert this change for now. This was causing all clients to fail to connect via `-p 25565:25565` port forwarding since the minecraft server process was only binding to localhost within the container. Maybe we should introduce a separate `HEALTHCHECK_SERVER_IP` variable.

PR #491, opened 2020-04-11
Enable upgrades with GENERIC_PACK

ChipWolf:
As discussed on Gitter, this commit changes the GENERIC_PACK feature to keep track of added files in a `manifest.txt` so that in the event of a change to the generic pack, the older version can be cleaned up to allow for the new pack to be introduced without older files causing issues.


itzg:
This change is now pushed to Docker Hub

Issue #492, opened 2020-04-12
/data working directory


rienafairefr:
In the dockerfile, WORKDIR is set as /data, but in many places you use hardcoded /data/xxx paths, and in other places not (e.g. writing `eula.txt`). This bit me (in a derived image) when I got a "permission denied" because of EULA writing, not realising that the process I had started started somewhere else than /data. Might be good to either use /data/xxx paths everywhere or assume the working directory of the process is /data everywhere ?

itzg:
`/data` is a very important assumption of all of the startup scripts. The `WORKDIR` is declared as a convenience and was not intended to ever be altered without being very sure about what you're doing. Can you clarify what you attempted that broke thing?

rienafairefr:
I didn't really alter WORKDIR, didn't alter your image, just added stuff to it, starting `/start` differently inside, and in doing so, the `/start` process started with another working directory than `/data`. It's in [this repo](https://github.com/rienafairefr/minecraft-on-demand) if you're curious (which is starting the minecraft server under `supervisord`).

itzg:
Gotcha. Very interesting solution you have there. 

So, you solved the issue here correct

https://github.com/rienafairefr/minecraft-on-demand/blob/master/src/supervisord.conf

I would prefer to avoid using relative directories throughout the startup scripts, so I’m not sure what else to do here.

rienafairefr:
The error in my case first happend on the EULA validation, [here](https://github.com/itzg/docker-minecraft-server/blob/1c2a0c506ce06b1ed5c9e785aaadb46bbd050829/start-configuration#L22), where the scripts writes eula.txt with `> eula.txt`, which assumes the working directory is /data, so I guess a `> /data/eula.txt` would have avoided the issue. Probably other examples in the scripts, I didn't look everywhere

itzg:
Ok, now I see what you mean. I’ll look through the scripts and see where I can be more consistent.

VoronkoM:
Hi all,

Having the same problem I think. The minecraft server folder is not created in the 
`\\QNAP\Container\container-station-data\lib\docker\volumes\******\_data`
And upon booting the container I get message:

>[init] Please accept the Minecraft EULA at                                                                                                                   
>[init]   https://account.mojang.com/documents/minecraft_eula                                                                                                 
>[init] by adding the following immediately after 'docker run':                                                                                               
>[init]   -e EULA=TRUE    

[...]

PR #493, opened 2020-04-15
add some /data prefixes in missing places

rienafairefr:
This is a tentative fix for #492, by adding a /data prefix. I probably missed some (`bash` is not my strong suit)


itzg:
This is now pushed to Hub.

PR #503, opened 2020-04-21
Allow specify paper build  and don't use legacy gamemode and difficulty

orblazer:
Hello,
This PR add environnent variable `PAPERBUILD` (default to `latest`) to specify wich paper build want download.
This PR also remove legacy usage of numeric `gamemode` and `difficulty`. From 1.14 this is legacy (source: https://minecraft.gamepedia.com/Java_Edition_18w48a)

Thanks

### DIFF
--- a/start-deployPaper
+++ b/start-deployPaper
@@ -3,9 +3,10 @@
 . /start-utils
 
 export SERVER=paper_server-${VANILLA_VERSION}.jar
+PAPERBUILD=${PAPERBUILD:-LATEST}
 if [ ! -f "$SERVER" ] || [ -n "$FORCE_REDOWNLOAD" ]; then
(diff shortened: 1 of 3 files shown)


start-deployPaper
```diff
@@ -3,9 +3,10 @@
 . /start-utils
 
 export SERVER=paper_server-${VANILLA_VERSION}.jar
+PAPERBUILD=${PAPERBUILD:-LATEST}
```
itzg, line 6: If you include `PAPERBUILD` in the filename then people can upgrade by bumping the variable. I'm thinking this would do it:
```suggestion
PAPERBUILD=${PAPERBUILD:-LATEST}
export SERVER=paper_server-${VANILLA_VERSION}-${PAPERBUILD}.jar
```

orblazer: Its an great idea

itzg:
The updated image is now available on Docker Hub.

Issue #463, opened 2020-03-26
Set the time for the container docker 


Bloopps:
Hi,
I have a time difference in my container logs, I can't figure out how to set the time for my time zone how can I do it?
Thanks for your work.

itzg:
If it's a timezone mismatch, then there are solutions like

https://forums.docker.com/t/synchronize-timezone-from-host-to-container/39116

If you're seeing this with Docker Desktop, this is/was an issue I kept seeing with the VM's clock drifting from the host clock `https://github.com/docker/for-win/issues/5593`

pauldadin:
It's resolved, I hadn't set my docker server time correctly.  
🤦‍♂

Bloopps:
It's resolved, I hadn't set my docker server time correctly.
🤦‍♂

Issue #263, opened 2018-12-05
MC server time 


haudankaivajasi:
Hi,

I have encountered a problem with the server time. Is there a way to make it follow the system time?
I have used the following line in a few other containers and it seems to be working in them but not on this one.
-v /etc/localtime:/etc/localtime:ro

 Is there any way to fix this?

itzg:
...oh, it's regarding the timezone? I thought you meant the clock itself gets out of sync from the host. It looks like I might need to be adding the `tzdata` package for it to work as you expect:

https://wiki.alpinelinux.org/wiki/Setting_the_timezone

itzg:
This fix is now pushed to Docker Hub. Let me know if this still doesn't resolve the issue for you.

itzg:
Yeah, when testing the before and after I confirmed it was using UTC previously. 

Yes, the container has to be recreated unfortunately to pick up a new image; however, if you have `/data` and such attached via volumes then it is safe to recreate the container.

haudankaivajasi:
Ok cool.
Pulled the new image and I'm restarting the server atm. The time in the console seems to still be 2 hours late :o should it be fixed or do I need to add another parameter or something?

itzg:
Hmm, no, the mounting of `/etc/localtime` like you had before should do it.

itzg:
To sanity test, can you try

```
docker pull itzgminecraft-server
docker run --rm -v /etc/localtime:/etc/localtime --entrypoint date itzg/minecraft-server
```

compared to

```
docker run --rm --entrypoint date itzg/minecraft-server
```

haudankaivajasi:
I ran the suggested lines and the first did return the correct date and time. but the latter and  when I create the server container, the time seems to be still off. The command I use is as follows:

> docker run -d -v /share/CACHEDEV1_DATA/Container/container_files/Minecraft:/data \
    -v /etc/localtime:/etc/localtime:ro \
    -e TYPE=FTB \
    -e FTB_SERVER_MOD=https://www.feed-the-beast.com/projects/ftb-infinity-evolved/files/2484486 \
    -e MEMORY=2G \
    -p 25565:25565 -e EULA=TRUE --name mc itzg/minecraft-server

itzg:
Thanks for running the sanity test commands -- your system definitely has and is running the updated image as I'd hope. It's then strange the real FTB container is not honoring the timezone setting for you. I ran a stripped down version of your command 

```
docker run -it --rm \
[...]

PR #510, opened 2020-04-27
Add explanation for timezone config on README

gabrielboliveira:
I searched on issue list to find how to do it (#263 and #463). Both ways worked.


itzg:
Thanks!

PR #512, opened 2020-04-28
Add possibility to use secrets in environment variables

orblazer:
Hello,
This PR add possibility to use secrets in environment variables.

**NOTE**: This could be adapted to server properties

Thanks.


itzg:

PR #519, opened 2020-05-03
Add `resource-pack-sha1` to server.properties

mrwacky42:
The `sed` command in
https://github.com/itzg/docker-minecraft-server/blob/master/start-finalSetup04ServerProperties#L16
does nothing if the line is not already in the file.

> [09:14:03] [Server thread/WARN]: You specified a resource pack without providing a sha1 hash. Pack will be updated on the client only if you change the name of the pack.


itzg:

Issue #490, opened 2020-04-11
start-deployBukkitSpigot, - mv: cant rename 'craftbukkit-*.jar': No such file or directory


v-mwalk:
In start-deployBukkitSpigot, line 23

docker run -d -it -v G:\docker\minecraft:/minecraft -v G:\docker\spigot:/data -e TYPE=SPIGOT -e BUILD_FROM_SOURCE=true  -e VERSION=1.15.2 -e EULA=TRUE -p 25565:25565 -p 25575:25575 -p 8123:8123 --name mc itzg/minecraft-server

jpcaldwell30:
Also having this issue.

jw-de:
The issue starts from version 1.14. Building 1.13.2 works as expected. This seems to be related to a change in the build tools. See the [Spigot Build Tools Wiki Page](https://www.spigotmc.org/wiki/buildtools/):

> You can find CraftBukkit (to but not including version 1.14) and Spigot in the same directory you ran the the BuildTools.jar in (craftbukkit-1.14.jar and spigot-1.14.jar).

itzg:
Thanks for the extra info @jw-de . I recently added a helper function to conditionalize by version, so I can branch the logic there.

Thomasolicious:
I still get this error when trying to run anything above 1.13.2. However I'm fairly new to the principles of github. Is this hotfix something that has been written in the source code of this image or does it require any input from my side when I try to run the docker image for 1.13.2+ to implement this hotfix?

itzg:
@Thomasolicious it's on my part to get that fix applied. I'll update that PR and merge it down...

itzg:
The latest image should resolve this.

Thomasolicious:
I'm still having the same issue, could this be due to something else?

I run this code for the docker:
``docker run -d -v /path/to/data:/data \``
``    -e TYPE=BUKKIT -e VERSION=LATEST \``
``    -e TZ=Europe/Amsterdam -e BUILD_FROM_SOURCE=true \``
``    -p 25565:25565 -e EULA=TRUE --name name itzg/minecraft-server --noconsole``

With ofcourse the path to the data being the specific path of my filesystem but I left that out of the above.

Response:
``[init] Changing ownership of /data to 1000 ...``
``[init] Running as uid=1000 gid=1000 with /data as 'drwxrwsr-x    1 1000     1000            16 May 12 18:30 /data'``
``[init] Resolved version given LATEST into 1.15.2``
[...]

PR #520, opened 2020-05-04
Short hotfix until version helper function is done

hejamu:
This hotfix helps with the different behaviour of BuildTools.jar in 1.14 and above. It's a bit sketchy, but without it, this docker is unusable with Spigot 1.14 or newer right now.

### DIFF
--- a/start-deployBukkitSpigot
+++ b/start-deployBukkitSpigot
@@ -20,7 +20,16 @@ function buildSpigotFromSource {
     cat /data/spigot_build.log
     exit 1
   fi
-  mv craftbukkit-*.jar /data/${SERVER}
+  if [  $VANILLA_VERSION != "1.15.2" -a \
+        $VANILLA_VERSION != "1.15.1" -a \
+        $VANILLA_VERSION != "1.15" -a \
+        $VANILLA_VERSION != "1.14.4" -a \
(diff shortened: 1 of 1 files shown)


start-deployBukkitSpigot
```diff
@@ -20,7 +20,16 @@ function buildSpigotFromSource {
...
   fi
-  mv craftbukkit-*.jar /data/${SERVER}
+  if [  $VANILLA_VERSION != "1.15.2" -a \
+        $VANILLA_VERSION != "1.15.1" -a \
+        $VANILLA_VERSION != "1.15" -a \
+        $VANILLA_VERSION != "1.14.4" -a \
+        $VANILLA_VERSION != "1.14.3" -a \
+        $VANILLA_VERSION != "1.14.2" -a \
+        $VANILLA_VERSION != "1.14.1" -a \
+        $VANILLA_VERSION != "1.14" ]; then
```
itzg, line 30: With the new [versionLessThan](https://github.com/itzg/docker-minecraft-server/blob/de866a322ffc55e93c27d055efe4937d58bc19ba/start-utils#L67), I think this could be done with
```suggestion
  if versionLessThan "1.14"; then
```

itzg:
Fixes #490

Issue #316, opened 2019-05-05
Update ARM Builds for Minecraft


nikhiljha:
I know this has been brought up before, but it looks like the `raspberrypi` tag isn't updated anymore. There are some really powerful ARM SBCs capable of running lots of Minecraft servers, and this image seems really well thought out. It would be nice if it was supported again.

nikhiljha:
I just tried to build the `Dockerfile` in `master` on my Odroid N2 and it built just fine. The images just need to be build on Docker Hub and this is already perfect :)

I uploaded it here for now: https://hub.docker.com/r/njha/minecraft

nikhiljha:
Actually just kidding, the 3 binaries: restify, rcon-cli, and mc-server-runner don't have aarch64 either. Gah.

nikhiljha:
Okay I got it to work. Here's what needs to be done.

1. Add "arm64" to targets in goreleaser for all three projects. (Can I suggest replacing "arm" with "arm64"? I don't think there are many (if any) non-arm64 ARM CPUs capable of running a Minecraft server that well anyway...)
2. Build and release.
3. ???
4. Profit.

itzg:
@nikhiljha thanks for doing the research on arm64 viability. Yeah, running the server on raspberrypi has proven to be pretty much unusable; however, I'll add the arm64 build rather than replace just so I can keep that option open. (Besides raspberrypi is what I have on my desk here to test with :) )

I'm going to keep a checklist here as I update the dependencies:

- [ ] restify
- [ ] rcon-cli
- [ ] mc-server-runner

nikhiljha:
<img width="787" alt="Screen Shot 2019-05-05 at 8 35 00 AM" src="https://user-images.githubusercontent.com/2773700/57196313-adf22c00-6f10-11e9-8b84-19c75a214931.png">

It works! Deployed in docker swarm as usual with no issues.

itzg:
Now my remaining challenge is blending the balena (formerly resin) cross-build support into the Dockerfile since Docker Hub builds via x86.

itzg:
@nikhiljha can you now test out the newly tagged image `itzg/minecraft-server:aarch64` ?

nikhiljha:
It doesn't seem to actually be `aarch64`? 

```
no suitable node (unsupported platform on 2 nodes)
```

itzg:
[...]

PR #536, opened 2020-05-20
upgrade base image

nikhiljha:
This should fix aarch64, see https://github.com/itzg/docker-minecraft-server/issues/316#issuecomment-631755949 and the comment before it.

As of yet untested, but a version bump is probably in order anyways.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8u212-jre-alpine
+FROM openjdk:8u252-jre-alpine
 
 LABEL maintainer "itzg"
 


nikhiljha:
Wait a second... https://stackoverflow.com/questions/61340138/why-new-java-8-images-published-by-openjdk-are-no-longer-alpine-based-but-debian

They no longer publish alpine images. That's a problem. 🤔 

Time to switch to OpenJ9 by default? :)

itzg:
For armv7 I already had to switch to an Ubuntu base image (for JIT support), so this might be the time to unify between the two branches. (Would make multiarch builds easier). I seem to remember several people had strange PATH issues with openj9 so I need to look back at what happened.

PR #571, opened 2020-06-30
Updated worldDest to use FTB_DIR

Kamaroth92:
FTB modpacks that have a top-level directory named for the modpack will result in extracted and cloned world data landing in the wrong location. This is fixed by extracting FTB worlds to FTB_DIR which is set during start-deployFTB.

### DIFF
--- a/start-finalSetup01World
+++ b/start-finalSetup01World
@@ -5,7 +5,7 @@ set -e
 isDebugging && set -x
 
 if [ $TYPE = "FEED-THE-BEAST" ]; then
-  worldDest=$FTB_BASE_DIR/$LEVEL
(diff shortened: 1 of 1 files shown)


itzg:
Can you point to an FTB modpack where this is the case? Your change seems valid, it's just confusing how some of these modpacks behave.

Kamaroth92:
This is an issue with https://www.curseforge.com/minecraft/modpacks/roguelike-adventures-and-dungeons but this hasn't been the only modpack I saw that's structured this way.

Strange behavior is spot on. This modpack in particular has an incorrect l launch script that I wish I could submit a pull request to the author for.

PR #580, opened 2020-07-10
Added 'FORCE_WORLD_COPY' environment variable

Trikolon:
Setting `FORCE_WORLD_COPY=TRUE` will overwrite the world on server start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -947,6 +947,9 @@ read-only volume attachment to ensure the clone source remains pristine.
 docker run ... -v $HOME/worlds:/worlds:ro -e WORLD=/worlds/basic
 ```
 
+### Overwrite world on start
+The world will only be downloaded or copied if it doesn't exist already. Set `FORCE_WORLD_COPY=TRUE` to force overwrite the world on every server start.
+
 ### Downloadable mod/plugin pack for Forge, Bukkit, and Spigot Servers
 
 Like the `WORLD` option above, you can specify the URL of a "mod pack"
--- a/start-finalSetupWorld
+++ b/start-finalSetupWorld
@@ -10,7 +10,7 @@ else
   worldDest=/data/$LEVEL
 fi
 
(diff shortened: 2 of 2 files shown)


Trikolon:
Wasn't sure where to mention this in the README. Feel free to move it  / let me know if you find a better fit.

start-finalSetupWorld
```diff
@@ -40,7 +40,7 @@ if [[ "$WORLD" ]] && [ ! -d "$worldDest" ]; then
     mv "$baseDir" "$worldDest"
```
itzg, line 40: This will need to switch to rsync also.

Oekn5w, line 40: I think that line fails, if the directory already exists

Oekn5w:
I think we should store the world that is to be synced / copied, in the case of a zip or download, we'd have to decompress / download every time, possibly overwriting the `WORLD` variable as long as the container lives

Trikolon:
> I think we should store the world that is to be synced / copied, in the case of a zip or download, we'd have to decompress / download every time, possibly overwriting the `WORLD` variable as long as the container lives

The tricky part is how to invalidate this cache. I agree that this could be optimized depending on the use case. But I think we can do that as follow up work if it's actually needed.

README.md
```diff
@@ -947,6 +947,9 @@ read-only volume attachment to ensure the clone source remains pristine.
 docker run ... -v $HOME/worlds:/worlds:ro -e WORLD=/worlds/basic
 ```
 
+### Overwrite world on start
```
itzg, line 950: Yep, this was a perfect spot for it.

itzg:
`latest` image is now pushed. I'll cut a new release, 1.5.0, and merge into other branches/tags.

itzg:
and released/merged https://github.com/itzg/docker-minecraft-server/releases/tag/1.5.0

PR #585, opened 2020-07-13
Add support for rcon password from file

code-monad:
Supporting RCON password from reading from file, which would cover the value of RCON_PASSWORD if it was defined.

works fine with Docker Secrets.

docker-compose Example:

```yaml
version: '3.7'
services:
  minecraft:
    image: test-mc
    ports:
      - "25565:25565"
    environment:
      EULA: "TRUE"
      VERSION: "1.12.2"
      TYPE: "FORGE"
      FORGEVERSION: "RECOMMENDED"
      CONSOLE: "false"
      ENABLE_RCON: "true"
      RCON_PASSWORD_FILE: /run/secrets/rcon_pass
    restart: unless-stopped
    secrets:
      - rcon_pass

secrets:
  rcon_pass:
    file: ./rcon_password # contents: blahblah
```


Dockerfile
```diff
@@ -69,7 +69,7 @@ WORKDIR /data
 ENV UID=1000 GID=1000 \
   JVM_XX_OPTS="-XX:+UseG1GC" MEMORY="1G" \
   TYPE=VANILLA VERSION=LATEST FORGEVERSION=RECOMMENDED SPONGEBRANCH=STABLE SPONGEVERSION= FABRICVERSION=LATEST LEVEL=world \
-  PVP=true DIFFICULTY=easy ENABLE_RCON=true RCON_PORT=25575 RCON_PASSWORD=minecraft \
+  PVP=true DIFFICULTY=easy ENABLE_RCON=true RCON_PORT=25575 RCON_PASSWORD=minecraft RCON_PASSWORD_FILE="" \
```
itzg, line 72: I am trying to minimize the number of variables pre-defined (even if empty) since it is overwhelming in Docker orchestration UIs, such as Portainer.

```suggestion
  PVP=true DIFFICULTY=easy ENABLE_RCON=true RCON_PORT=25575 RCON_PASSWORD=minecraft \
```

start-configuration
```diff
@@ -45,6 +45,26 @@ if [[ $PROXY ]]; then
...
 
+if [[ $RCON_PASSWORD_FILE ]]; then
+    log ""
+    if [ ! -e ${RCON_PASSWORD_FILE} ]; then
+	log "Initial RCON password file ${RCON_PASSWORD_FILE} does not seems to exist."
+	log "Please ensure your configuration."
+	log "If you are using Docker Secrets feature, please check this for further information: "
+	log " https://docs.docker.com/engine/swarm/secrets"
+	log "ENV value RCON_PASSWORD would be used."
+	log ""
```
itzg, line 56: I would prefer it exit immediately with an error since falling back to `RCON_PASSWORD` would be very misleading if someone intended to use `RCON_PASSWORD_FILE`

```suggestion
  log ""
	exit 1
```

start-configuration
```diff
@@ -45,6 +45,26 @@ if [[ $PROXY ]]; then
...
+    else
+	if [[ ${RCON_PASSWORD} ]]; then
+	    
+	    log "RCON_PASSWORD will be covered by values from password file ${RCON_PASSWORD_FILE}"
```
itzg, line 60: This log may be overly concerning since most users will probably just leave the default `RCON_PASSWORD`. I'd say just remove this conditional block.

code-monad:
> 
> 
> Great idea! Just a couple little things to fix, please.

I just fixed these. please check again.

PR #588, opened 2020-07-18
Use -prune for "REPLACE_ENV_VARIABLES_EXCLUDE_PATHS" feature.

Trikolon:
Using the -prune option to exclude paths in "find" should be a lot more efficient than -not since it prevents find from traversing excluded directories / files.

### DIFF
--- a/start-utils
+++ b/start-utils
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+function join_by { local d=$1; shift; echo -n "$1"; shift; printf "%s" "${@/#/$d}"; }
+
 function isURL {
   local value=$1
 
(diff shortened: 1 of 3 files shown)


start-utils
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+function join_by { local d=$1; shift; echo -n "$1"; shift; printf "%s" "${@/#/$d}"; }
```
itzg, line 3: Super cool!

itzg:
This is now pushed to Docker Hub

start-utils
chb0github:
```
 thing=(abc def 123)
$ printf "%s|" ${thing[@]} | sed 's/.$//'
abc|def|123
```
You've basically cramed 5 lines into 1. 

If you're case you could do
``` 
join(){ 
   local d=${1:?You need a delimiter} 
   printf "%s$d" "${@:1}" | sed 's/.$//'; 
}
```
the `?` in the variable assignment is an assertion - probably a good idea

PR #596, opened 2020-07-23
More filenames for modpack start scripts

### LINKED ISSUE #595: More startup script naming options
I appreciate the addition of the `USE_MODPACK_START_SCRIPT=false`, but it does not seem to solve the issue I've been having. When trying to update my ATM5 server, I get `Please make sure you are using the server version of the FTB modpack!`. After changing the startup script's name from `startserver.sh` to `ServerStart.sh`, the error goes away and the pack installs and starts correctly. Would it be possible to add more names to the startup script of the container? It looks like it could be added here: 
```
entryScriptExpr="
  -name ServerStart.sh
  -o -name ServerStartLinux.sh
  -o -name LaunchServer.sh
  -o -name server-start.sh
  "
```
(in `start-deployFTB`)

nrgbistro:
Added more names to possible modpack startup scripts. Fixes #595

### DIFF
--- a/start-deployFTB
+++ b/start-deployFTB
@@ -66,9 +66,12 @@ fi
 
 entryScriptExpr="
   -name ServerStart.sh
+  -o -name serverstart.sh
   -o -name ServerStartLinux.sh
   -o -name LaunchServer.sh
   -o -name server-start.sh
+  -o -name startserver.sh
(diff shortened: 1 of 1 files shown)


itzg:
This image is now available at Docker Hub

PR #599, opened 2020-07-26
Try to add caching of docker images

itzg:


itzg:
Github action fails to access that directory:

```
[warning]EACCES: permission denied, scandir '/var/lib/docker/image'
```

PR #646, opened 2020-10-11
IPv6 support for autopause

Paul1365972:
The AutoPause Feature currently only supports IPv4, this PR fixes just that. Tested and working as expected on a GCP Server. The regex for matching loopback addresses was taken from [Stackoverflow](https://stackoverflow.com/a/8426365/6500551), seems to work just fine, but I also manually checked some edge cases


itzg (approved):
Very cool. Thanks!

itzg:
This change is now available in the latest image.

Paul1365972:
Amazing

PR #659, opened 2020-11-02
Remove distro-provided JRE

MarcDoughty:
We don't need Java 8 from Ubuntu, we have 15 from AdoptOpenJDK.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@ LABEL org.opencontainers.image.authors="Geoff Bourne <itzgeoff@gmail.com>"
 RUN apt-get update \
   && DEBIAN_FRONTEND=noninteractive \
   apt-get install -y \
-    openjdk-8-jre-headless \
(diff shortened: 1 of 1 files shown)


james-d-elliott:
Keeping in mind java8 may be required in some situations (spigot + denizen plugin).

itzg:
Good comments but all is good. 

The mainline will remain at 8 to maximize compatibility. 

The targeted branch is named "multiarch-latest" since it derives the same Ubuntu base and multi-processor architecture support from "multiarch". 

The "adopt" variants are amd64 only and Alpine, so that they're smaller.

james-d-elliott:
Ah gotcha, wasn't 100% sure on the structure of the branches and hub images or that this was an adopt based Dockerfile, guess I should have read clearly. Nice change.

itzg (approved):
Sorry, I forgot to review and merge this :)

PR #674, opened 2020-12-03
Fix to get latest vanilla version

ramon1611:
I suppose that mojang has changed the structure of the version_manifest.json so that it's not possible to retrieve the latest version-manifest by setting $VANILLA_VERSION to "latest".

### DIFF
--- a/start-deployVanilla
+++ b/start-deployVanilla
@@ -9,7 +9,13 @@ export SERVER="minecraft_server.${VANILLA_VERSION// /_}.jar"
 if [ ! -e $SERVER ] || [ -n "$FORCE_REDOWNLOAD" ]; then
   log "Downloading $SERVER ..."
   debug "Finding version manifest for $VANILLA_VERSION"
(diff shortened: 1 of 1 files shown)


itzg (commented):
I'm not sure how it would arrive at `VANILLA_VERSION` containing "latest" since that is an internal variable that gets resolved at:

https://github.com/itzg/docker-minecraft-server/blob/a368f4cf573b6069030c1b55aa633977cc4303d3/start-configuration#L63

to be a specific version since the `.latest` section of the manifest looks like

```json
{
  "release": "1.16.4",
  "snapshot": "20w49a"
}
```

Can you describe more about how you observed that behavior?

itzg:
...re-reading your PR description I'm wondering if you're mistaking `VERSION` to be the same as `VANILLA_VERSION` where `VERSION` is the one for users to set to values like "latest", "snapshot", "1.16.4", etc.

ramon1611:
I've set `VERSION` to "latest" but it failed
Log said that it could not download the server file (404, not found)

but I've not seen that the other script always handles this behavior, so it is very strange that it fails to download :D
so I will try it again tomorrow...

itzg:
Yeah, that is strange. Can you provide the specific startup logs that lead up to the 404?

PR #683, opened 2020-12-11
Fixing some spigot world import issues

### LINKED ISSUE #682: Improve world finding in zip for spigot-based worlds
Spigot worlds are actually 3 different directories `world`, `world_nether`, and `world_the_end`. This is handled properly as far as copying them over via [finalSetupWorld](https://github.com/itzg/docker-minecraft-server/blob/master/start-finalSetupWorld#L46-L51), but above that it tries to find the world directory and will pick the first it finds. This has to run once for a spigot world to see if it selects the proper world and then `WORLD_INDEX` must be used to select the actual world of the 3. This seems troublesome and something that the selection code could work around. I thought about adding a `grep -v`, but got concerned that maybe there is someone out there that named their world `world_nether` for reasons and has a `world_nether_nether` directory.

I'm not sure the proper script change here, but it was a sticking point for me so I thought I would document it.

hyperbolic2346:
Spigot worlds were incorrectly identified as multiple worlds and when copying the worlds the code would not copy the nether and end properly. This change will hide the nether and end worlds from the multiple world search and also copy the nether and end if it exists.

There is existing code that I thought did the latter on first glance, but it is actually to convert a world *to* a spigot world only.

Fixes #682

### DIFF
--- a/start-finalSetupWorld
+++ b/start-finalSetupWorld
@@ -26,7 +26,7 @@ if [[ "$WORLD" ]] && ( isTrue "${FORCE_WORLD_COPY}" || [ ! -d "$worldDest" ] ); 
     (cd /tmp/world-data && unzip -o -q "$zipSrc")
 
     if [ "$TYPE" = "SPIGOT" ]; then
-      baseDirs=$(find /tmp/world-data -name "level.dat" -exec dirname "{}" \;|grep -v '_nether\|_the_end')
+      baseDirs=$(find /tmp/world-data -name "level.dat" -not -path "*_nether*" -not -path "*_the_end*" -printf "%h")
     else
       baseDirs=$(find /tmp/world-data -name "level.dat" -exec dirname "{}" \;)
     fi
--- a/start-finalSetupWorld
+++ b/start-finalSetupWorld
@@ -25,7 +25,12 @@ if [[ "$WORLD" ]] && ( isTrue "${FORCE_WORLD_COPY}" || [ ! -d "$worldDest" ] ); 
     mkdir -p /tmp/world-data
     (cd /tmp/world-data && unzip -o -q "$zipSrc")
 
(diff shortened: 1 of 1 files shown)


start-finalSetupWorld
```diff
@@ -25,7 +25,12 @@ if [[ "$WORLD" ]] && ( isTrue "${FORCE_WORLD_COPY}" || [ ! -d "$worldDest" ] );
     mkdir -p /tmp/world-data
     (cd /tmp/world-data && unzip -o -q "$zipSrc")
 
-    baseDirs=$(find /tmp/world-data -name "level.dat" -exec dirname "{}" \;)
+    if [ "$TYPE" = "SPIGOT" ]; then
+      baseDirs=$(find /tmp/world-data -name "level.dat" -exec dirname "{}" \;|grep -v '_nether\|_the_end')
```
itzg, line 29: This code seems totally fine, but am wondering if it would work and streamline it do filter within `find` using the following. While looking that up I noticed the `-printf` action might be handy. So together could be:

```suggestion
      baseDirs=$(find /tmp/world-data -name "level.dat" -not -path "*_nether*" -not -path "*_the_end*" -printf "%h")
```

itzg (commented):
Looks good as it is, but curious to squeeze in some "find" optimization there.

itzg (approved):
Awesome and thanks for the `find` tweak

qnten:
When using the newest docker image containg this fix, the world download fails:
```
[init] Resolved version given LATEST into 1.16.4
[init] Resolving type given PAPER
[init] Downloading PaperMC 1.16.4 (build 323) ...
[init] Unzipping world
find: unrecognized: -printf
BusyBox v1.29.3 (2019-01-24 07:45:07 UTC) multi-call binary.
```

I think the `-printf` flag isn't supported in alpine (https://github.com/google/cadvisor/issues/1556).

itzg:
I'm sorry, I should have tested that also myself. Fix is on the way...

PR #688, opened 2020-12-13
feat: improve REMOVE_OLD_MODS option

### LINKED ISSUE #662: feat: option to clear only plugins/mods inside the data directory
Currently there is only the `REMOVE_OLD_MODS` flag that will clear everything inside the plugins/mods directory.

However this is not always what we want. Clearing everything would also remove configurations and flat file data, when we only need to remove old plugins that no longer exist inside the `/plugins` mount.

I am willing to do a PR on this, if ok?

Silthus:
A `REMOVE_OLD_MODS_INCLUDE`, `REMOVE_OLD_MODS_EXCLUDE` and `REMOVE_OLD_MODS_DEPTH` option can now be specified to filter the plugins or mods that should be deleted.

`REMOVE_OLD_MODS_INCLUDE` defaults to `*` which mirrors the current behauvior of deleting everything inside the `plugins` or `mods` directory.

Closes #662

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,13 +301,13 @@ or downloading a world with the `WORLD` option.
 
 There are two additional volumes that can be mounted; `/mods` and `/config`.
 Any files in either of these filesystems will be copied over to the main
-`/data` filesystem before starting Minecraft. If you want old mods to be removed as the `/mods` content is updated, then add `-e REMOVE_OLD_MODS=TRUE`.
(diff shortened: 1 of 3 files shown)


itzg (approved):
Very nice!

Silthus:
Thank you for approving the PR!

One question: could you please run the update script that publishes the image with all the different java versions?

Also maybe [semantic-release](https://github.com/semantic-release/semantic-release) would be good think to look into. That could automate your complete release process, label the images with a semantic version and publish them with different flavors to the hub.

itzg:
I'll think more about using semantic-release or something similar. The image tags are built from branches, which are "released" with the [docker-versions-create.sh](https://github.com/itzg/docker-minecraft-server/blob/master/docker-versions-create.sh) script. Sometimes there are merge conflicts if I'm adding features specific to a non-mainline tag, so up to this point it's a semi-manual process to do releases.

I can kick off a release after #689 , so both of them get picked up.

Issue #664, opened 2020-11-10
feat: copy server configs into server dir (analogue to bungee)


Silthus:
Currently there is a `/config` mount that says it copies the server configs into the data dir. However after inspecting the source it only copies the data to `/data/config`.

https://github.com/itzg/docker-minecraft-server/blob/master/start-minecraftFinalSetup#L67

There is a missing option to copy all server configs into the root server dir. Currently every config needs to be mounted directly as 1:1 files.

However I am not sure what the purpose of the `/data/config` mount is. If this is intended another mount and copy process like `/bukkit` or something would be good.

itzg:
Admittedly, the wording there could be better to clarify the destination path. Too many users probably depend on the existing behavior, so that particular path and logic can't be changed at this point.

What files do you find need to be managed the way you described? The startup script already manages some of them, then there's generic pack support for the remaining content and then world data support.

Silthus:
Maybe I am just overlooking something. I have the following `docker-compose.yml` where I want to configure the `spigot.yml` and `bukkit.yml` files. However I could not find a way to copy them into the container without the volume mount. The problem is that those update a timestamp in their header which then gets detected by git as changed.

```yml
version: '3.7'
services: 
  main:
    image: itzg/minecraft-server:adopt11
    command: --world-dir=/data/worlds
    user: ${CURRENT_USER}
    environment:
      LEVEL: world
      SERVER_NAME: main
    volumes:
      - ./servers/main:/data
      - ./plugins:/plugins
      - ./worlds:/data/worlds
      - ./configs/main/server.properties:/data/server.properties
      - ./configs/main/bukkit.yml:/data/bukkit.yml
      - ./configs/main/spigot.yml:/data/spigot.yml
      - ./configs/main/paper.yml:/data/paper.yml
      - ./dynmap/main:/data/plugins/dynmap/web
      - /etc/timezone:/etc/timezone:ro
    env_file:
      - secrets.env
      - properties.env
    depends_on:
      - db
    networks:
      - minecraft
networks:
  minecraft:
    driver: bridge
```

itzg:
Thanks! That helped to clarify what top level files you meant.

The `GENERIC_PACK` option [in code here](https://github.com/itzg/docker-minecraft-server/blob/master/start-finalSetupModpack#L141) is apparently not documented, but also doesn't help your case much since it assumes a zip file currently.

[...]

PR #689, opened 2020-12-15
feat: add COPY_CONFIG_DEST option

Silthus:
A new `COPY_CONFIG_DEST` option now allows you to specifcy where the files in `/config` should be copied to.
Defaults to `/data/config` to mirror the old behaviour.

Closes feat: copy server configs into server dir (analogue to bungee) #664


itzg (approved):
Very nice and including the example in the docs update helps a lot.

itzg:
1.10.0 releases are building now to pick up this and #688 on all the branches / image tags.

Issue #700, opened 2020-12-26
Enviroment variables are not replaced on files copied through the config mount


Misterio77:
Hey!

I've been trying to get some variables replaced on config files on a paper server, but it seems that the step that copies from `/config` mount [only occurs at the last setup step](https://github.com/itzg/docker-minecraft-server/blob/master/start-minecraftFinalSetup#L63), which is right after the [one that replaces env vars](https://github.com/itzg/docker-minecraft-server/blob/master/start-finalSetupEnvVariables).

I don't think this is intended? I thought of two different solutions:

1.  Move Env vars step to [start-minecraftFinalSetup](https://github.com/itzg/docker-minecraft-server/blob/master/start-minecraftFinalSetup)
2. Move the copy step to its own file, and make that run before finalSetupEnvVariables

I think the latter is a bit cleaner, so i'll try and write a PR for it, if that's okay?

itzg:
Yes, option 2 makes sense and a PR would be great.

PR #701, opened 2020-12-26
Envvars in files from `/config` and `/mods` now replaced as expected

Misterio77:
The operation that copies files from `/config` and `/mods` mounts were previously executed in the final step, so env vars located in files copied from these mounts did not get replaced (#700).

Moved these operations to the file which previously contained just the operation that copies `/plugins`.

This file was renamed to "start-finalSetupMounts", to get all these together.


Misterio77:
I also added `MAGMA` to the types on which plugins should be copied from `/plugins`

itzg:
I have been wanting to create a diagram to document the flow the scripts. So, I hope you don't mind that I'll first commit the pre-PR flow and then an update to the diagram with your PR changes.

itzg (approved):
Much cleaner organization. Thanks!

PR #708, opened 2020-12-30
Fixing AUTOPAUSE on Raspberry Pi 4s

ErstBlack:
Adding net-tools, replacing killall with pkill, and replacing ps -a with ps -ax

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get update \
     imagemagick \
     gosu \
     sudo \
+    net-tools \
     curl wget \
     jq \
     dos2unix \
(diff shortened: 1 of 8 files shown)


ErstBlack:
So it looks like there's still an issue with the resume functionality, but that seems more to do with the fact that there's a handful of interfaces which the [java_clients_connected](https://github.com/itzg/docker-minecraft-server/blob/master/files/autopause/autopause-fcns.sh#L19) call is missing out on, probably because I'm running it in a stack rather than a standalone container.

itzg (approved):
LGTM. @Oekn5w , any requests?

Oekn5w:
also fails to resume for me.
To the matter of the PR: missed a `ps` in `health.sh` and `killall` in `files/sudoers-mc` , `pkill`resides in `/usr/bin/pkill`

Oekn5w:
@ErstBlack replacing `/sbin/su-exec` with `gosu` fixes also the resuming part of the functionality.

Oekn5w:
looks good, thanks for troubleshooting and fixing this

itzg (approved):

PR #711, opened 2021-01-01
Autopause update: allow network interface selection

Oekn5w:
The initial iteration of the functionality failed when the `eth0` interface was not available, mostly in non-default network configurations. Now, the interface is selectable via environment variable. With this change, I think we can remove the experimental tag from the feature.

Also, the container isn't stopped anymore, if the autopause fails to start up. The issues around this feature were barely traceable because the container didn't really run. Instead, a message is written periodically to catch the user's attention.


Oekn5w:
@itzg the alpine `ps` version does not support the x flag. The merge to the `multiarch` branches from `docker-create-versions.sh` will fail with conflicts. I opted against also creating a PR for multiarch because once you've manually resolved the conflicts, future merges should work without problems. During testing, I did merge and build the multiarch amd64 image, the changes (adapted according to what we saw the last few days) also worked there.

README.md
```diff
@@ -180,12 +180,10 @@ healthy
 
 Some orchestration systems, such as Portainer, don't allow for disabling the default `HEALTHCHECK` declared by this image. In those cases you can approximate the disabling of healthchecks by setting the environment variable `DISABLE_HEALTHCHECK` to `true`.
 
-## Autopause (experimental)
+## Autopause
```
itzg, line 183: Are you totally comfortable removing the "experimental" designation? I personally don't have the experience with it to support it, so I'll continue to delegate to you if you don't mind.

README.md
```diff
@@ -206,7 +206,7 @@ Enable the Autopause functionality by setting:
 -e ENABLE_AUTOPAUSE=TRUE
 ```
 
-There are 4 more environment variables that define the behaviour:
+There are 5 more environment variables that define the behaviour:
```
itzg, line 209: That's when I change my wording to be like
```suggestion
The following environment variables define the behaviour of auto-pausing:
```

itzg (commented):
Changes look great. Yes, I'll take care of any merge conflicts there and then it'll be smooth merging from there.

README.md
Oekn5w: From what I gathered, after resolving these interface problems and the multiarch tag, this should work now. I'm happy to provide support if needed. We could retain the experimental tag, because this is in no way supported / intended by Mojang if you want.

README.md
Oekn5w: Fair enough, done

itzg (approved):
Thanks!

PR #719, opened 2021-01-09
fix/681

Dubhar:
Remove all PaperMC jars that are not needed.

How it works:

- Use the filename to be downloaded/executed as starting point e.g. _SERVER=paper-1.16.4-392.jar_
- In this filename replace the build number with a wildcard for any build number e.g. _paperJarSearchString=paper-1.16.4-[0-9]*.jar_
- Find and delete all files matching the $paperJarSearchString except the one file named $SERVER
- Also print which files were deleted

### DIFF
--- a/start-deployPaper
+++ b/start-deployPaper
@@ -56,6 +56,10 @@ else
     zarg=(-z "$SERVER")
   fi
 
+  log "Removing old PaperMC versions ..."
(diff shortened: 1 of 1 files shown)


itzg (approved):
Very nice! Thanks for contributing this.

itzg:
This has been branched out and pushed to Docker Hub

PR #722, opened 2021-01-12
Fix autopause multiarch interface selection

Oekn5w:
The fix on master (#711) didn't merge correctly into the ubuntu-based image.


itzg (approved):
Thanks, I was afraid I had misinterpreted your merge guidance. I should have created a PR for you to review.

Oekn5w:
In hindsight we'd have been better off if we had a folder for the autopause alpine and the autopause ubuntu stuff on the repo folder structure with the corresponding line changed in the Dockerfiles. But I think, this is now feature-complete.

PR #723, opened 2021-01-13
Add an option to override the icon which already has been set.

eugeneindigo:
Initially thank you for this awesome work that you've created. It brings joy to me and my family.

I am hosting my server on AWS and I have realized that I cannot change the icon that I've initially set. I have made a change to allow the server icon to be overridden.


itzg (approved):
Very nice! Thanks.

itzg:
This change is now available on Docker Hub

PR #741, opened 2021-01-31
Document LEVEL_TYPE: BIOMESOP

MCOfficer:
This can only be found by digging through issues (#156), so it would make sense to mention it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -959,7 +959,7 @@ environment variable set to `false`, such as
 ### Level Type and Generator Settings
 
 By default, a standard world is generated with hills, valleys, water, etc. A different level type can
-be configured by setting `LEVEL_TYPE` to an expected type, such as
+be configured by setting `LEVEL_TYPE` to an expected type, for example
 
 - DEFAULT
 - FLAT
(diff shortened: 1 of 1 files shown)


MCOfficer:
Turns out that BoP post-1.15 no longer needs `BIOMESOP` but `BIOMESOPLENTY`. Would you like me to add another exception, or just remove the check entirely?

itzg:
That PR you linked it so old that the validation of `LEVEL_TYPE` was long since removed. It just converts to upper case and stores whatever value the use gives:

https://github.com/itzg/docker-minecraft-server/blob/3de2bf88df93a0ef8ef41830dafb22ca6b45962e/start-finalSetupServerProperties#L76

README.md
```diff
@@ -967,6 +967,7 @@ be configured by setting `LEVEL_TYPE` to an expected type, such as
 - AMPLIFIED
 - CUSTOMIZED
 - BUFFET
+- BIOMESOP (if Biomes O' Plenty is installed)
```
itzg, line 970: Since this list only serves as examples now, you could clarify the `BIOMESOPLENTY` you found for 1.15+

MCOfficer:
> That PR you linked it so old that the validation of `LEVEL_TYPE` was long since removed. It just converts to upper case and stores whatever value the use gives:

Wow, I really should've checked the master branch :laughing: 

Other than adding BIOMESOPLENTY, i also took the liberty to reword it to "for example", because i really thought that list was something akin to an enum.

itzg:
> Other than adding BIOMESOPLENTY, i also took the liberty to reword it to "for example", because i really thought that list was something akin to an enum.

Thanks. Yes, in your defense the section was indeed still worded as if it was an enum of allowed values.

itzg (approved):
Thanks for updating that section.

PR #803, opened 2021-03-13
Fix typo in example docker-compose

miki164:
After copy pasting the docker-compose error occured:
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.mc.volumes contains an invalid type, it should be an array

I've just add "-" and should work now.


itzg (approved):
Thanks!

PR #831, opened 2021-04-12
COPY_MODS_DEST env variable 

zjael:
Using curseforge modpacks creates a FeedTheBeast folder in /data which contains the /mods folder, this env variable would allow to copy mods to that folder properly.


itzg (approved):
Thanks!

PR #833, opened 2021-04-18
Replace env variables in *.hjson and *.json

CDFN:


itzg (approved):
Great addition! I'm curious though, what's using hjson?

Siphalor:
Not sure if this is related but I use Hjson for my mods :>

PR #850, opened 2021-04-28
Support FTP servers for file downloads

### LINKED ISSUE #849: Support downloads from FTP servers
Right now when using an ftp server url (e.g. `ftp://192.168.1.10/modpack.zip`) an error is thrown that the URL is not valid. However, `curl` supports downloads from an FTP server, so there shouldn't be a reason that an URL starting with `ftp` is not valid. I considered creating a PR, but don't want to change the `isURL` function as it might be used in some places that require HTTP. Maybe split into `isHttpUrl` and `isFtpUrl` ?

RouNNdeL:
Since `curl` can natively download files from FTP servers, given a `ftp://` URL, support those URLs. Closes #849


itzg (approved):
Thanks!

PR #858, opened 2021-05-04
Support local paths for modpacks / mods

picapi:
I noticed while testing some things with this image that local paths were available for using Worlds, but not for Mods or Modpacks.
I spent a little bit of time trying to implement this in my own branch, and think it would be useful upstream.

I understand that the usecase can also be served through utilising volumes, however having similar functionality for these environment variables may aid usability.
I'm also unsure whether the checking of paths is sufficient - this will check whether a path ends in the correct extension right now, but not much more - although we do also check whether the command succeeds.


itzg (approved):
I like it! It's great to have the consistency across the variables.

Issue #864, opened 2021-05-12
Java 16 images required for 21w19a onwards


Stealthii:
With Mojang's change to using Java 16 in snapshot 21w19a, images will need prepared with OpenJDK 16 in order to run instances under this and future versions.

As the game is compiled with Java 16, and has a class file version of 60, Java 15 and below will not execute these jars.

itzg:
README has been updated and build is in progress

https://github.com/itzg/docker-minecraft-server#running-minecraft-server-on-different-java-version

itzg:
`itzg/minecraft-server:java16` is now available. I'm doing a quick test myself with that image and `VERSION=SNAPSHOT`.

m4lfunction:
> `itzg/minecraft-server:java16` is now available. I'm doing a quick test myself with that image and `VERSION=SNAPSHOT`.

Just tested, seems to be working fine.

Stealthii:
Can confirm this works as expected under type `FABRIC` and `VANILLA`. Happy to close this!

Panderine:
Hi! It's me again the openj9 guy. Would you be so kind to add java16-openj9 version? I think you can remove java15-openj9. Thanks again for everything!!!

PR #865, opened 2021-05-12
Bump java version to 16 to support snapshot 21w19a onwards

wabscale:
When I restarted my snapshot server this morning to get the latest 21w19a this morning I got this:

```
minecraft_1  | [init] Running as uid=1000 gid=1000 with /data as 'drwxrwxr-x    4 1000     1000          4096 May 12 14:26 /data'
minecraft_1  | [init] Resolved version given SNAPSHOT into 21w19a
minecraft_1  | [init] Resolving type given VANILLA
minecraft_1  | [init] server.properties already created, skipping
minecraft_1  | [init] Checking for JSON files.
minecraft_1  | [init] Setting initial memory to 16G and max to 16G
minecraft_1  | [init] Starting the Minecraft server...
minecraft_1  | Error: LinkageError occurred while loading main class net.minecraft.server.Main
minecraft_1  | 	java.lang.UnsupportedClassVersionError: net/minecraft/server/Main has been compiled by a more recent version of the Java Runtime (class file version 60.0), this version of the Java Runtime only recognizes class file versions up to 55.0
minecraft_1  | 2021-05-12T14:26:21.848Z	WARN	mc-server-runner	sub-process failed	{"exitCode": 1}
minecraft_1  | 2021-05-12T14:26:21.848Z	INFO	mc-server-runner	Done
minecraft_minecraft_1 exited with code 1
``` 

Bumping the base image to `adoptopenjdk/openjdk16:alpine-jre` fixed this issue.

It looks like this snapshot version, and likely the next major releases will require this more recent version: https://www.minecraft.net/en-us/article/minecraft-snapshot-21w19a

I realize that there may be some other places in this repository that may need testing for java 16, so please let me know if I've missed anything major.

Addresses https://github.com/itzg/docker-minecraft-server/issues/864

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM adoptopenjdk/openjdk11:alpine-jre
+FROM adoptopenjdk/openjdk16:alpine-jre
 
 LABEL org.opencontainers.image.authors="Geoff Bourne <itzgeoff@gmail.com>"
 


Stealthii:
This might be better tackled as a new branch `adopt16`, and docker tags for `java16` until general support for Java 16 is guaranteed in the various variants of Minecraft Java server (spigot-like, sponge, forge, etc.). Indeed 21w19a requires jre16, but many current server variants have issues on jre14 - some forge-like servers still won't work on jre11 and require jre8 (although these are considered legacy and the `java8` tag exists for those requirements).

Java 11 remains as the most compatible JRE available for Minecraft servers right now.

wabscale:
Yes I was thinking that kind of approach would be better at least until the latest major release requires java 16.

I see you have these alternative builds as separate branches. Would it maybe make more sense to make this a docker build arg? You would be maintain all your alternative java versions in a single branch. It would looks something like this:

```
ARG JAVA_VERSION=openjdk11
FROM adoptopenjdk/${JAVA_VERSION}:alpine-jre
```

Then you could just have a few of these (or more cleanly in a loop):

```shell
docker build -t itzg/minecraft-server:latest .
docker build -t itzg/minecraft-server:adopt15 --build-arg JAVA_VERSION=openjdk15 .
docker build -t itzg/minecraft-server:adopt16 --build-arg JAVA_VERSION=openjdk16 .
```

itzg:
I'm wondering if that's a mistake on Mojang's part. I think they're bundled Java 8 with their launcher. 

In any case, I appreciate the PR but let me manage the alternate Java base images since I have a CI methodology setup for all that. 

I'll get a `java16` tag out shortly.

itzg:
> Would it maybe make more sense to make this a docker build arg? 

It's not that simple for various reasons. I have created a complexity for myself with a mix of Alpine and Debian base images. Even then some of the variants don't play nice version by version, so the branches enable a nice workflow for myself to manage all the image variants but leverage git to customize where needed.

wabscale:
Awesome, thanks for the quick turnaround. I think you should also leave a note about this 21w19a java 16 business in the versions section of the readme: https://github.com/itzg/docker-minecraft-server/blob/master/README.md#versions

itzg:
Yeah, I stand corrected

https://www.minecraft.net/en-us/article/minecraft-snapshot-21w19a

I'm very surprised they went with a non-LTS Java version.

itzg:
> I think you should also leave a note about this 21w19a java 16 business in the versions section of the readme: https://github.com/itzg/docker-minecraft-server/blob/master/README.md#versions

Yeah, I think you're right. I should send a bill to Mojang/Microsoft for that extra bit of work 😉

wabscale:
Ha yes. It is weird that they are going with this seemingly unnecessary breaking change :thinking:

itzg:
My bet is that they'll revert in the next snapshot back to Java 11 or 8.

itzg:
...in any case, README has been updated

https://github.com/itzg/docker-minecraft-server#running-minecraft-server-on-different-java-version

and will get the branch/build going shortly.

itzg:
Thank you again for the PR. I'll close out and am posting updates to #864

PR #872, opened 2021-05-13
Updated examples to version 3.8

W4ff1e:


itzg (commented):
Thanks! There's also some compose file snippets within the README.md

itzg (approved):
Thanks!

basisbit:
Why was this change done? There should be no reason to increase the minimal required docker-compose version, while not using any featureset of the newer docker-compose version. This however makes using this repository more difficult on Linux distributions who aren't that current with their software repositories (like LTS Ubuntu or Debian).

itzg:
@basisbit these files serve only as **examples**. The intent is that the end user modify them as appropriate. BTW I (and Docker) recommend avoiding the distro supplied docker packages and instead use the docker.io repo for your package manager. 

With that said, I do realize now that the "lower" version indicator was correct in saying that the declarations therein are compatible down to that specification version.

@W4ff1e , I vaguely remember there was a specific compose feature that prompted the change. Do you remember if that was the case?

basisbit:
> BTW I (and Docker) recommend avoiding the distro supplied docker packages and instead use the docker.io repo for your package manager.

afaik this only applies for docker, but not for docker-compose, these should be fine from the distro repositories and docker-compose is not included in docker.io repository (or was not the last time I checked).

The samples as well as stuff that is described in the documentation all works fine with docker-compose version 3.6.
3.7 and 3.8 probably didn't add anything relevant to docker-minecraft-server use case, see https://docs.docker.com/compose/compose-file/compose-versioning/#version-37

W4ff1e:
@itzg it was for setting the health timeout on lower end linux machines

itzg:
> afaik this only applies for docker, but not for docker-compose, these should be fine from the distro repositories and docker-compose is not included in docker.io repository (or was not the last time I checked).

You're partly right, it's not included in the packages (not sure why, but probably because it's a python package). You have to "curl install" it separately:

https://docs.docker.com/compose/install/

So, you may not be as constrained by the distro as you thought you were.

PR #873, opened 2021-05-14
fix helm chart links in README

gilesknap:


itzg (approved):
That's embarrassing I hadn't changed that yet. Thank you so much for fixing it.

PR #898, opened 2021-05-27
fix: update Sky Factory version link

niekcandaele:
SF 4.1.0 doesn't work with the latest Forge anymore. 4.2.2 does.


itzg (approved):
Thanks. I probably shouldn't have such a specific link in the docs in the first place.

PR #901, opened 2021-05-30
fix bad find statement

speedenator:
After working w/ a CurseForge mod, I tracked down a bug in the find statement that when the forge server is in */FeedTheBeast/mods/forge-<version>.jar, the find statement to locate it won't actually work due to the path */mods/*. Removed the leading * (also from the * in front of libraries) and added in a maxdepth (forge installer will also make a copy in /libraries/net/minecraftforge/forge/<version>/).

### DIFF
--- a/start-deployCF
+++ b/start-deployCF
@@ -91,7 +91,9 @@ if ! isTrue ${USE_MODPACK_START_SCRIPT:-true}; then
     echo "${FTB_SERVER_MOD}" > $installMarker
   fi
 
-  export SERVER=$(find ${FTB_BASE_DIR} -type f \( -path "*/libraries/*" -o -path "*/mods/*" \) -prune -o -name "forge*.jar" -not -name "forge*installer.jar" -print)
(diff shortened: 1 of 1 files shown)


start-deployCF
```diff
@@ -91,7 +91,9 @@ if ! isTrue ${USE_MODPACK_START_SCRIPT:-true}; then
     echo "${FTB_SERVER_MOD}" > $installMarker
   fi
 
-  export SERVER=$(find ${FTB_BASE_DIR} -type f \( -path "*/libraries/*" -o -path "*/mods/*" \) -prune -o -name "forge*.jar" -not -name "forge*installer.jar" -print)
+  # change from */mods/* to /mods/* as leading * makes things fail
+  # and add maxdepth 2
```
itzg, line 95: You can remove these inline comments explaining the change since the PR will provide that history.

itzg (commented):
It seems like your change is for the better, but I want to look back at past PRs / changes in this area because it seems like one change will help some modpacks, but break others.

PR #917, opened 2021-06-09
Support Java versions above 11 for Airplane

Jawa-Juice:
This PR enables the use of Airplane on Java versions above 11, in anticipation of 1.17.


itzg (approved):
Excellent! Thanks.

PR #922, opened 2021-06-16
Workaround: Change default RELEASE of Yatopia

basisbit:
api.yatopiamc.org/v2/stableBuild API has been broken for a couple of months (see https://github.com/YatopiaMC/Yatopia/issues/488 ) and it looks like it won't be fixed soon. Thus, changed the default RELEASE value from stable to latest, so that Yatopia can still be easily set up without explicitly specifying a RELEASE env.


itzg (approved):
Thanks!

PR #936, opened 2021-06-24
Correct AirplanePurpur branch name

Jawa-Juice:
Of course branch names are case sensitive.. stupid me.


itzg (approved):
Ah, that makes sense.

PR #937, opened 2021-06-26
Allow to enable the whitelist without modifying the whitelist file

P4sca1:
Currently there is no way to just enable the whitelist without adding a player to the whitelist file.
This change allows to set `WHITELIST=true` to set the server properties accordingly without modifying the whitelist file.
Players can then be added / removed from the whitelist using the `whitelist add` and `whitelist remove` commands.
For convenience `WHITELIST=false` will now disable the whitelist and has the same behaviour like `WHITELIST=` (empty string).

Signed-off-by: Pascal Sthamer <sthamer.pascal@gmail.com>

### DIFF
--- a/start-finalSetupServerProperties
+++ b/start-finalSetupServerProperties
@@ -27,7 +27,7 @@ function setServerProp {
 }
 
 function customizeServerProps {
-  if [ -n "$WHITELIST" ]; then
+  if [ -n "$WHITELIST" && "$WHITELIST" != "false" ]; then
     log "Creating whitelist"
     setServerProp "whitelist" "true"
     setServerProp "white-list" "true"
--- a/start-finalSetupServerProperties
+++ b/start-finalSetupServerProperties
@@ -27,7 +27,7 @@ function setServerProp {
 }
 
 function customizeServerProps {
-  if [ -n "$WHITELIST" && "$WHITELIST" != "false" ]; then
+  if [ -n "$WHITELIST" ] || isTrue ${ENABLE_WHITELIST:-false}; then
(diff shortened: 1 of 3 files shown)


start-finalSetupServerProperties
```diff
@@ -27,7 +27,7 @@ function setServerProp {
 }
 
 function customizeServerProps {
-  if [ -n "$WHITELIST" ]; then
+  if [ -n "$WHITELIST" && "$WHITELIST" != "false" ]; then
```
itzg, line 30: This is too confusing overloading `WHITELIST`. 

How about
```suggestion
  if [ -n "$WHITELIST" ] || isTrue ${ENABLE_WHITELIST:-false} ; then
```

start-minecraftFinalSetup
```diff
@@ -13,7 +13,7 @@ if isTrue "${OVERRIDE_OPS}"; then
   rm -f /data/ops.json
 fi
 
-if [ -n "$WHITELIST" ]; then
+if [ -n "$WHITELIST" && "$WHITELIST" != "true" && "$WHITELIST" != "false" ]; then
```
itzg, line 16: ...then can just revert this to
```suggestion
if [ -n "$WHITELIST" ]; then
```

itzg (changes requested):
Great idea, but let's simplify it a bit.

P4sca1:
I like the idea of introducing a new `ENABLE_WHITELIST` env variable.
I have pushed another commit to address your suggestion.

itzg (approved):
Thanks!

Issue #938, opened 2021-06-28
version variable gets auto resolved to latest even when set.


Audition-CSBlock:
Hi, 
I've been having issues with the version variable, whenever i specify 1.16.5 in my variable, it still resolves it to "LATEST" and corrupts my world saves. It has done this on 2 big worlds of mine and all the data is now corrupt.

This usually happens right after i update the docker image, if i start up a fully new one with the version also set to 1.16.5; it just downloads the specified version and works. is there any fix for this?

I'm running these docker images on my unraid server, using the docker tag: "latest".

Audition-CSBlock:
I figured it out quite quickly; the variables in docker are case-sensitive in some cases. using the variable "VERSION" instead of "version" fixed my issue. I hope my mistakes help someone in the future :)

PR #939, opened 2021-06-28
Improve readme with extra info

Audition-CSBlock:
Today i opened an issue #938 and i don't want people to make the same mistake i did and potentially ruining their worlds, so i included a tiny line that says "case sensitive" near the explanation of the variable.


itzg (approved):
All environment variables (regardless of Docker) are case sensitive, but `VERSION` is important enough that this is a good place to mention it.

PR #942, opened 2021-07-02
Fix unquoted COPY_MODS_DEST and COPY_CONFIG_DEST variables

### LINKED ISSUE #941: COPY_MODS_DEST variable can't handle paths with spaces
### Use case

I'm trying to copy an Optifine mod in on top of a modpack that requires it, but this modpack unpacks to the path `/data/FeedTheBeast/Minecraft Zombie Server`. Setting `COPY_MODS_DEST="/data/FeedTheBeast/Minecraft Zombie Server"` doesn't work because usages of the variable aren't quoted.

### Solution

Change instances of `$COPY_MODS_DEST` to `"$COPY_MODS_DEST"`

austinkeller:
Closes #941

### DIFF
--- a/start-finalSetupMounts
+++ b/start-finalSetupMounts
@@ -25,16 +25,16 @@ fi
 
 if [ -d /mods ]; then
   log "Copying any mods over..."
-  mkdir -p $COPY_MODS_DEST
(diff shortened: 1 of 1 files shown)


itzg (approved):
Thanks! I have been gradually fixing other spots like that as I find them.

PR #954, opened 2021-07-07
fix: don't echo double backslashes to server.properties

### LINKED ISSUE #949: Doubled escapes in first entry in server.properties
The following `echo` command doubles backslashes, which is not necessary as in the case few lines up with `sed`:

https://github.com/itzg/docker-minecraft-server/blob/c090095e08d29f822ff45d98774bdf7abd6ecec0/start-finalSetupServerProperties#L22

This results in the following `server.properties` file with `SERVER_NAME` environment variable value `Old\u016Fv server`:

```
server-name=Old\\u016Fv server
```

P4sca1:
Using `sed`, we need to escape backslashes, but with `echo` we can just use the variable as-is.
Closes #949

### DIFF
--- a/start-finalSetupServerProperties
+++ b/start-finalSetupServerProperties
@@ -19,7 +19,7 @@ function setServerProp {
       sed -i "/^${prop}\s*=/ c ${prop}=${var//\\/\\\\}" "$SERVER_PROPERTIES"
     else
       log "Adding ${prop} with '${var}' in ${SERVER_PROPERTIES}"
-      echo "${prop}=${var//\\/\\\\}" >> "$SERVER_PROPERTIES"
+      echo "${prop}=${var}" >> "$SERVER_PROPERTIES"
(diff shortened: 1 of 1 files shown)


oldium (approved):
Looks good

itzg (approved):
Thanks! The further discussion on the issue helped me to see what was described originally.

oldium:
It looks like auto-merging of changes to branches is broken: https://github.com/itzg/docker-minecraft-server/actions/workflows/merge-variants.yml. It fails on merging `.github/workflows/build-multiarch.yml` to branches without `workflows` permission.

itzg:
Yep, that's by design in GitHub actions. And then I forgot to do the merge myself manually. Doing that now...

itzg:
Merged and new images pushed

oldium:
Great, thanks!

PR #975, opened 2021-07-14
fix: SPIGET_RESOURCES not always download

### LINKED ISSUE #974: Avoid re-downloading resources with SPIGET_RESOURCES
Right now the resources listed in SPIGET_RESOURCES are downloaded on every container startup. If the container restarted in a short period of time, the rate limiting of the spiget endpoint will start returning 504's.

I'm thinking of having the script create a marker file `/data/plugins/.{resource}` since that'll also work with the `REMOVE_OLD_MODS` feature where the plugins directory gets deleted:

https://github.com/itzg/docker-minecraft-server/blob/4357066cdf66bc82f2fb32393bbafc9122dba3cb/start-spiget#L50

DanielRamp:
implemented to only re-download SPIGET resources with REMOVE_OLD_MODS enabled

fixes #974 partially

todo:
- [ ] save timestamp instead of empty file
- [ ] compare timestamp in file with timestamp from SPIGET API

^ todos not needed because getting the timestamps from the API also might rate limit.

### DIFF
--- a/start-spiget
+++ b/start-spiget
@@ -26,21 +26,28 @@ getResourceFromSpiget() {
 
   log "Downloading resource ${resource} ..."
 
-  tmpfile="/tmp/${resource}.zip"
-  url="https://api.spiget.org/v2/resources/${resource}/download"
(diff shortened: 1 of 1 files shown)


itzg (approved):
Looks great as is. Let me know if you're good with me merging this.

PR #977, opened 2021-07-15
Update to new Purpur download API

Jawa-Juice:
This PR moves the Purpur deploy script to their new V2 downloads API.


itzg (approved):
Thanks

PR #978, opened 2021-07-15
Feature/spiget slowdown

DanielRamp:
implements **file time check** and **resource version check** via SPIGET API

`SPIGET_DOWNLOAD_TOLERANCE` environment variable for setting the min allowed age before even consider checking SPIGET API for version metadata. _(currently undocumented/no README additions)_

enhancement/followup of #974 / #975

maybe there is a better name for the variables, especially for SPIGET_DOWNLOAD_TOLERANCE, if you want you might change them.

Also it might be useful to put all these many commits into less, for better overview of the project history maybe?

### DIFF
--- a/start-spiget
+++ b/start-spiget
@@ -21,6 +21,48 @@ containsJars() {
   return 1
 }
 
+getResourceFromSpiget() {
+  resource=${1?}
+
+  log "Downloading resource ${resource} ..."
+
+  if [ ! -d /data/plugins ]; then
+    mkdir -p /data/plugins
+  fi
+
+  SPIGET_DOWNLOAD_TOLERANCE=2 # in minutes
+
+  versionfile="/data/plugins/.${resource}-version.json"
+  versionfileNew="/tmp/.${resource}-version.json"
+  versionfileTimestamp=$(date +%s -r $versionfile)
+  versionfileMinTimestamp=$(($(date +%s) - 60 * $SPIGET_DOWNLOAD_TOLERANCE))
+
+  if [ "$versionfileTimestamp" -ge "$versionfileMinTimestamp" ]; then
+    log "resource '${resource}' not checked because version meta file newer than '${SPIGET_DOWNLOAD_TOLERANCE}' minutes"
+  else
+    urlVersion="https://api.spiget.org/v2/resources/${resource}/versions/latest"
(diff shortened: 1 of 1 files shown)


itzg:
> Also it might be useful to put all these many commits into less, for better overview of the project history maybe?

Quick note on this: don't worry at all. I always squash and merge PRs so it'll end up as one commit on the master branch anyway.

start-spiget
```diff
@@ -26,32 +27,67 @@ getResourceFromSpiget() {
...
-    if ! curl -o "${tmpfile}" -fsSL -H "User-Agent: itzg/minecraft-server" "${extraCurlArgs[@]}" "${url}"; then
-      log "ERROR failed to download resource '${resource}' from ${url}"
-      exit 2
-    fi
+  versionfile="/data/plugins/.${resource}-version.json"
+  versionfileNew="/tmp/.${resource}-version.json"
+
+  if [ -f "$versionfile" ]; then
+    versionfileTimestamp=$(date +%s -r $versionfile)
+    versionfileMinTimestamp=$(($(date +%s) - 60 * $SPIGET_DOWNLOAD_TOLERANCE))
```
itzg, line 39: Double check me on this, but I think you can simplify this and the condition below with a `find` trick like:

```
if [[ -n $(find "$versionfile" -mmin +${SPIGET_DOWNLOAD_TOLERANCE}) ]]; then
  urlVersion="https://api.spiget.org/v2/resources/${resource}/versions/latest"
  # and so on
```

start-spiget
```diff
@@ -26,32 +27,67 @@ getResourceFromSpiget() {
...
 
-    if containsJars "${tmpfile}"; then
-      log "Extracting contents of resource ${resource} into plugins"
-      unzip -o -q -d /data/plugins "${tmpfile}"
-      touch "/data/plugins/.${resource}"
-      rm "${tmpfile}"
+    if [ "$versionfileTimestamp" -ge "$versionfileMinTimestamp" ]; then
```
itzg, line 41: ...suggestion above flips the conditional, so the log here would go in an `else`.

itzg (commented):
Looks great. Just a couple nitpicky suggestions.

DanielRamp (commented):
now corrected, with even one more fixes :)

itzg (approved):
Changes look great. I'll do a sanity check in a little while and then I'll merge it.

PR #985, opened 2021-07-21
Reflect different jar names for airplane/airplanepurpur

Jawa-Juice:
Not sure if I missed this or if it's yet another change to the CI, but here we are again.


itzg (approved):
Thanks again

PR #996, opened 2021-08-07
Add rlcraft example

Midnight-Conqueror:
After quite a bit of trial and error I finally figured out how to get rlcraft running using docker compose. Here is that example

### DIFF
--- a/examples/docker-compose-rlcraft.yml
+++ b/examples/docker-compose-rlcraft.yml
@@ -1,0 +1,20 @@
+version: "3.8"
+
+services:
+    mc:
+      image: itzg/minecraft-server:1.9.0
+      volumes:
+        - /modpacks:/modpacks
+        - mc-rlcraft:/data
+      ports:
+        - 25565:25565
+      environment:
+        EULA: "true"
+        TYPE: FORGE
+        VERSION: "1.12.2"
+        FORGEVERSION: "14.23.5.2854"
+        MEMORY: "4G"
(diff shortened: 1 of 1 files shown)


examples/docker-compose-rlcraft.yml
```diff
@@ -0,0 +1,20 @@
+version: "3.8"
+
+services:
+    mc:
+      image: itzg/minecraft-server:1.9.0
+      volumes:
+        - /modpacks:/modpacks
```
DanielRamp, line 7: I think the modpacks volume (``/modpacks:/modpacks``) contains a typo. Probalby one of eighter is meant:
a) relative path: ``./modpacks:/modpacks``
b) docker-volume: ``modpacks:/modpacks``

itzg: @Midnight-Conqueror I agree wit @DanielRamp . Can you please adjust this?

examples/docker-compose-rlcraft.yml
```diff
@@ -0,0 +1,20 @@
+version: "3.8"
+
+services:
+    mc:
+      image: itzg/minecraft-server:1.9.0
```
itzg, line 5: I would prefer not to pin this to a specific version. Instead reference one of the tags.

itzg (changes requested):

itzg:
Superseded by #1036

PR #1004, opened 2021-08-18
forge update for MC 1.7.2 with mc172 postfix

ronnierap:
Updates in de URLs will now work also with: <maven forge repo>/1.7.2-10.12.2.1161-mc172/forge-1.7.2-10.12.2.1161-mc172-installer.jar


ronnierap:
There is also the issues what 1.7.2 has not been fixed yet:
https://forums.minecraftforge.net/announcement/13-some-old-installers-currently-do-not-work-use-latest-for-your-version/

So all the installers for 1.7.2 currently don't work, even if you are able to download the latest one.

ronnierap:
Closing this for now, seems 1.7.2 needs an update at the forge side.

PR #1009, opened 2021-08-21
Add support for CUSTOM_SERVER_PACK

ronnierap:
If you have custom servers that require multiple files you can now specify them as a zip package (which will extract contents to the /data dir)


itzg:
I seem to be missing docs for `GENERIC_PACK`, but your PR seems to be redundant with the intended combination of `TYPE=CUSTOM` and `GENERIC_PACK` where the custom server logic accounts for it here

https://github.com/itzg/docker-minecraft-server/blob/28b7afda1538b9fc2141560984bde6f0e3eeb83a/start-deployCustom#L23

and then the generic pack is processed here

https://github.com/itzg/docker-minecraft-server/blob/28b7afda1538b9fc2141560984bde6f0e3eeb83a/start-finalSetupModpack#L154-L186

Issue #743, opened 2021-02-01
error: Please make sure you are using the server version of the FTB modpack!


ibootstrapper:
I'm using the below docker-compose.yml file to run the server with RLCRaft though it exits with an error: "Please make sure you are using the server version of the FTB modpack!" any ideas what i'm doing wrong?
```
version: '3'
services:
  minecraft:
    ports:
      - "25565:25565"
    volumes:
            - "./data:/data"
    environment:
      MAX_MEMORY: 4G
      DIFFICULTY: hard
      VERSION: 1.12.2
      TYPE: CURSEFORGE
      CF_SERVER_MOD: modpacks/RLCraft+1.12.2+-+Beta+v2.8.2.zip
    image: itzg/minecraft-server
    restart: always
  rcon:
    image: itzg/rcon
    ports:
      - "4326:4326"
      - "4327:4327"
    volumes:
            - "./db:/opt/rcon-web-admin/db"
  ```

**Comment by itzg** [comment]
From the file name it looks like you may not be using the server version of the modpack. Try instead:

https://www.curseforge.com/minecraft/modpacks/rlcraft/files/2935323

**Comment by isamuelson** [comment]
I ended up changing to the following and it works though there are many time lags
```
version: '3'

services:
  rlcraft:
    ports:
      - "25565:25565"
    environment:
      EULA: "TRUE"
      MAX_MEMORY: 4G
      MAX_BUILD_HEIGHT: 200
      VIEW_DISTANCE: 10
      LEVEL_TYPE: LARGEBIOMES
      MAX_PLAYERS: 100
      CONSOLE: "false"
      DIFFICULTY: hard
      VERSION: 1.12.2
      TYPE: FORGE
      FORGEVERSION: 14.23.5.2838
      GENERIC_PACK: modpacks/RLCraft+Server+Pack+1.12.2+-+Beta+v2.8.2.zip
      GENERIC_PACK_STRIP_DIRS: "1"
      DISABLE_HEALTHCHECK: "true"
    image: itzg/minecraft-server
    restart: always
    volumes:
      - ./data:/data
      - ./modpacks:/modpacks

  rcon:
    image: itzg/rcon
    ports:
      - "4326:4326"
      - "4327:4327"
    volumes:
      - "./db:/opt/rcon-web-admin/db" #

```

itzg:
Nice workaround! I now vaguely remember having to do something like that for RLcraft.

itzg:
[...]

PR #1036, opened 2021-09-09
Add RLCraft example

brutalgg:
Adding an RLCraft compose example for future reference.

### DIFF
--- a/examples/docker-compose-rlcraft.yml
+++ b/examples/docker-compose-rlcraft.yml
@@ -1,0 +1,23 @@
+version: '3.8'
+
+services:
+  rlcraft:
+    image: itzg/minecraft-server:java8
+    container_name: rlcraft
+    volumes:
+    - rlcraft-modpack:/modpacks:ro
+    - rlcraft-data:/data
+    environment:
+      EULA: "true"
+      TYPE: "FORGE"
+      VERSION: "1.12.2"
+      FORGEVERSION: "14.23.5.2854"
+      MEMORY: "4G"
+      CF_SERVER_MOD: /modpacks/RLCraft_Server_Pack_1.12.2_Beta_v2.8.2.zip
+    ports:
+      - 25565:25565
+    restart: unless-stopped
+
+volumes:
+  rlcraft-data:
+  rlcraft-modpack:
--- a/examples/docker-compose-rlcraft.yml
+++ b/examples/docker-compose-rlcraft.yml
@@ -9,11 +9,15 @@ services:
     - rlcraft-data:/data
     environment:
       EULA: "true"
-      TYPE: "CURSEFORGE"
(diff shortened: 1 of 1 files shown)


brutalgg:
I saw that #996 had been created but it wasn't finished and I happened to need this exact compose file so I created it and figured I'd share if it was valuable.

examples/docker-compose-rlcraft.yml
```diff
@@ -0,0 +1,23 @@
...
+services:
+  rlcraft:
+    image: itzg/minecraft-server:java8
+    container_name: rlcraft
+    volumes:
+    - rlcraft-modpack:/modpacks:ro
+    - rlcraft-data:/data
+    environment:
+      EULA: "true"
+      TYPE: "FORGE"
```
itzg, line 12: TYPE needs to be "CURSEFORGE" to make use of CF_SERVER_MOD variable.

brutalgg: Fixed in a1e4203

brutalgg:
Seems making that changed has now caused some problems. Going to try and troubleshoot a fix.

itzg:
FWIW this is the one that worked for me

https://github.com/itzg/mc-modpacks-starter/blob/master/rlcraft/docker-compose.yml

brutalgg:
I'll try your exact compose @itzg but I'm pretty sure I had a similar one which was running into the error detailed in issue #743. Either way I've fixed the example compose to a working version for me with RLCraft actually being loaded this time.

brutalgg:
> FWIW this is the one that worked for me
> 
> https://github.com/itzg/mc-modpacks-starter/blob/master/rlcraft/docker-compose.yml

I double checked this config using a unmodified zip download from curseforge and it resulted in the following loop.

```
[init] Unpacking FTB server modpack /modpacks/RLCraft_Server_Pack_1.12.2_Beta_v2.8.2.zip ...
..........
[init] Please make sure you are using the server version of the FTB modpack!
```

examples/docker-compose-rlcraft.yml
```diff
@@ -0,0 +1,27 @@
...
+      VERSION: "1.12.2"
+      FORGEVERSION: "14.23.5.2855"
+      DIFFICULTY: "hard"
+      MAX_TICK_TIME: "-1"
+      VIEW_DISTANCE: "6"
+      ALLOW_FLIGHT: "true"
+      MEMORY: "4G"
+      GENERIC_PACK: "/modpacks/RLCraft_Server_Pack_1.12.2_Beta_v2.8.2.zip"
```
itzg, line 20: Ah yes, I remember this now being the solution for RLCraft since their server modpack is "broken".

itzg (approved):
Thanks!

PR #1109, opened 2021-11-05
document that default RCON_PASSWORD = minecraft in README

vorburger:
because while experimenting with RCON, it's highly confusing why without a RCON_PASSWORD it happens
to "just work" e.g. with https://github.com/willroberts/minecraft-client (which uses "minecraft" as the default password itself!)
but not with https://github.com/itzg/rcon-cli (which has no default password).

Default passwords are always a bad idea, and should ideally ideally just be removed (if you are willing to make that backward compatible breaking change).

### DIFF
--- a/README.md
+++ b/README.md
@@ -875,6 +875,7 @@ The server icon which has been set doesn't get overridden by default. It can be 
 ### Rcon
 
 To use rcon use the `ENABLE_RCON` and `RCON_PASSWORD` variables.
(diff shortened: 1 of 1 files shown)


itzg (approved):
Thanks for the detail and the cautionary note. I don't really like default passwords either, so I'm not sure why I started with one there 😀 But yeah, at this point I've gotten that default stuck there.

PR #1139, opened 2021-11-29
Whitelist UUID Support

Hi-ImKyle:
- Added a check to see if WHITELIST contains a "-", if it does then parse it as a list of UUIDs rather than normal Usernames

NOTE: It's a crude fix and I am by no means good at bash. I just wanted to give it a shot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -846,7 +846,13 @@ To whitelist players for your Minecraft server, pass the Minecraft usernames sep
 
     docker run -d -e WHITELIST=user1,user2 ...
 
+or 
+
+    docker run -d -e WHITELIST=uuid1,uuid2 ...
+
 If the `WHITELIST` environment variable is not used, any user can join your Minecraft server if it's publicly accessible.
+
+> NOTE: When using uuids in the whitelist, please make sure it is the dashed variant otherwise it will not parse correctly.
 
 > NOTE: When `WHITELIST` is used the server properties `white-list` and `whitelist` will automatically get set to `true`.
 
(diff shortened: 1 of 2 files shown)


Hi-ImKyle:
To make it use Mojang's API instead
Change `https://playerdb.co/api/player/minecraft/{}` to `https://api.mojang.com/user/profile/{}`
Change `'.["data"]["player"] | {"uuid": .id, "name": .username}'` to '. | {"uuid": .id, "name": .name}'

Hi-ImKyle:
The endpoint for Mojang to get Username from UUID is not reversible unlike PlayerDB's, so if you choose Mojang's instead you will need to add a warning to the [whitelisting players section](https://github.com/itzg/docker-minecraft-server#whitelist-players) explicitly stating that it's either they use Usernames OR UUID's, not both and it NEEDS to be the dashed UUID format in order for it to pass the UUID check(idk any other way).

```
Wrong:
-e WHITELIST=Notch,ba2828f4f94646599543c41438e6a0e0

Wrong:
-e WHITELIST=Notch,ba2828f4-f946-4659-9543-c41438e6a0e0

Correct:
-e WHITELIST=Notch,Kyleee`

Correct
-e WHITELIST=069a79f4-44e9-4726-a5be-fca90e38aaf5,ba2828f4-f946-4659-9543-c41438e6a0e0
```

itzg (commented):
This is great! Yeah, let's go with the third party API since it had the flexibility of either username or UUID. If they shut down at some point it can be switched over. 

Can you make a small change to the whitelist section in the docs to mention that either identifier can be used? With that I'm good to merge this.

Hi-ImKyle:
Should be enough of a note to warn users to use the dashed variant if they wish to use uuids.

itzg (approved):

PR #1149, opened 2021-12-02
Feature: Enforcing resource packs

Stealthii:
I've added a `server.properties` override for the 'require-resource-pack' option, to allow server operators to set this in tandem with the other resource pack options.

As an additional change, I've removed the details around escaping requirements, as this is now handled with the setServerPropValue function.

- Add server.properties override for 'require-resource-pack'
- Removed escape requirement note


itzg (approved):
Thanks!

itzg:
This has now been merged and published to all the image variants.

PR #1183, opened 2021-12-11
fix: log4j patch being applied to <1.7 versions

StoneLabs:
It appears we missed this in our first patch @itzg.

I've ___not___ tested this, but it should be fine.


itzg (approved):
Ah yes. Thanks.

Issue #1167, opened 2021-12-07
1.18 fabric not launching


cootshk:
### Describe the problem

server unable to launch fabric 1.18

### Container definition

```
Paste run command or compose file here
```
sudo docker run -d -p 25565:25565 -p 25565:25565/udp -v /____:/data -e EULA=TRUE -e TYPE=fabric -e SERVRE_NAME=HypixelBestOnDotEighteen -e OPS=______ -e MAX_PLAYERS=50 -e FORCE_GAMEMODE=true -e SPAWN_PROTECTION=1 -e MOTD="Sodium on 1.18" -e ALLOW_FLIGHT=TRUE -e PLAYER_IDLE_TIMEOUT=5 -e OP_PERMISSION_LEVEL=4 -e MAX_MEMORY=8G --name=mc6 itzg/minecraft-server

### Container logs

```
Paste logs here
```

[init] Running as uid=1000 gid=1000 with /data as 'drwxrwxr-x 3 1000 1000 4096 Dec  7 03:20 /data'
[init] Resolved version given LATEST into 1.18
[init] Resolving type given fabric
[init] Checking Fabric version information.
[init] server.properties already created, skipping
[init] Updating ops
[init] Checking for JSON files.
[init] Setting initial memory to 1G and max to 8G
[init] Starting the Minecraft server...
[03:21:13] [ERROR] [FabricLoader/]: Uncaught exception in thread "main"
java.lang.RuntimeException: An exception occurred when launching the server!
        at net.fabricmc.loader.impl.launch.server.FabricServerLauncher.main(FabricServerLauncher.java:68)
Caused by: java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
        at java.base/java.lang.reflect.Method.invoke(Unknown Source)
        at net.fabricmc.loader.impl.launch.server.FabricServerLauncher.main(FabricServerLauncher.java:66)
Caused by: java.lang.RuntimeException: Error invoking MC server bundler: java.lang.UnsupportedClassVersionError: net/minecraft/bundler/Main has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 60.0
[...]

Issue #1192, opened 2021-12-12
Can't get fabric to use the main version


Coding-Kiwi:
### Describe the problem

I did not configure the `VERSION` variable, so the server started with version 1.18
then I restarted the server, meanwhile 1.18.1 was released so the server automatically uses 1.18.1, sadly some fabric mods are not yet compatible. So I changed the `VERSION` to `1.18`

The server log correctly says

```
[init] Resolved version given 1.18 into 1.18
```

but it then continues to use the subversion
```
[INFO] [FabricLoader/GameProvider]: Loading for game Minecraft 1.18.1
[main/INFO]: Loading for game Minecraft 1.18.1
```

1.18.0 does not work

### Container definition

```
minecraft:
    image: itzg/minecraft-server
    restart: unless-stopped
    ports:
      - "25565:25565"
      - "24454:24454/udp"  
    volumes:
      - /etc/minecraft/cavesandcliffs:/data
    environment:
      TYPE: "FABRIC"
      EULA: "TRUE"
      CONSOLE: "FALSE"
      MAX_MEMORY: "5G"
      JVM_OPTS: "-Dlog4j2.formatMsgNoLookups=true"
      VERSION: "1.18"
```

### Container logs

```
minecraft_1         | [init] Running as uid=1000 gid=1000 with /data as 'drwxrwxr-x 10 1000 1000 4096 Dec 12 17:56 /data'
minecraft_1         | [init] Resolved version given 1.18 into 1.18
minecraft_1         | [init] Resolving type given FABRIC
minecraft_1         | [init] server.properties already created, skipping
minecraft_1         | [init] Checking for JSON files.
minecraft_1         | [init] Setting initial memory to 1G and max to 5G
minecraft_1         | [init] Starting the Minecraft server...
minecraft_1         | Starting net.fabricmc.loader.impl.game.minecraft.BundlerClassPathCapture
minecraft_1         | [19:03:07] [INFO] [FabricLoader/GameProvider]: Loading for game Minecraft 1.18.1
minecraft_1         | [19:03:08] [main/INFO]: Loading for game Minecraft 1.18.1
minecraft_1         | [19:03:08] [main/WARN]: Mod resolution failed
[...]

PR #1197, opened 2021-12-13
fix fabric not downloading/downgrading server jar (#1192)

StoneLabs:
See #1192.

### DIFF
--- a/scripts/start-deployFabric
+++ b/scripts/start-deployFabric
@@ -24,7 +24,7 @@ fi
 
 export SERVER=fabric-server-${VANILLA_VERSION}-${FABRIC_INSTALLER_VERSION}.jar
 
-if [[ ! -e ${SERVER} ]]; then
+if [ ! \( -e ${SERVER} -a -e "server-${VANILLA_VERSION}.jar" \) ]; then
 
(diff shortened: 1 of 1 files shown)


scripts/start-deployFabric
```diff
@@ -70,8 +70,12 @@ if [[ ! -e ${SERVER} ]]; then
     exit 10
   fi
 
+  mv server.jar "server-${VANILLA_VERSION}.jar"
   mv fabric-server-launch.jar "${SERVER}"
 fi
 
+# Specify which server jar to run
+echo "serverJar=server-${VANILLA_VERSION}.jar" >  fabric-server-launcher.properties
```
itzg, line 78: Change this to a debug log or remove

itzg:
Looks like this will fix #1167 too

itzg (commented):
Thanks. The indirection of Fabric and its launcher is annoying to me, so thank you for having the patience to adapt to its behavior.

scripts/start-deployFabric
itzg: I'm sorry, for some reason (maybe because I was using the github mobile app), my eyes stopped before the redirect.

PR #1198, opened 2021-12-13
Created new template for docker-compose-paper.yml

woeisme:
for submission to update docker-compose-paper.yml example in https://github.com/itzg/docker-minecraft-server/tree/master/examples


itzg (approved):
Thanks. This one is much better than the example I had before.

PR #1199, opened 2021-12-13
Remove Airplane 1.16 as unsupported

Jawa-Juice:
This PR removes Airplane 1.16, as it is no longer supported or hosted.


itzg (approved):
Thanks!

PR #1205, opened 2021-12-15
Added: More opencontainers-lables

GoliathLabs:
Added some more opencontainers labels to the build-multiarch workflow. I removed the author label from the dockerfile so that the author label is set for the person who owns the repo (e.g. if someone forked this repo and made some customizations to it).

Feel freel to leave some feedback 😄

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM eclipse-temurin:17-jdk
 
-LABEL org.opencontainers.image.authors="Geoff Bourne <itzgeoff@gmail.com>"
-
 RUN apt-get update \
   && DEBIAN_FRONTEND=noninteractive \
   apt-get install -y \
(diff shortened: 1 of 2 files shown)


itzg:
Other than the author change, these seem to defeat the point of the metadata action

https://github.com/docker/build-push-action/blob/master/docs/advanced/tags-labels.md

Can you clarify if I used that action wrong?

GoliathLabs:
Yes, you are right about that. I have only used the "workaround" in my projects so far with the separate metadata action I added here. Maybe it's better the way you did it. Still, we should adjust it to include the respective owner automatically.

Dockerfile
```diff
@@ -1,7 +1,5 @@
 FROM eclipse-temurin:17-jdk
 
-LABEL org.opencontainers.image.authors="Geoff Bourne <itzgeoff@gmail.com>"
```
itzg, line 3: This is the only change I'll accept

itzg (changes requested):

PR #1210, opened 2021-12-17
fix: verify that the downloaded resource is a valid plugin

dhoffend:
I came across the issue that the downloaded resources was in fact some kind of html output. Looks like the download failed somehow (in my case the dynmap plugin (274).

Example:

    http://api.spiget.org/v2/resources/274/download --> https://dev.bukkit.org/projects/dynmap/files/3561925

Sure ... the owner of dynmap is using a link that points to a webpage instead of the real download.

So the curl download generates a html output which gets saves to 274.jar ... which then creates errors when the server trys to load the plugin. At least we should verify that the downloaded .jar file is a valid minecraft plugin with `plugin.yml` in it.

### DIFF
--- a/scripts/start-spiget
+++ b/scripts/start-spiget
@@ -22,6 +22,20 @@ containsJars() {
   return 1
 }
 
+containsPlugin() {
+  file=${1?}
+
+  pat='plugin.yml$'
+
+  while read -r line; do
+    if [[ $line =~ $pat ]]; then
+      return 0
+    fi
+  done <<<$(unzip -l "$file")
+
+  return 1
+}
+
 getResourceFromSpiget() {
   resource=${1?}
 
@@ -81,9 +95,12 @@ downloadResourceFromSpiget() {
     log "Extracting contents of resource ${resource} into plugins"
     unzip -o -q -d /data/plugins "${tmpfile}"
     rm "${tmpfile}"
-  else
+  elif containsPlugin "${tmpfile}"; then
     log "Moving resource ${resource} into plugins"
     mv "${tmpfile}" "/data/plugins/${resource}.jar"
+  else
+    log "ERROR downloaded resource '${resource}' seems to be not a valid plugin"
+    exit 2
   fi
 
 }


dhoffend:
I reuploaded the branch to match the output of `unzip -l`

itzg (approved):
Thanks

Peterkal2112:
Hello,
I encounterred that error today:

[init] Running as uid=0 gid=0 with /data as 'drwxrwxrwx 1 0 0 8192 Jan 24 08:51 /data'
[init] Resolved version given  into 1.18.1
[init] Resolving type given PAPER
[init] Removing old PaperMC versions ...
[init] Downloading PaperMC 1.18.1 (build 175) ...
[init] Getting plugins via Spiget
[init] Downloading resource 274 ...
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /tmp/274.zip or
        /tmp/274.zip.zip, and cannot find /tmp/274.zip.ZIP, period.
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /tmp/274.zip or
        /tmp/274.zip.zip, and cannot find /tmp/274.zip.ZIP, period.
[init] ERROR downloaded resource '274' seems to be not a valid plugin


This message is looping ..

I can download that plugin manually and upload that to folder but that is just a wacky manual workaround.

itzg:
@Peterkal2112 can you double check that you have re-pulled the image that you are using? I tried `docker run -it --rm -e EULA=TRUE -e TYPE=PAPER -e SPIGET_RESOURCES=274 -e SETUP_ONLY=true itzg/minecraft-server` and it stopped correctly just after the 

```
[init] ERROR downloaded resource '274' seems to be not a valid plugin
```

Issue #1211, opened 2021-12-18
Add auto-stop capability, when all players have disconnected for a duration


itzg:
### Enhancement Type

A completely new feature

### Describe the enhancement

https://discord.com/channels/660567679458869252/660569641550217327/921738644300120105

PR #1212, opened 2021-12-19
Autostop feature

Oekn5w:
implements #1211 according to Discord description.

### DIFF
--- a/files/autostop/autostop-daemon.sh
+++ b/files/autostop/autostop-daemon.sh
@@ -24,11 +24,11 @@ do
     if mc_server_listening ; then
       TIME_THRESH=$(($(current_uptime)+$AUTOSTOP_TIMEOUT_INIT))
       logAutostop "MC Server listening for connections - stopping in $AUTOSTOP_TIMEOUT_INIT seconds"
-      STATE=K
+      STATE=II
     fi
     ;;
-  XK)
-    # Knocked
+  XII)
+    # Initial idle
     if java_clients_connected ; then
       logAutostop "Client connected - waiting for disconnect"
       STATE=E
--- a/files/autostop/autostop-daemon.sh
+++ b/files/autostop/autostop-daemon.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+# needed for the clients connected function residing in autopause
+. /autopause/autopause-fcns.sh
+
+. ${SCRIPTS:-/}start-utils
+
+# wait for java process to be started
+while :
+do
+  if java_process_exists ; then
+    break
+  fi
+  sleep 0.1
+done
+
+STATE=INIT
+
+while :
+do
+  case X$STATE in
+  XINIT)
+    # Server startup
+    if mc_server_listening ; then
+      TIME_THRESH=$(($(current_uptime)+$AUTOSTOP_TIMEOUT_INIT))
+      logAutostop "MC Server listening for connections - stopping in $AUTOSTOP_TIMEOUT_INIT seconds"
+      STATE=K
+    fi
+    ;;
+  XK)
+    # Knocked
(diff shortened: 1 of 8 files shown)


javabrett:
Thanks for the PR @Oekn5w .

As my launch-time version of this amounts to a oneline `sed` of the existing autopause `pause.sh` script:

```
sed -i -e "s/pkill -STOP java/rcon-cli stop/" /autopause/pause.sh;/start
```

... and the functionality (in terms of detect-idle, zero players) is otherwise identical to autopause, I wondered if an approach that shoehorns a custom action into the existing autopause could work, provided it isn't confusing ... as an alternative to duplicating `/autopause` as `/autostop` - or are there other differences (I haven't compared them yet).

So I was debating one of the following:

- Adding `ENABLE_AUTOSTOP`, having it set and trigger `ENABLE_AUTOPAUSE` transparently, but with `rcon-cli stop` as the action in `pause.sh` when the new variable is set.  Downside is the slightly confusing running of the autopause scripts (including potentially `pause.sh` but with autostop behaviour, but this seems minor.
- Adding variable `AUTOPAUSE_STOP` or default `false`.  Makes it clear that autostop is a variant of the autopause behaviour, requires `ENABLE_AUTOPAUSE=true`.  Variant would be `AUTOPAUSE_ACTION=pause` with option for `stop`.
- Variants where just `pause.sh` is copied to `stop.sh`.

My current leaning is towards the first of these.  WDYT - or are there other advantages in forking `autopause` to `autostop` entirely?

Oekn5w:
@javabrett

As there are major differences with how the feature works and requires outside container configuration, I thought this approach is more viable to avoid confusion compared to cobbling this together with a startup sed command.

Autopause does require a lot of validation and configuration for the knockd stuff, which is not needed for Autostop. Additionally we can provide a nice log message with the correct name.

files/autostop/autostop-daemon.sh
```diff
@@ -0,0 +1,69 @@
...
+  case X$STATE in
+  XINIT)
+    # Server startup
+    if mc_server_listening ; then
+      TIME_THRESH=$(($(current_uptime)+$AUTOSTOP_TIMEOUT_INIT))
+      logAutostop "MC Server listening for connections - stopping in $AUTOSTOP_TIMEOUT_INIT seconds"
+      STATE=K
+    fi
+    ;;
+  XK)
```
itzg, line 30: Does this state need to be here?

files/autostop/stop.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+. /start-utils
+
+logAutostopAction "Stopping Java process"
+
+rcon-cli stop >/dev/null
```
itzg, line 7: Some users may have rcon disabled. Even though that could be documented as a requirement, I'm thinking it would be more robust to send a TERM signal to pid 1, which is what docker stop is doing.

itzg (commented):
I agree with the general observation @javabrett that it looks like more code duplication than expected; however, I do also agree @Oekn5w with the improved usability with the current approach.

files/autostop/autostop-daemon.sh
Oekn5w: renamed

files/autostop/stop.sh
Oekn5w: done, rcon is even invoked when it is enabled by the server itself

itzg (approved):
I still need to research how can have GitHub actions do a one off test-image build for PRs, as needed. However, maybe I'll save that for another time and just merge this PR.

itzg:
I found some ideas for PR-driven test image building; however, I'll experiment with that separately.

Issue #1152, opened 2021-12-03
Support downloading & applying datapacks


cadenkriese:
### Enhancement Type

A completely new feature

### Describe the enhancement

Recreate the MODS and MODS_FILE environment variables but for datapacks that would be applied to world/datapacks.

gtdrakeley:
I am looking for exactly this feature. +1

PR #1214, opened 2021-12-19
Create datapack download script

cadenkriese:
Mostly copied from setupModpack script, downloads zip files into $LEVEL/datapacks.

Note that this currently does not handle custom world directories as the setupWorld script also doesn't handle it. Plus I would expect datapacks on Bukkit / derivative servers to be quite a niche case.

Implements #1152


itzg (approved):
Thanks for adding this! Do you feel good about the testing of this and ready for me to merge?

cadenkriese:
Yup, I just did a bit more testing to make sure, but it should be good to merge.

itzg:
The latest image now includes this change.

PR #1219, opened 2021-12-22
Direct RUN based on previous setup if RUN_FROM_ARGSFILE is true

tasarren:
This helps building dependencies and running the server itself separately.

At first I thought `EXEC_DIRECTLY` and `SETUP_ONLY` were made for this purpose, but it only works if you copy stdout into your custom script.

With this changes, running `SETUP_ONLY` and `SAVE_RUNARGS` will create a `/data/runargs.txt` file that will be readed later by the same `/start` script if existing and if `RUN_FROM_ARGSFILE` is true.. skipping the whole script process and just run the server.

With this method we went from ~2 minutes server upgrade to just 10 seconds, because the whole dependency flow has been done by an initializer.

I am not sure how to merge everything into java8 and other branches to test it out, too many conflicts that I'm unaware of, so any guidance would be nice.

This also gives a bit of extra security, at least in our deployments because don't inject ENV variables anymore into our running server, those are sent to the initializer and then pruned because don't need them anymore until the next deploy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1418,6 +1418,17 @@ When the container is signalled to stop, the Minecraft process wrapper will atte
 
 If you are using a host-attached data directory, then you can have the image setup the Minecraft server files and stop prior to launching the server process by setting `SETUP_ONLY` to `true`. 
 
+You can choose to store the executable arguments created with `SETUP_ONLY` for later use into `runargs.txt` by setting `SAVE_RUNARGS` to `true`
+
+#### Run from arguments file
+
+If you've chosen to store the executable arguments into `runargs.txt` you can boot your server faster by setting `RUN_FROM_ARGSFILE` to `true`.
+
+This means it will bypass every script used to build your server, like downloading new plugin versions, server specific updates, security checks and fixes... So it is faster boot but comes with a cost. 
+**Use with caution**
+
+> **NOTE:** You can not use `SETUP_ONLY=true` and `RUN_FROM_ARGSFILE=true` at the same time, because the later will skip every script used in the setup process.
+
 ## Autopause
 
 ### Description
--- a/scripts/start
+++ b/scripts/start
@@ -5,6 +5,19 @@
 umask 0002
 chmod g+w /data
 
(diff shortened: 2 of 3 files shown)


itzg (commented):
This is an interesting idea. I'm good with merging it, but want to point out that of course it's faster startup time because it bypasses all upgrade checks. In a fixed VERSION scenario that's probably handy...except when something like the paper installer needs to be redownloaded to pick up a vulnerability fix like the recent Log4jShell. Users need to opt into and choose the shortcut, which is why I bring up the point but not as a blocker. 

Don't worry about the other variants. I'll take care of that merging and only the gosu usage needs to be addressed manually with the alpine variant.

itzg:
Can you add a section to the README near the SETUP_ONLY section?

tasarren:
Sorry for the delay. Let me know if some modifications are needed, English is not my language so I can make some spelling mistakes.

I've running my servers with this method lately and so far so good, let's hope it doesnt bring any bugs ^^

Happy new year :partying_face:

README.md
```diff
@@ -1418,6 +1418,17 @@ When the container is signalled to stop, the Minecraft process wrapper will atte
...
 If you are using a host-attached data directory, then you can have the image setup the Minecraft server files and stop prior to launching the server process by setting `SETUP_ONLY` to `true`. 
 
+You can choose to store the executable arguments created with `SETUP_ONLY` for later use into `runargs.txt` by setting `SAVE_RUNARGS` to `true`
+
+#### Run from arguments file
+
+If you've chosen to store the executable arguments into `runargs.txt` you can boot your server faster by setting `RUN_FROM_ARGSFILE` to `true`.
+
+This means it will bypass every script used to build your server, like downloading new plugin versions, server specific updates, security checks and fixes... So it is faster boot but comes with a cost. 
+**Use with caution**
```
itzg, line 1428: I really appreciate the caution notice.

itzg (approved):

itzg (changes requested):

PR #1224, opened 2021-12-26
Add note about mc-status under autopause section

CDFN:
I had problems with mc-status with autopause enabled and I don't want other people to struggle with the same issue 😎 Thanks for help with finding out what was the issue, @itzg 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -1458,8 +1458,9 @@ Enable the Autopause functionality by setting:
 ```
 
 Autopause is not compatible with `EXEC_DIRECTLY=true` and the two cannot be set together.
-You should use mc-health instead mc-status for healtcheck/readiness probe with autopause enabled. mc-status is known to have issues with paused servers.
-    
+
+> When configuring kubernetes readiness/liveness health checks with auto-pause enabled, be sure to reference the `mc-health` wrapper script rather than `mc-status` directly.
+
 The following environment variables define the behaviour of auto-pausing:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1458,7 +1458,8 @@ Enable the Autopause functionality by setting:
 ```
 
 Autopause is not compatible with `EXEC_DIRECTLY=true` and the two cannot be set together.
-
+You should use mc-health instead mc-status for healtcheck/readiness probe with autopause enabled. mc-status is known to have issues with paused servers.
```
itzg, line 1461: I don't like the negative wording of this. mc-health is simply a wrapper of mc-status that is autopause aware and should be used for health checks:

https://github.com/itzg/docker-minecraft-server/blob/39519427bbc638442773329d88a8bd2d35c42691/bin/mc-health#L12

itzg (changes requested):

README.md
CDFN: Yeah I get your point. What would be your suggestion to reword that? Can't think of "neutral" way of saying that.

itzg: > When configuring kubernetes readiness/liveness health checks with auto-pause enabled, be sure to reference the `mc-health` wrapper script rather than `mc-status` directly.

CDFN:
Oh I completely forgot about this PR, sorry 😀 Thanks for doing it for me. Hopefully it helps others! 😄

itzg:
I almost forgot too and had been out visiting family anyway 😃

PR #1225, opened 2021-12-26
UPDATE DOCKER FILE

valentindoche:
Specify tcp and udp to work with docker managers (Cyberpanel)


itzg (commented):
You can always add more port mappings at container creation time. I'm not sure why a non-standard port needs to be in the Dockerfile.

PR #1229, opened 2021-12-27
Support downloading CurseForge modpack when USE_MODPACK_START_SCRIPT is false

Megasware128:
Not the best at bash scripts but I thought this is a clean way to get this done

### DIFF
--- a/scripts/start-deployCF
+++ b/scripts/start-deployCF
@@ -65,9 +65,7 @@ downloadModpack() {
 }
 
 if ! isTrue "${USE_MODPACK_START_SCRIPT:-true}"; then
-  if ! [ -f "${FTB_SERVER_MOD}" ]; then
-    downloadModpack
-  fi
+  downloadModpack
 
   needsInstall=true
   installMarker=/data/.curseforge-installed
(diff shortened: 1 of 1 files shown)


scripts/start-deployCF
```diff
@@ -35,10 +35,38 @@ FTB_SERVER_MOD=${FTB_SERVER_MOD:-$CF_SERVER_MOD}
...
+  fi
+  FTB_SERVER_MOD=${srv_modpack}
+}
+
 if ! isTrue "${USE_MODPACK_START_SCRIPT:-true}"; then
   if ! [ -f "${FTB_SERVER_MOD}" ]; then
-    log "ERROR unable to find requested modpack file ${FTB_SERVER_MOD}"
-    exit 2
+    downloadModpack
   fi
```
itzg, line 70: I'm thinking the conditional can be taken away and this if-block just replaced with a call to `downloadMockpack`. That function is doing URL and file-existence checks.

itzg (changes requested):
Nice refactoring! I think there's a little simplification can be done thanks to your work so far.

scripts/start-deployCF
Megasware128: Yeah you're right. Just tested it with a mounted volume zip, still works

itzg (approved):
Thanks!

PR #1230, opened 2021-12-27
Fix script not deleting old version files when upgrading

Megasware128:
This was broken for me when upgrading Valhelsia 3.4.4 to 3.4.8 on tag java8. I'm using the same strategy as the other serverJar variable to fix this


itzg (approved):
Thanks!

PR #1265, opened 2022-01-09
fix: Set default MOTD if unset

### LINKED ISSUE #1261: Changing MOTD doesn't work
### Describe the problem

I ran the latest image with ```-e MOTD="My Server"``` and the MOTD doesn't change. It keeps the default one 

I tried running the java17 image and the MOTD changes properly. That one was updated a day ago so I guess it was before 977d082638347092261e834d7e5dd2f54db0f5f6 when code related to MOTD was changed

### Container definition

```podman run -d -it -p 25565:25565 -e EULA=TRUE -e MOTD="My Server" itzg/minecraft-server```
```podman run -d -it -p 25565:25565 -e EULA=TRUE -e MOTD="My Server" itzg/minecraft-server:java17```

### Container logs

image

image

### PR DESCRIPTION
**Pull request by drewburr**
Fixes #1261 

`-v "$MOTD"` is used to check if a environment variable is unset, and returns 0 (FALSE) if the value is set.

Thank you to @Stealthii for identifying this issue.

### DIFF
--- a/scripts/start-setupServerProperties
+++ b/scripts/start-setupServerProperties
@@ -48,7 +48,7 @@ function customizeServerProps {
 
   # If not provided, generate a reasonable default message-of-the-day,
   # which shows up in the server listing in the client
-  if ! [ -v "$MOTD" ]; then
+  if [ -v "$MOTD" ]; then
     # snapshot is the odd case where we have to look at version to identify that label
     if [[ ${ORIGINAL_TYPE} == "VANILLA" && ${VERSION} == "SNAPSHOT" ]]; then
       label=SNAPSHOT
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/start-setupServerProperties [hunk 1]
```diff
@@ -48,7 +48,7 @@ function customizeServerProps {
 
   # If not provided, generate a reasonable default message-of-the-day,
   # which shows up in the server listing in the client
-  if ! [ -v "$MOTD" ]; then
+  if [ -v "$MOTD" ]; then
```
itzg, line 51: This is supposed to be
```suggestion
  if ! [ -v MOTD ]; then
```

itzg (changes requested):
Thanks for helping out on this. You had the right line and my original intent was correct, just a typo on the `-v` usage.

itzg:
Sorry for committing my own suggestion. Wanted to get this merged and published ASAP.

PR #1270, opened 2022-01-10
Docs: Flare flags

Jawa-Juice:
This PR documents the Flare Flags variable, as the profiler is now available in [plugin](https://github.com/TECHNOVE/FlarePlugin) form. Also adds a note under Pufferfish, since it's built-in (Pufferfish re-implements part of Airplane's featureset for 1.18).


itzg (approved):
Thanks!

PR #1292, opened 2022-01-21
ci: Reference branch and html URL from completed workflow in discord workflow

itzg:
cc @shotah


shotah:
Looks good to me, did you run it through actions yet?

PR #1293, opened 2022-01-21
fix: changed packwiz url to reflect new repo address

Micah-Nailima:
Was having issues with pulling the packwiz installer bootstrap and noticed that the repository has been changed from comp500/packwiz-installer-bootstrap to packwiz/packwiz-installer-bootstrap. Seems like it does redirect when calling the curl command but I was having issues with it not doing so and this fixed it.


itzg (approved):
Thanks!

Issue #1295, opened 2022-01-23
Airplane not downloading since Download URL is not existing anymore


gira0:
### Describe the problem

If you want to start the container with Airplane as the selected Version, the container errors immediately since the download URL doesn't resolve anymore.

Link to online DNS Resolving by Google: https://dns.google/query?name=ci.tivy.ca&rr_type=A&ecs= 

![image](https://user-images.githubusercontent.com/19488367/150698900-ac10c0bd-b9e6-469c-ba31-25eca63ddabd.png)

### Container definition

```
podman run -d -v /opt/minecraft/:/data 
     -e TYPE=AIRPLANE -e FORCE_REDOWNLOAD=true \
     -p 25565:25565 -e EULA=TRUE --name mc itzg/minecraft-server
```

### Container logs

```
[init] Running as uid=1000 gid=1000 with /data as 'drwxrwxr-x. 2 1000 1000 82 Jan 23 21:03 /data'
[init] Resolved version given LATEST into 1.18.1
[init] Resolving type given AIRPLANE
[init] Using Airplane-1.17 branch
[init] Removing old Airplane versions ...
[init] Downloading Airplane from https://ci.tivy.ca/job/Airplane-1.17/lastSuccessfulBuild/artifact/launcher-airplane.jar ...
[mc-image-helper] 21:34:40.305 ERROR : Operation failed: ci.tivy.ca: Name or service not known
[init] ERROR: failed to download from https://ci.tivy.ca/job/Airplane-1.17/lastSuccessfulBuild/artifact/launcher-airplane.jar (status=0)

```

itzg:
@gira0 the fix is now ready to test

gira0:
Works now. Thanks for the fast help!

PR #1296, opened 2022-01-23
Update Airplane download URL, clarify version support

Jawa-Juice:
This PR updates the Airplane download URL, closing #1295. Also clarifies lack of support for versions >1.17.1. Should be gtg, but please sanity check.


itzg (approved):
Looks good. Thanks!

PR #1304, opened 2022-01-25
Fix loading world from compressed tar archives

wlritchi:
The MIME type detected for a tar file depends on the compression scheme in use; match any of the schemes supported by mc-backup, as well as uncompressed tar files. `.tar.zst` archives require the `zstd` package; add it to the Dockerfile.

The existing test labelled as `tgz` actually contained just an uncompressed tar file padded to 10KiB with zeroes. I assume this was unintended (tar won't automatically compress according to the file extension unless you pass -a). Rename that test to `.tar`, and add equivalent tests for `.tar.gz`, `.tar.bz2`, and `.tar.zst` (the three types of compression supported by mc-backup).


itzg (approved):
Thanks! Yes, I can now picture myself doing only `tar cf` to create that test file.

PR #1305, opened 2022-01-25
Support Spigot-style and vanilla-style world zips, for any server type

### LINKED ISSUE #1290: Support loading Spigot-family world zips on non-Spigot servers
### Enhancement Type

Improve an existing feature

### Context
Servers in the Spigot family arrange their world files somewhat differently than other servers; for a world directory `foo/`, instead of locating the nether and end dimensions in `foo/DIM-1/` and `foo/DIM1/`, Spigot locates them in `foo_nether/DIM-1/` and `foo_the_end/DIM1/`. The `start-setupWorld` script is aware of this possibility, and knows to copy these directories for Spigot-family servers.

### Current behaviour
`start-setupWorld` determines whether to create Spigot-style dimensions based on the family of the server about to be launched. When loading a Spigot-style world zip for a standard-style server jar, the script picks only one of the dimensions (at random) and fails to copy the other two.

### Desired behaviour
`start-setupWorld` detects the existence of Spigot-style or standard-style dimensions based on what appears in the world zip, and renames them if needed. In particular, a world zip created using either style of dimensions should be correctly loaded by either style of server.

wlritchi:
Spigot-family servers put the Nether and End dimensions into separate world directories. Previously, the start-setupWorld script could only accept this layout when passing it to a Spigot-family jar.

This PR adjusts the approach for detecting the Spigot world layout, so that it can be applied when loading any server jar. It also adjusts the logic for world copying so that the extra dimensions will be loaded into the correct place. If the script finds both a Spigot-style and a vanilla-style Nether (or End), the Spigot-style one will take precedence. This matches the default behaviour of Spigot-family jars.

- [x] Unit tests added
- [x] Manually tested on real worlds

Fixes #1290


tests/setuponlytests/vanilla_world_for_spigot_server/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+
+services:
+  mc:
+    restart: "no"
+    image: ${IMAGE_TO_TEST:-itzg/minecraft-server}
+    environment:
+      EULA: "TRUE"
+      SETUP_ONLY: "TRUE"
+      VERSION: ${MINECRAFT_VERSION:-LATEST}
+      TYPE: "PAPER"
```
shotah, line 11: Noticed in the folder name you called out spigot, but see you are using Paper. Probably just a typo or me not knowing enough about what is going on. :)

tests/setuponlytests/spigot_world_for_spigot_server/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+
+services:
+  mc:
+    restart: "no"
+    image: ${IMAGE_TO_TEST:-itzg/minecraft-server}
+    environment:
+      EULA: "TRUE"
+      SETUP_ONLY: "TRUE"
+      VERSION: ${MINECRAFT_VERSION:-LATEST}
+      TYPE: "PAPER"
```
shotah, line 11: I need to know more of why we are using paper instead of spigot when that is declared in the folder name.

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
```diff
@@ -0,0 +1,5 @@
+mc-image-helper assert fileExists world/level.dat && \
```
shotah, line 1: Excited to see these, will need to go back and add for forge api and vanilla tweaks.

scripts/start-setupWorld
```diff
@@ -50,20 +46,65 @@ if [[ "$WORLD" ]] && ( isTrue "${FORCE_WORLD_COPY}" || [ ! -d "$worldDest" ] );
...
+    log "Copying world..."
     rsync --remove-source-files --recursive --delete "$baseDir/" "$worldDest"
+
     if [ "$FAMILY" = "SPIGOT" ]; then
-      log "Copying end and nether ..."
-      [ -d "${baseDir}_nether" ] && rsync --remove-source-files --recursive --delete "${baseDir}_nether/" "${worldDest}_nether"
-      [ -d "${baseDir}_the_end" ] && rsync --remove-source-files --recursive --delete "${baseDir}_the_end/" "${worldDest}_the_end"
+      if [ -d "${baseDir}_nether" ]; then
+        log "Copying Spigot Nether..."
+        rsync --remove-source-files --recursive --delete "${baseDir}_nether/" "${worldDest}_nether"
```
shotah, line 85: Looks great, curious if we couldn't make all these rsync's a single method that takes in "${baseDir}_nether" and "${worldDest}_nether". It really may clean up these lines 82 - 107. Either way this is great stuff and it doesn't need to happen, just thought it may make it easier to read and troubleshoot.

shotah (commented):
Excited to see all the tests! Only real question I have is about the Paper vs Spigot.

scripts/start-setupWorld
itzg: Just an FYI, I am gradually converting rsync usage over to `sync` subcommand of mc-image-helper, which takes a subset of the arguments here

https://github.com/itzg/mc-image-helper#sync-and-interpolate

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
itzg: Agreed, I'm so happy to see others already using the new assertions I just added

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
```diff
@@ -0,0 +1,5 @@
...
+mc-image-helper assert fileExists world_nether/DIM-1/some_spigot_nether_file && \
+mc-image-helper assert fileExists world_the_end/DIM1/some_spigot_end_file && \
+! mc-image-helper assert fileExists world_nether/DIM-1/some_vanilla_nether_file && \
```
itzg, line 4: Nice! I don't even need to add the "not" versions of assertions now that I see this technique.

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
itzg, line 1: Might need to test the test, but it should be executing these will fail-fast so probably don't even need to chain these together with `&&`

tests/setuponlytests/spigot_world_for_spigot_server/docker-compose.yml
itzg: Spigot is effectively a dead release stream; however, it lives on as a "family" designation where Paper is a member of that family.

itzg (approved):
Changes look great. I'll wait to merge until I hear back if there's any final tweaks you want to make based on review.

tests/setuponlytests/vanilla_world_for_spigot_server/docker-compose.yml
wlritchi: I'm mostly protecting against the possibility of a future refactor accidentally making the Spigot logic conditional on `TYPE` (i.e. apply only to Spigot itself) instead of `FAMILY` (apply to all Spigot-style jars). It felt like overkill to write a whole separate test for it, but might as well catch it here.

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
wlritchi: It might still be nice to have "not" versions so that the output can be silent when the test passes and explanatory when it fails, like we get for the non-negated case. Not a huge deal though.

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
wlritchi: Unfortunately not; I [tested without `&&` and with some of the asserts negated](https://github.com/wlritchi/docker-minecraft-server/actions/runs/1770211926) and it passed, even when it shouldn't have.

We could also add a `set -e` at the top of each verify script rather than `&&`-chaining, or we could `source` the verify scripts from a shell with `set -e`.

wlritchi:
From my perspective the review suggestions here are all good, but I don't see any of them as blockers to this PR; unless you feel strongly about them I'd be inclined to leave them for a future cleanup. Personally I'd still like to do some real-world testing, but if you think the unit tests are good enough that works for me too.

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
itzg: Great point. Assertions are easy to add, so I'll get that queued up.

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
itzg: I thought the latter was the case, so I need to look into that.

itzg:
No expectations on my part for any changes that are needed. I'm good with merging at this point or waiting until you have done more real world testing. Just let me know when you're ready.

tests/setuponlytests/conflicting_world_for_spigot_server/verify.sh
itzg: I found and fixed the missing `-e` 

https://github.com/itzg/docker-minecraft-server/commit/26809ef1f01654dfe001b5244c752a91461a270b

itzg:
Hi @wlritchi , just wanted to check in and see how testing was going.

wlritchi:
Sorry for the lack of activity with this; I've been on call for my day job over the last week so I haven't had a chance to put much time into side projects. I've got some time today, so I'm planning to do a few tests of this PR on my infra within the next 24 hours; I'll update you with the results. Thanks for checking in!

itzg:
No problem at all. I did my oncall rotation just a couple of weeks ago, so I hear ya.

wlritchi:
Manual testing failed; Spigot zips fail to load the overworld correctly (on either server type). I'm still determining what went wrong, but will update with a fix and tests that cover the oversight.

wlritchi:
Update to my above comment: for reasons unknown, Spigot zips are failing to retain links from portals in the nether to portals in the overworld, but only on first load. I've seen this same behaviour with the existing implementation, so it's not a regression nor a problem with this PR (still worth opening a ticket to look at though).

Since I had logged out in the nether on my testing worlds, this led me to arrive at a new, unrecognized location when I returned to the overworld on the final servers. I incorrectly believed that the overworld was being regenerated; in fact I was just somewhat far from where I expected to be.

Apologies for the false alarm above, but I wanted to block this PR from being merged while there was any doubt. **I now assess that this PR passes manual tests, and can be merged safely**.

itzg:
> Apologies for the false alarm above, but I wanted to block this PR from being merged while there was any doubt. I now assess that this PR passes manual tests, and can be merged safely.

Thanks for being thorough. Merge is on the way!

itzg:
Build is going https://github.com/itzg/docker-minecraft-server/actions/runs/1803546434

PR #1313, opened 2022-01-29
build: switch to matrix builds for Debian based variant builds

itzg:
Will be able to streamline and improve robustness of build and test of all the Debian variants. That will cover all except the legacy Alpine build of `java8` and the upcoming GraalVM base image, which is RPM based.

cc @shotah

### DIFF
--- a/.github/workflows/build-multiarch.yml
+++ b/.github/workflows/build-multiarch.yml
@@ -116,7 +116,7 @@ jobs:
         uses: docker/build-push-action@v2.8.0
         with:
           platforms: ${{ matrix.platforms }}
-          push: false
+          push: ${{ github.ref_name == 'master' }}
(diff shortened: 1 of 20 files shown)


shotah:
@itzg Looks pretty good, lots of changes, but that is why we have the tests in there. I think my only feedback was provided directly via chat. This is a great implementation and I think it'll make it much easier to test and expand in the future. Example if it is reasonable or not to create build, test, and deploy workflows to pass the variants into to break up the build process, but definitely not needed.

itzg:
I'm going to investigate skipping the server download during tests at a later time. As I was about to integrate a stub server jar I realized that I also want to the tested version to float for some of the variants via

```
      VERSION: ${MINECRAFT_VERSION:-LATEST}
```

PR #1316, opened 2022-01-30
Use Log4J2 JVM flag when configuration file download fails

ThisIsAreku:
This commit fallback to using the `-Dlog4j2.formatMsgNoLookups=true` flag to allow the container to start even if the download for a patched Log4J2 configuration file fail.
 I had the issue this morning with purpurmc.org returning a 503 Service Unavailable.


itzg (approved):
Great idea.

PR #1337, opened 2022-02-06
Add GameType filter, and add ForgeAPI Verify Files to validate Mod Downloads

shotah:


scripts/start-setupForgeApiMods
```diff
@@ -192,11 +194,11 @@ if [ "$MODS_FORGEAPI_FILE" ] && [ -z "$MODS_FORGEAPI_PROJECTIDS" ]; then
   fi
 
   # Needs loop here to look up release types befor calling download.
-  jq -n "$required_dependencies" | jq -c '.[]?' | while read current_project; do
+  jq -c '.[]?' $MODS_FORGEAPI_FILE | while read current_project; do
```
shotah, line 197: Bug that was fixed.

scripts/start-setupForgeApiMods
```diff
@@ -97,6 +97,7 @@ modFileByProjectID(){
     # Checking for a individual release type input, if not use global
     if [ $project_id_release_type ]; then
       updateReleaseNumber $project_id_release_type
+      unset $project_id_release_type
```
shotah, line 100: Bug that was fixed.

shotah:
@itzg here's the PR.

scripts/start-setupForgeApiMods
itzg: It looks like `unset` takes a variable name, so would be
```suggestion
      unset project_id_release_type 
```

itzg (commented):
Just the `unset` change I think is needed

itzg:
> I wonder if there is a way to be more verbose in the docker file

Specifically do you mean the docker compose file in each test directory?

scripts/start-setupWorld
```diff
@@ -34,6 +34,7 @@ if [[ "$WORLD" ]] && ( isTrue "${FORCE_WORLD_COPY}" || [ ! -d "$worldDest" ] );
     # Stage contents so that the correct subdirectory can be picked off
     mkdir -p /tmp/world-data
     if ! extract "$WORLD" /tmp/world-data; then
+      log "ERROR extracting world from $WORLD"
```
itzg, line 37: Thanks!

itzg (approved):
Awesome!

Issue #1342, opened 2022-02-08
TPPI2 modpack doesn't work when restarting/redeploying the container


Megasware128:
### Describe the problem

Hi, while deploying a new container with TPPI2 works alright, I can join it, do stuff, run it for couple of hours. When I stop the container I'm unable to start it again. I see a couple of things go wrong:

- [x] First thing is the checksum check doesn't work. The modpack hasn't changed, yet the script believes it to be. I have opened a pull request to fix this: #1345

- [ ] Second thing is, this doesn't happen the second start but the third start the container asks me if I want to replace the mods so I have to manually accept it:

`replace mods/ironchest-1.7.10-6.0.62.742-universal.jar? [y]es, [n]o, [A]ll, [N]one, [r]ename: A`
This is supposed to be automatically right?

- [ ] ~~Third thing which crashes the container is it tries to delete the logs folder but it's unable to do it~~ this happened because I emptied my logs directory:

```
+ find /data -mindepth 1 -depth -type d -empty -delete
find: cannot delete ‘/data/logs’: Resource busy
```

### Container definition

```
version: '3.8'

services:
  mc:
    image: itzg/minecraft-server:java8
    volumes:
    - data:/data
    - ./logs:/data/logs
    environment:
      EULA: "true"
      TYPE: FORGE
      MEMORY: 4G
      DEBUG: "TRUE"
      LEGACY_JAVA_FIXER: "TRUE"
      VERSION: 1.7.10
      FORGEVERSION: 10.13.4.1614
      GENERIC_PACK: https://dist.creeper.host/FTB2/privatepacks/TPPI2/1_0_2/TPPI2Server.zip
    ports:
      - "25565:25565"
volumes:
  data: {}
```

### Container logs

```
+ export HOME=/data
+ HOME=/data
++ id -u
++ id -g
++ ls -lnd /data
+ log 'Running as uid=1000 gid=1000 with /data as '\''drwxrwxrwx   10 1000     1000          4096 Feb  8 12:16 /data'\'''
+ local oldState
++ shopt -po xtrace
+ oldState='set -o xtrace'
+ shopt -u -o xtrace
[init] Tue, 08 Feb 2022 12:33:05 +0000 Running as uid=1000 gid=1000 with /data as 'drwxrwxrwx   10 1000     1000          4096 Feb  8 12:16 /data'
+ '[' '!' -e /data/eula.txt ']'
+ isTrue false
+ local oldState
++ shopt -po xtrace
[...]

PR #1345, opened 2022-02-08
Distro aware checksum

Megasware128:
This fixes the first problem described in issue #1342. Otherwise the checksum would always be failing on Alpine

### DIFF
--- a/scripts/start-setupModpack
+++ b/scripts/start-setupModpack
@@ -215,7 +215,7 @@ if [[ "${GENERIC_PACKS}" ]]; then
   log "Checking if generic packs are up to date"
   if isTrue "${SKIP_GENERIC_PACK_UPDATE_CHECK:-false}" && [ -f "$sum_file" ]; then
     log "Skipping generic pack update check"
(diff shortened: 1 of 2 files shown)


itzg (approved):
Awesome! This approach is one of the last pieces of the puzzle to add Alpine to the matrix builds, such as

https://github.com/itzg/docker-minecraft-server/actions/runs/1806504323

itzg:
...and FYI the final piece in the puzzle: https://github.com/itzg/docker-minecraft-server/pull/1349

Megasware128:
@itzg might be fun to know that the get distro line was written by GitHub Copilot 😆

itzg:
That is a fun fact especially since I was impressed it can be determined in one shot like that. Oh...I might be thinking of further retrieving the distro version.

Megasware128:
@itzg This is what Copilot makes of it:
```Bash
# Get distro version
distro_version=$(cat /etc/os-release | grep -E "^VERSION_ID=" | cut -d= -f2 | sed -e 's/"//g')
```
Basically same strategie

itzg:
Aha, it's the pre systemd days that I'm remembering...and didn't realize systemd also included os-release

http://0pointer.de/blog/projects/os-release

PR #1349, opened 2022-02-10
build: add alpine support to matrix builds

itzg:


build/run.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+set -e
+
+distro=$(cat /etc/os-release | grep -E "^ID=" | cut -d= -f2 | sed -e 's/"//g')
+
+"$(dirname "$0")/${distro}/$1".sh
```
shotah, line 7: Nice folder shim!

tests/setuponlytests/test.sh
```diff
@@ -40,7 +40,7 @@ $logs
     echo "${folder} PASS"
   fi
 
-  docker-compose down -v --remove-orphans
+  docker-compose down -v --remove-orphans > /dev/null
```
shotah, line 43: Do we need the -v if we are just outputting to Null anyways? Notting seeing -v referenced, but pretty sure it's short for --verbose. https://docs.docker.com/compose/reference/

itzg: It's for removing any names volumes

https://docs.docker.com/compose/reference/down/

scripts/start-utils
```diff
@@ -237,11 +237,15 @@ function extract() {
   esac
 }
 
+function getDistro() {
+  cat /etc/os-release | grep -E "^ID=" | cut -d= -f2 | sed -e 's/"//g'
```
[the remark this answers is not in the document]
itzg: That particular one apparently came from Github Copilot via a recent PR

itzg:
@shotah Thanks for taking a look at the PR BTW

tests/setuponlytests/test.sh
shotah: Doh, I think I remember that now. :) Glad you caught it!

shotah:
> @shotah Thanks for taking a look at the PR BTW

Hey I'm excited for alpine to be reintroduced. This does updated it for Java17 and Latest correct?

itzg:
Yep, it became super easy to add back an alpine variant for Java 17. It's also going to make it much easier to add GraalVM, which has a base image on CentOS.

PR #1362, opened 2022-02-14
Removing unused ttf-dejavu package

itzg:
The package ttf-dejavu adds 17MB to the alpine images (and probably similar for Ubuntu). I unable to identify through source history why the package was added in the first place. Being a server environment, fonts are not typically applicable.


shotah:
:+1:

Issue #1233, opened 2021-12-28
Auto Pause fails to initialize if UID & GID is specified


FineWolf:
### Describe the problem

In the case where ENABLE_AUTOPAUSE=true, UID=### (not 0) and GID=### (not 0), ENABLE_AUTOPAUSE fails to initialize with the following error:

```log
[init] Updated server port in knockd config
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
```

### Container definition

```js
[
    {
       // Docker config
        "Config": {
            "Hostname": "2b7bf2f09776",
            "Domainname": "",
            "User": "",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "ExposedPorts": {
                "25565/tcp": {},
                "25575/tcp": {}
            },
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": [
                "OVERRIDE_SERVER_PROPERTIES=true",
                "OPS=[REMOVED]",
                "TYPE=SPIGOT",
                "SPIGET_RESOURCES=81534",
                "INIT_MEMORY=512M",
                "MAX_MEMORY=4G",
                "MOTD=[REMOVED]",
                "ENABLE_ROLLING_LOGS=true",
                "USE_AIKAR_FLAGS=true",
                "UID=1000",
                "GID=100",
                "TZ=America/New_York",
                "MODE=survival",
                "SERVER_NAME=[REMOVED]",
                "ENFORCE_WHITELIST=true",
                "ENABLE_AUTOPAUSE=true",
                "EULA=true",
                "DIFFICULTY=normal",
                "ENABLE_WHITELIST=true",
                "SERVER_PORT=25566",
                "SNOOPER_ENABLED=false",
                "MAX_TICK_TIME=-1",
                "VIEW_DISTANCE=10",
                "PATH=/opt/java/openjdk/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "LANG=en_US.UTF-8",
                "LANGUAGE=en_US:en",
                "LC_ALL=en_US.UTF-8",
                "JAVA_VERSION=jdk-17.0.1+12",
                "JAVA_HOME=/opt/java/openjdk",
                "MEMORY=1G",
                "VERSION=LATEST",
                "ENABLE_RCON=true",
                "RCON_PORT=25575",
                "RCON_PASSWORD=[REMOVED]",
[...]

PR #1367, opened 2022-02-15
use user instead of group in sudoers file

### LINKED ISSUE #935: Autopause does not work on synology
docker image, tag 'latest'
spigot 1.17, java 16
synology ds1019+ (64 bit)
eth0 interface available in container, docker mode bridged

max-tick-time set to -1
rcon enabled with password, both in server.properties as well as in docker env vars
outside rcon port not mapped to container
using non-default uid and gid
autopause enabled, default settings

I am seeing one log entry: "[init] Autopause functionality enabled"
After that no more logentries related to autopause and spigot is not stopped

Tried to run knockd in debug mode and I see output when connecting and disconnecting

Any ideas?

Oekn5w:
Fixes failing to start autopause when `GID` is set to `100` (id of existing `users` group)

Fixes #935 and #1233

Alpine branches might need an extra PR afterwards.

### DIFF
--- a/files/sudoers-mc
+++ b/files/sudoers-mc
@@ -1,2 +1,2 @@
-%minecraft ALL=(ALL) NOPASSWD:/usr/bin/pkill
-%minecraft ALL=(ALL) NOPASSWD:/usr/sbin/knockd
+minecraft ALL=(ALL) NOPASSWD:/usr/bin/pkill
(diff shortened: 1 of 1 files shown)


itzg (approved):
Thanks! All variants including Alpine are now built from master. So this will get them all.

PR #1384, opened 2022-02-24
ForgeAPI: Random failure fix

shotah:
AC: Tests no longer randomly fail for forgeapi projectids
AC: ForgeAPI no longer attempts to redownload downloaded packages. 

Changes:
* Allowed remove old mods to work.
* Using // empty over unset
* Added downloaded modid array to prevent recheck and download of mods.
* Added file check in front of download. (I know '--skip-up-to-date' does this, but messaging is key here I think)
* Added personal build script to gitignore


itzg (approved):
Nice cleanup

PR #1397, opened 2022-03-04
Fixed missing export, fixed run order, added real world working example

shotah:


itzg (approved):
Great follow up. Thanks.

PR #1399, opened 2022-03-04
added variables for pufferfish, also added setup only test

shotah:
Fixing pufferfish bug, as well as adding a test for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,10 +539,10 @@ A [Pufferfish](https://github.com/pufferfish-gg/Pufferfish) server, which is "a 
 
     -e TYPE=PUFFERFISH
 
-> NOTE: The `VERSION` variable is used to select a Pufferfish branch to download from. The available options are "LATEST"
+> NOTE: The `VERSION` variable is used to select branch latest, 1.18, or 1.17. Use PUFFERFISH_BUILD to really select the SERVER VERSION number.
 
 Extra variables:
-- `PUFFERFISH_BUILD=lastSuccessfulBuild` : set a specific Pufferfish build to use
+- `PUFFERFISH_BUILD=lastSuccessfulBuild` : set a specific Pufferfish build to use. Example: selecting build 47 => 1.18.1, or build 50 => 1.18.2 etc
 - `FORCE_REDOWNLOAD=false` : set to true to force the located server jar to be re-downloaded
 - `USE_FLARE_FLAGS=false` : set to true to add appropriate flags for the built-in [Flare](https://blog.airplane.gg/flare) profiler
 
--- a/README.md
+++ b/README.md
@@ -539,7 +539,7 @@ A [Pufferfish](https://github.com/pufferfish-gg/Pufferfish) server, which is "a 
 
     -e TYPE=PUFFERFISH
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -539,7 +539,7 @@ A [Pufferfish](https://github.com/pufferfish-gg/Pufferfish) server, which is "a
 
     -e TYPE=PUFFERFISH
 
-> NOTE: The `VERSION` variable is used to select a Pufferfish branch to download from. The available options are "LATEST" and "1.18"
+> NOTE: The `VERSION` variable is used to select a Pufferfish branch to download from. The available options are "LATEST"
```
itzg, line 542: That last sentence is dangling. I'm thinking the whole note can just become "VERSION can only be set to LATEST, which is the default."

itzg (changes requested):
Thanks! Yeah, that R0.1 qualifier is awkward but at least it'll get caught during tests if they ever change it.

scripts/start-deployPufferfish
```diff
@@ -1,44 +1,43 @@
...
 
-log "Using Pufferfish-${PUFFERFISH_BRANCH} branch"
+PUFFERFISH_BUILD_JSON=$(curl -X GET -s "https://ci.pufferfish.host/job/Pufferfish-${VANILLA_VERSION%.*}/${PUFFERFISH_BUILD}/api/json")
+# Example: "url": "https://ci.pufferfish.host/job/Pufferfish-1.18/50/",
+PUFFERFISH_BUILD_URL=$(jq -n "$PUFFERFISH_BUILD_JSON" | jq -jc '.url // empty' )
+# Example: "fileName": "pufferfish-paperclip-1.18.2-R0.1-SNAPSHOT-reobf.jar",
+PUFFERFISH_BUILD_FILENAME=$(jq -n "$PUFFERFISH_BUILD_JSON" | jq -jc '.artifacts[].fileName // empty' )
+PUFFERFISH_BUILD_DOWNLOAD_URL="${PUFFERFISH_BUILD_URL}artifact/build/libs/${PUFFERFISH_BUILD_FILENAME}"
 
-export SERVER=pufferfish-${PUFFERFISH_BRANCH}-${PUFFERFISH_BUILD}.jar
```
itzg, line 23: SERVER is a very important variable used in the downstream scripts. Is the server starting up with this removed?

scripts/start-deployPufferfish
```diff
@@ -1,44 +1,43 @@
...
 
 log "Removing old Pufferfish versions ..."
 shopt -s nullglob
 for f in pufferfish-*.jar; do
-  [[ $f != $SERVER ]] && rm $f
+  [[ $f != "$PUFFERFISH_BUILD_FILENAME" ]] && rm "$f"
```
itzg, line 29: ...why the change from SERVER in the first place?

scripts/start-deployPufferfish
shotah: using setup only for testing. Let me set file name to server.

scripts/start-deployPufferfish
shotah: because  I was getting away from the static concatenation of filename. Will pivot file name to SERVER.

itzg (approved):
🐡!

PR #1402, opened 2022-03-05
Pufferfish: Remove regex processing

bram-pkg:


scripts/start-deployPufferfish
```diff
@@ -8,7 +8,7 @@ isDebugging && set -x
 
 IFS=$'\n\t'
 
-if [ "${VANILLA_VERSION%.*}" != "1.18" ] && [ "${VANILLA_VERSION%.*}" != "1.17" ]; then
+if [[ "${VANILLA_VERSION}" != "1.18" ]] && [[ "${VANILLA_VERSION}" != "1.17" ]]; then
```
itzg, line 11: It looks like the retrieval on line 18 also needs a similar tweak.

itzg (commented):
Thanks for jumping on this

scripts/start-deployPufferfish
bram-pkg: Changed!

itzg (approved):

PR #1451, opened 2022-04-01
Fix for lower sub version with a higher ID number

shotah:


scripts/start-setupForgeApiMods
```diff
@@ -77,7 +77,7 @@ retrieveVersionTypeNumber(){
 }
 
 modFileByProjectID(){
-  project_id=$(echo $1 | tr -d '"')
+  project_id=$(echo "$1" | tr -d '"')
```
itzg, line 80: Could be just this, but can save it for a future PR
```suggestion
  project_id="$1"
```

itzg (approved):

scripts/start-setupForgeApiMods
shotah: Sorry was at the gym, yes I saw the trim. I almost cut it out. Good call though! Thanks for the merge!

Issue #1452, opened 2022-04-02
MODE & DIFFICULTY do not work with SPONGEVANILLA because versionLessThan is wrong


vorburger:
### Describe the problem

That logic in [scripts/start-setupServerProperties](https://github.com/itzg/docker-minecraft-server/blob/4f876626641bd69b05e10c39fbbd6fb84f812a2b/scripts/start-setupServerProperties) which does the thing to recognize both numerical OR names for `MODE` & `DIFFICULTY` but then tries to convert and it as numerical value OR name depending on the server version doesn't appear to work for `TYPE=SPONGEVANILLA` servers, which are still 1.12.2 - but somehow `versionLessThan() { mc-image-helper compare-versions "${VANILLA_VERSION}" lt "${1?}"` seems to get that wrong.

Hopefully the logs provided help to illustrate the problem clearly enough?

### Container definition

```Dockerfile
FROM itzg/minecraft-server:java8
ENV EULA=TRUE
ENV TYPE=SPONGEVANILLA
ENV OVERRIDE_SERVER_PROPERTIES=true
ENV DIFFICULTY=0
ENV MODE=1
```

### Container logs

```
[init] Setting difficulty to 'peaceful' in /data/server.properties
[init] Setting mode
[init] Setting gamemode to 'creative' in /data/server.properties
(...)
[00:02:59 INFO]: Default game type: SURVIVAL
```

itzg:
You didn't specify `VERSION`. With LATEST resolving to 1.18.2, then that is not less than 1.12.2.

vorburger:
> You didn't specify `VERSION`. With LATEST resolving to 1.18.2, then that is not less than 1.13.

Right! But... should one have to specify `VERSION` when using `-e TYPE=SPONGEVANILLA`, which [according to its doc](https://github.com/itzg/docker-minecraft-server#running-a-spongevanilla-server) _by default will run the latest STABLE version_? Isn't `VERSION` sort of redundant in this case? Imagine `-e TYPE=SPONGEVANILLA -e SPONGEBRANCH=EXPERIMENTAL -e VERSION 1.12.2` - that's "correct", but in a perfect world one could imagine that it "knows" that automatically?

But perhaps simply adding a hint to the README is sufficient - how about #1453?

Up to you if you want to just close this with a doc clarification, or keep it upon for more future automagic.

itzg:
[...]

PR #1453, opened 2022-04-02
Add hint that SPONGEVANILLA requires VERSION to README (#1452)

vorburger:
see #1452


itzg (approved):
Seems a bit extraneous to have to explain `VERSION` with potentially every server variant, but if that bit helps then let's go with it.

PR #1455, opened 2022-04-02
Fixed version comparisons when using SPONGEVANILLA

### LINKED ISSUE #1452: MODE & DIFFICULTY do not work with SPONGEVANILLA because versionLessThan is wrong

itzg:
Fixes #1452

FYI @vorburger , I reverted the README change from #1453 since I realized `VERSION` wasn't the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,11 +609,10 @@ Enable SpongeVanilla server mode by adding a `-e TYPE=SPONGEVANILLA` to your com
     
 By default the container will run the latest `STABLE` version.
 If you want to run a specific version, you can add `-e SPONGEVERSION=1.11.2-6.1.0-BETA-19` to your command-line.
-Because the Minecraft version of the SpongeVanilla server cannot be automatically guessed, the `VERSION` should also be set accordingly.
     
 Beware that current [Sponge](https://www.spongepowered.org) `STABLE` versions for Minecraft 1.12 require using [the Java 8 tag](#running-minecraft-server-on-different-java-version):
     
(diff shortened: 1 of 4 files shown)


vorburger:
> FYI @vorburger , I reverted the README change from #1453 since I realized `VERSION` wasn't the issue.

Yup, makes sense!

PR #1481, opened 2022-04-18
Fix alpine incompatibility

### LINKED ISSUE #1480: mc-send-to-console doesn't work with alpine
### Describe the problem

mc-send-to-console uses gosu to send commands to console. However, alpine uses su-exec as replacement for gosu.

### Container definition

N/A

### Container logs

N/A

CDFN:
Fixes #1480

### DIFF
--- a/bin/mc-send-to-console
+++ b/bin/mc-send-to-console
@@ -12,8 +12,13 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
   exit 1
 fi
 
+distro=$(cat /etc/os-release | grep -E "^ID=" | cut -d= -f2 | sed -e 's/"//g')
 if [ "$(id -u)" = 0 ]; then
-  gosu minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
+  if [[ $distro == alpine ]]; then
+    exec su-exec minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
+  else
+    exec gosu minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
+  fi
 else
-  echo "$@" > "${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}"
+  echo "$@" >"${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}"
 fi
--- a/bin/mc-send-to-console
+++ b/bin/mc-send-to-console
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+. "/start-utils"
 : "${CONSOLE_IN_NAMED_PIPE:=/tmp/minecraft-console-in}"
 
 if [ $# = 0 ]; then
@@ -13,7 +13,7 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
 fi
 
 if [ "$(id -u)" = 0 ]; then
-  if [[ $distro == alpine ]]; then
+  if [[ getDistro == alpine ]]; then
     exec su-exec minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
   else
     exec gosu minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
--- a/bin/mc-send-to-console
+++ b/bin/mc-send-to-console
@@ -12,7 +12,6 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
   exit 1
 fi
 
-distro=$(cat /etc/os-release | grep -E "^ID=" | cut -d= -f2 | sed -e 's/"//g')
 if [ "$(id -u)" = 0 ]; then
   if [[ $distro == alpine ]]; then
     exec su-exec minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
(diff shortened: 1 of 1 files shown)


bin/mc-send-to-console
```diff
@@ -12,8 +12,13 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
   exit 1
 fi
 
+distro=$(cat /etc/os-release | grep -E "^ID=" | cut -d= -f2 | sed -e 's/"//g')
```
shotah, line 15: Should this be a global variable?

bin/mc-send-to-console
```diff
@@ -12,8 +12,13 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
...
 if [ "$(id -u)" = 0 ]; then
-  gosu minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
+  if [[ $distro == alpine ]]; then
```
shotah, line 17: I don't know enough here, but is there more than just this one command that needs to get flipped? I can do some searching too. :)

bin/mc-send-to-console
CDFN: Definitely no, I'm not very into bash scripting. Thanks for looking, I'll fix in a bit :)

bin/mc-send-to-console
CDFN: I haven't looked at other scripts, I just came accross this one when using on own server.

bin/mc-send-to-console
CDFN: It looks like there's already global variable called `distro` so we can just use it. I removed mine completely.

bin/mc-send-to-console
CDFN:
```bash
cdfn@cdfn-ryzen dev/docker-minecraft-server (fix/mc-send-to-console-alpine %) » grep -r "gosu"
scripts/start:    exec gosu ${runAsUser}:${runAsGroup} "${SCRIPTS:-/}start-configuration" "$@"
build/ubuntu/install-packages.sh:  gosu \
bin/mc-send-to-console:    exec gosu minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
```
It looks like there are only 3 usages, first one is covered with check already, second one is installation script for debian based distros and third is corrected via this PR.

bin/mc-send-to-console
itzg: No, there's not a global variable. Instead there's a helper function used as:

```
  distro=$(getDistro)
```

however, `mc-send-to-console` doesn't currently import start-utils, which would need this at the top

```
. "/start-utils"
```

bin/mc-send-to-console
itzg: The only other use of `gosu` is in `start`, which is already wrapped in a conditional:

https://github.com/itzg/docker-minecraft-server/blob/901b889f2d95c1a188070ca43d310bee4202d166/scripts/start#L48-L53

bin/mc-send-to-console
CDFN: That's weird. I rebuilt the container and the command seems to work just fine.
```
root@7b2c0b045c7c:/data# cat /usr/local/bin/mc-send-to-console 
#!/bin/bash

: "${CONSOLE_IN_NAMED_PIPE:=/tmp/minecraft-console-in}"

if [ $# = 0 ]; then
  echo "ERROR: pass console commands as arguments"
  exit 1
fi

if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
  echo "ERROR: named pipe ${CONSOLE_IN_NAMED_PIPE} is missing"
  exit 1
fi

if [ "$(id -u)" = 0 ]; then
  if [[ $distro == alpine ]]; then
    exec su-exec minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
  else
    exec gosu minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
  fi
else
  echo "$@" >"${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}"
fi
root@7b2c0b045c7c:/data# mc-send-to-console op CDFN
root@7b2c0b045c7c:/data# 
```

itzg: Since `distro` is unset, it'll be an empty string. So it'll be going in the `else` block, using `gosu` without any errors on a debian based image.

CDFN: Oh... I forgot that when you won't set --build-args it uses debian by default so that would be why :facepalm: I'll re-test with alpine based image, sorry for that :sweat_smile:

itzg: ...fyi if you're doing a docker build with the default, then it's building a debian image. Add the build args `BASE_IMAGE=openjdk:8-jdk-alpine3.8; DISTRO=alpine` to build a java8 alpine image.

CDFN: Should be sorted now :)

bin/mc-send-to-console
```diff
@@ -13,7 +13,11 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
 fi
 
 if [ "$(id -u)" = 0 ]; then
-  gosu minecraft bash -c "echo $* > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
+  if [[ getDistro == alpine ]]; then
```
itzg, line 16: I think this will need to be the following to actually run the function and not be a bare string "getDistro", which will never equal "alpine":

```suggestion
  if [[ $(getDistro) == alpine ]]; then
```

CDFN: Yes, again I have tested with wrong command zsh suggested me and ran build without args :facepalm: Commit fixing it is there.

itzg: Totally understand. I am grateful that IntelliJ makes it easy for me to switch between build types:

image

CDFN:
Yes :innocent:

PR #1483, opened 2022-04-20
added Loliserver support

Sylensky:
This PR adds support for the more recent hybrid server type LoliServer.
I am not very happy with how releases or updates are handled.
Built jars are just available on a jenkins server which the `lastSuccessfulBuild` gets `curl`ed and used.

In advance iam not sure if all necessary changes are made so i would be thankful for a review/test.

### DIFF
--- a/scripts/start-deployLoliserver
+++ b/scripts/start-deployLoliserver
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# shellcheck source=start-utils
+. "${SCRIPTS:-/}start-utils"
+set -o pipefail
+set -e
+
+latestAsset=$(
+  curl -fsSL https://cdn.ci.loliidc.cn:30011/job/LoliServer-1.16.5/lastSuccessfulBuild/artifact/projects/LoliServer/build/libs | \
+    jq '.assets[] | select(.name | match(".*-server.jar"))'
+)
+
+if [[ -z "${latestAsset}" ]]; then
+  log "ERROR: latest release of Loliserver is missing server.jar asset"
+  exit 1
+fi
+
+isDebugging && log "Latest asset ${latestAsset}"
+latestJarName=$(echo ${latestAsset} | jq --raw-output '.name')
+latestJarId=$(echo ${latestAsset} | jq --raw-output '.id')
+
+
+export SERVER="/data/${latestJarName}"
+
+if [ ! -f ${SERVER} ]; then
+  log "Downloading ${latestJarName}"
+  curl -H "Accept:application/octet-stream" -o "$SERVER" -fsSL https://cdn.ci.loliidc.cn:30011/job/LoliServer-1.16.5/lastSuccessfulBuild/artifact/projects/LoliServer/build/libs/${latestJarId}
+fi
+
+export FAMILY=HYBRID
+
+exec "${SCRIPTS:-/}start-setupWorld" "$@"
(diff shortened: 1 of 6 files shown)


scripts/start-deployLoliserver
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# shellcheck source=start-utils
+. "${SCRIPTS:-/}start-utils"
+set -o pipefail
+set -e
+
+latestAsset=$(
+  curl -fsSL https://cdn.ci.loliidc.cn:30011/job/LoliServer-1.16.5/lastSuccessfulBuild/artifact/projects/LoliServer/build/libs | \
```
itzg, line 9: Overall your code looks great. My concern is accessing this URL especially what appears to be an ephemeral port. 

Please try to use the artifact from the latest GitHub action such as https://github.com/Loli-Server/LoliServer/actions/runs/2195635097#artifacts

itzg (changes requested):

itzg:
Also adding a "setup only" test like this would be great:

https://github.com/itzg/docker-minecraft-server/tree/master/tests/setuponlytests/pufferfish

scripts/start-deployLoliserver
Sylensky: `latestAssetID=$(
  curl -sL https://api.github.com/repos/Loli-Server/LoliServer/actions/workflows/gradle.yml/runs | jq -r '.workflow_runs[0].id?'
)

latestAsset=$(
  curl -fsSL https://api.github.com/repos/Loli-Server/LoliServer/actions/artifacts/${latestAssetID}/zip | \
    jq '.assets[] | select(.name | match(".*.zip"))'
)`

When i try to get the latest artifact id and try to get that `.zip` package, curl throws a `403 Forbidden` on my head. Not sure what iam doing wrong.

`curl -H "Accept: application/vnd.github.v3+json" -o "$SERVER" -fsSL https://api.github.com/repos/Loli-Server/LoliServer/actions/artifacts/${latestAsset}/zip`

itzg: It looks like you need to grab the artifacts list from the run and use the `archive_download_url` from that:

```
curl --request GET \
  --url https://api.github.com/repos/Loli-Server/LoliServer/actions/runs/2195635097/artifacts
```

itzg: nm, the variable names were confusing me. I'm getting a 403 also using curl. That API seems to require the caller be authenticated, which is does seem to imply with

> Anyone with read access to the repository can use this endpoint

Sylensky: I was about to say that. Since `curl`ing the artifact to see if its there responds with missing scopes.

```
{
  "message": "You must have the actions scope to download artifacts.",
  "documentation_url": "https://docs.github.com/rest/reference/actions#download-an-artifact"
}
```

itzg: Well, I guess we're stuck with their Jenkins location. It feels like the README section you added needs some kind of warning/disclaimer...but I'm not sure that it should say. If you can't think of anything either, then I'm good with merging this.

Sylensky: I added a small disclaimer in [0b3cc9e](https://github.com/itzg/docker-minecraft-server/commit/0b3cc9e8bda6cba3e0fc198797e7ea623584dcbc)

itzg (approved):

itzg:
Sorry for the delay in merging this.

Scrumplex:
Looks like this server isn't a thing anymore. The GitHub page, as well as the JAR download seem to be gone.
See <https://github.com/Loli-Server/LoliServer>

Sylensky:
It moved to this repo here https://github.com/Luohuayu/CatServer with different branches for the versions.
Unfortunately i run currently low on time to create PR.

PR #1512, opened 2022-05-10
Fix memory env vars being overridden if set to ""

md5nake:
This commit ensures expected behaviour when trying out the example below, detailed in the README.

> To let the JVM calculate the heap size from the container declared memory limit, unset MEMORY with an empty value, such as -e MEMORY="". By default, the JVM will use 25% of the container memory limit as the heap limit; however, as an example the following would tell the JVM to use 75% of the container limit of 2GB of memory:
>  `-e MEMORY="" -e JVM_XX_OPTS="-XX:MaxRAMPercentage=75" -m 2000M`

### DIFF
--- a/scripts/start-finalExec
+++ b/scripts/start-finalExec
@@ -110,9 +110,9 @@ if [[ ${GUI,,} = false ]]; then
   EXTRA_ARGS+=" nogui"
 fi
 
-: "${MEMORY:=1G}"
-: "${INIT_MEMORY:=${MEMORY}}"
-: "${MAX_MEMORY:=${MEMORY}}"
+: "${MEMORY=1G}"
+: "${INIT_MEMORY=${MEMORY}}"
+: "${MAX_MEMORY=${MEMORY}}"
 
 expandedDOpts=
 if [ -n "$JVM_DD_OPTS" ]; then


scripts/start-finalExec
```diff
@@ -110,9 +110,9 @@ if [[ ${GUI,,} = false ]]; then
   EXTRA_ARGS+=" nogui"
 fi
 
-: "${MEMORY:=1G}"
-: "${INIT_MEMORY:=${MEMORY}}"
-: "${MAX_MEMORY:=${MEMORY}}"
+: "${MEMORY=1G}"
```
itzg, line 113: That looks like a syntax error, but I'll let the tests help determine that.

itzg: My apologies, I never knew the colon was an option since I have never seen it done that way over all my years of experience. 

I'm still uncertain why the change was needed. An initial issue report and/or more background about what you tried would have helped the review process.

itzg (approved):
I confirmed that I broke the behavior in 97fe6b1f14f597d6746d90d191f00acbf8b8ef5c and this PR fixes it the right way.

Admittedly it is awkward to "unset" the variable to indicate a different behavior, so I am going to introduce something like `MEMORY_PCT` to explicitly drive the host/container based derivation.

PR #1514, opened 2022-05-10
Add initial support for Quilt server

### LINKED ISSUE #1484: Add Quilt server support
### Enhancement Type

New server type

### Describe the enhancement

Quilt has been released today and it would improve this project to add it to the supported server types.

https://quiltmc.org/install/server/

fischer-felix:
Exactly what it says on the tin, support for the [Quilt](https://quiltmc.org) server type.

The start script is basically just a heavily modified version of the one for fabric, just changing the way it is installed.

While Fabric provides an easy way to obtain the server jar directly, Quilt uses their own installer tool written in java, thus I download this tool and then execute it with the correct directories and versions set.

This should resolve #1484.


scripts/start-deployQuilt
```diff
@@ -0,0 +1,53 @@
...
+elif [[ $QUILT_LAUNCHER_URL ]]; then
+  export SERVER=quilt-server-$(echo -n "$QUILT_LAUNCHER_URL" | mc-image-helper hash)
+# Official quilt installer
+else
+  if [[ ${QUILT_INSTALLER_VERSION^^} = LATEST ]]; then
+    log "Checking Quilt Loader version information."
+    QUILT_INSTALLER_VERSION=$(maven-metadata-release https://maven.quiltmc.org/repository/release/org/quiltmc/quilt-installer/maven-metadata.xml)
+  fi
+  export INSTALLER=quilt-installer-${QUILT_INSTALLER_VERSION}.jar
+  export SERVER=quilt-server-launch.jar
```
itzg, line 30: It would be good to qualify the file name by the server version to enable in place upgrades. If the installer doesn't let you override that, then you could rename the jar to the versioned name.

scripts/start-deployQuilt
```diff
@@ -0,0 +1,53 @@
...
+  fi
+fi
+
+
+if [[ ! -e ${SERVER} ]]; then
+  log "$SERVER does not exist, cannot launch server!"
+  exit 1
+fi
+
+export FAMILY=QUILT
```
itzg, line 52: Try to use one of the existing family values unless it is indeed not compatible with any other. If that's the case then you'll need to double check any later conditions on FAMILY that might need to be extended.

itzg (commented):
Thanks! Can you also add a test similar to `tests/setuponlytests/pufferfish`?

fischer-felix:
I hope I added the test correctly, I more or less just took the one from Pufferfish and modified the docker-compose.yml and verify.sh. Hope I changed the right things.

itzg:
New changes look great. The PR verification should be giving it a test run now.

itzg (approved):

tests/setuponlytests/quilt/verify.sh
```diff
@@ -0,0 +1 @@
+mc-image-helper assert fileExists "/data/quilt-server-launch-*.jar"
```
itzg, line 1: Looks like it should be
```suggestion
mc-image-helper assert fileExists "/data/quilt-server-*-launch.jar"
```

fischer-felix: should be fixed now

PR #1527, opened 2022-05-22
Add jarfile options and custom worlds directory path (Multiverse) to docs

skylerspaeth:
This PR aims to make it easier for users to find documentation on setting a Multiverse/other multi-world plugin's worlds directory path.

I couldn't find a built-in option to set a worlds directory that Multiverse would actually respect and use, so this seems to be the solution. I also added a section on jar file options and pointed it at the aforementioned worlds dir section as an example. 

If this would be more fitting for the [working-with-world-data](https://github.com/itzg/docker-minecraft-server#working-with-world-data) or [working-with-mods-and-plugins](https://github.com/itzg/docker-minecraft-server#working-with-mods-and-plugins) section, I'm happy to update the PR.


itzg (approved):
Thanks! That section location is as good as any. I'm surprised I was missing the `EXTRA_ARGS` docs the whole time.

PR #1531, opened 2022-05-24
Fixed older versions of purpur that were not removed

bensuperpc:
Fixed older versions of purpur that were not removed

Signed-off-by: Bensuperpc <bensuperpc@gmail.com>


itzg (approved):
Thanks!

PR #1548, opened 2022-06-05
updating forgeapi tests and adding is false

shotah:
* fixed tests
* found bug in ! isTrue logic, created a isFalse to simplify.


itzg (approved):
Good stuff. Thanks!

PR #1555, opened 2022-06-08
Resolve latest Paper version when no published builds

itzg:
As discussed https://discordapp.com/channels/660567679458869252/660567682751528981/984105818557403156


shotah (commented):
Nice work! Sorry you have to disable the forge stuff. Will have to look at that later and see if we can't pin the tests to a relevant version while updates like this happen.

PR #1559, opened 2022-06-12
Add RCON Web Dashboard

kaysond:
I threw this together a while back and realized I forgot to open a PR! This PR adds a web-based RCON dashboard to the docker container, enabled by the `ENABLE_DASHBOARD` env var. The dashboard itself is based on https://github.com/Rauks/Minecraft-RCON but is updated with newer bootstrap CSS, and support for the features in this container. It shows the current server status, and (if the server is running) the version of minecraft running. It has support for executing control actions on the server (start, stop, restart) and supports autopause/resume (including manual pause and resume actions). It does require adding a couple of packages to the image, even if you're not using the dashboard, but its a very small size increase.

If you're interested in adding this, I can rebase, update the documentation, and describe in more detail how its able to execute server actions.

image


itzg (commented):
I appreciate the initiative, but this should be a separate image like I did here: https://github.com/itzg/docker-rcon-web-admin

"Bundling" servers goes against the Docker philosophy of a single-purposed containers. Specifically that means avoidance of running concurrent processes...especially an entire Apache server. FYI, knockd was an exception that I grudgingly accepted.

I would recommend publishing your own image and I would be glad to accept a PR linking to it in https://github.com/itzg/docker-minecraft-server#deployment-templates-and-examples

kaysond:
I understand your concern. Unfortunately, I don't think there's really an easy and secure way of executing server actions, which requires control over that process, without being inside the same container.

And that's why I put this together in the first place. I could've just run the original dashboard in a separate container, but I wanted to be able to 1) see the version of Minecraft running, and 2) restart the shell wrapper to update it without having to get a terminal open (to make it easier for admins).

So what I did was make a new entrypoint file that wraps around the current one, which handles all server control actions by checking for the placement of certain files and their contents, which act as semaphores. When it finds those, it executes the action and deletes the file. The dashboard simply creates files and checks for their deletion. Supporting auto pause also adds another element of complexity. 

If you can think of a better way that allows secure access from the outside... I'm all ears. I contemplated using the file-semaphores via a bind mount that's shared between two containers but that seemed too clunky, complex, and unsafe. Another option could be a unix socket, but I didn't want to deal with cross platform issues. 

The "right" way would probably be to write a separate server controller program that handles the server downloading, env setup, etc, and listens on a tcp port somewhere for server commands. But since that's all already in shell scripts I didn't think it would be worth the effort. 

> "Bundling" servers goes against the Docker philosophy of a single-purposed containers. Specifically that means avoidance of running concurrent processes...especially an entire Apache server. FYI, knockd was an exception that I grudgingly accepted.

This philosophy seems to be going out of fashion. It's nice for organization and simplicity, but there's nothing fundamentally wrong with having multi-process containers. If you're familiar with Linuxserver.io, many of their containers run multiple services. Something like Jellyfin spawns ffmpeg processes for transcoding. It's also worth mentioning that in this case, Apache doesn't get started unless you're actually using the dashboard.

kaysond:
> I would recommend publishing your own image and I would be glad to accept a PR linking to it in https://github.com/itzg/docker-minecraft-server#deployment-templates-and-examples

Because any server control stuff would be specific to this image, I don't think it would be worth the effort to maintain an entirely separate repo

itzg:
I am very confused about your follow up.

RCON == "remote console"...**remote**

and RCON is password protected and can be constrained to internal docker networks.

> Unfortunately, I don't think there's really an easy and secure way of executing server actions

What server actions do you mean beyond server console commands?

> This philosophy seems to be going out of fashion. 

Well that's my philosophy, so I guess I'm out of fashion.

kaysond:
> What server actions do you mean beyond server console commands?

It can start or stop the server, which spawns/kills the Minecraft process but leaves the container and dashboard running. (Also restart which is just stop->start). It can also pause/resume, if you're using auto pause, which means suspending and resuming the process. 

> I am very confused about your follow up.
>
> RCON == "remote console"...remote
>
> and RCON is password protected and can be constrained to internal docker networks.

I'm talking about the internal security of controlling the Minecraft server process itself. Its probably a good idea to make sure it's not trivially easy to kill the server process outside of the dashboard or whatever control method is used. Obviously you'd want some kind of protection on rcon/any dashboard too.

> Well that's my philosophy, so I guess I'm out of fashion.

I mean I agree with you. I would separate nginx, mariadb, and redis, for example, into different containers, for example. I just don't think there's a need for a strict 1-container-1-process rule (and it seems you don't really either). Many containers now use inits where this isn't the case.

I realize Apache is.... Apache, but I wouldn't have put it in the image if I could think of any other way of having it control the server process...

kaysond:
Also it does add a lot of value when you do turn it on! And when you don't the only cost is slight increase in image size

itzg:
I think we'll have to agree to disagree. I would share more of my thoughts on Docker architecture and container lifecycle management via existing Docker/Kubernetes APIs...but I get the impression you would disagree.

I'm afraid I can't accept the PR as currently architected.

kaysond:
> I think we'll have to agree to disagree. I would share more of my thoughts on Docker architecture and container lifecycle management via existing Docker/Kubernetes APIs...but I get the impression you would disagree.

Please do. I'm open to it.

Issue #1556, opened 2022-06-08
Unpin tests from 1.18.2 to LATEST


itzg:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

Follow up to VERSION pinning done in https://github.com/itzg/docker-minecraft-server/pull/1555

PR #1566, opened 2022-06-15
Added support for Pufferfish 1.19

DiscowZombie:
Added support for Pufferfish 1.19

Compilation & execution have been tested.

### DIFF
--- a/scripts/start-deployPufferfish
+++ b/scripts/start-deployPufferfish
@@ -8,7 +8,7 @@ isDebugging && set -x
 
 IFS=$'\n\t'
 
-if ! versionLessThan 1.17; then
+if versionLessThan 1.17; then
   log "ERROR: Pufferfish server type only supports versions 1.17, 1.18 or 1.19, use PUFFERFISH_BUILD to select the the correct build 47 => 1.18.1, 50 => 1.18.2 etc"
   exit 1
 fi
--- a/scripts/start-deployPufferfish
+++ b/scripts/start-deployPufferfish
@@ -8,8 +8,8 @@ isDebugging && set -x
 
 IFS=$'\n\t'
 
-if [[ "${MAJOR_VANILLA_VERSION}" != "1.18" ]] && [[ "${MAJOR_VANILLA_VERSION}" != "1.17" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/start-deployPufferfish
```diff
@@ -8,8 +8,8 @@ isDebugging && set -x
 
 IFS=$'\n\t'
 
-if [[ "${MAJOR_VANILLA_VERSION}" != "1.18" ]] && [[ "${MAJOR_VANILLA_VERSION}" != "1.17" ]]; then
-  log "ERROR: Pufferfish server type only supports versions 1.18 or 1.17, use PUFFERFISH_BUILD to select the the correct build 47 => 1.18.1, 50 => 1.18.2 etc"
+if [[ "${MAJOR_VANILLA_VERSION}" != "1.18" ]] && [[ "${MAJOR_VANILLA_VERSION}" != "1.17" ]] && [[ "${MAJOR_VANILLA_VERSION}" != "1.19" ]]; then
```
itzg, line 11: I don't think I had the `versionLessThan` function before, so I wonder if this could now be:
```suggestion
if ! versionLessThan 1.17; then
```

itzg (commented):
Thanks for addressing this. Can you see if that suggestion works?

itzg:
I think this will help with #1556

scripts/start-deployPufferfish
DiscowZombie: Using the `versionLessThan` function will accept string such as "1.18.2" while the download will fail because the `MAJOR_VANILLA_VERSION` should be `1.18` and `PUFFERFISH_BUILD` > 50. This may, in my opinion, mislead users (even I agreed that we should find a better solution than making a PR for each new major release).

itzg: I'm not sure if your concern is actually applicable. Did you try out the suggestion and it didn't work?

bram-pkg: I think it would be at least good to sort them by version.

DiscowZombie:
The code now uses the _versionLessThan_ function and has been tested with both 1.17, 1.18 and 1.19 !

itzg (approved):
I'm sorry I gave you the suggestion logic inverted.

PR #1570, opened 2022-06-16
Added support for optimized SIMD operations

DiscowZombie:
Pufferfish and its forks can use SIMD operation optimization when a JVM flag is given.


itzg (approved):
Nice. Thanks!

Issue #1562, opened 2022-06-13
packwiz bootstrap isn't found, container stops


envyniv:
### Describe the problem

issue in title, also I've tried adding the `-bootstrap-no-update` to both PACKWIZ_URL and to the symlink generated by the container, which solves the same issue on clients, but here it seems to have no effect

### Container definition

```
mc-modded:
    image: itzg/minecraft-server:java8
    ports:
      - 25566:25565
    environment:
      EULA: "true"
      TYPE: "forge"
      VERSION: "1.7.10"
      PACKWIZ_URL: "https://envyniv.github.io/modpack/pack.toml"
      MEMORY: 3G
      ENFORCE_WHITELIST: "true"
      WHITELIST: notfunny_,supercatge73,GIOgame99,ZORDASIX23
      OPS: notfunny_
      MAX_PLAYERS: 4
      OVERRIDE_WHITELIST: "true"
      MOTD: "Server §l§cPazzosgravatoanoinonpia§r Ufficiale 2.0 - Docker Update"
      ICON: "https://thumbs.dreamstime.com/b/funny-man-watermelon-helmet-googles-looks-like-parasitic-caterpillar-34231117.jpg"
      OVERRIDE_ICON: "true"
      #ENABLE_RCON: "true"
      #RCON_PASSWORD: "docker"
      FORCE_GAMEMODE: "true"
      #SNOOPER_ENABLED: "false"
      #MAX_TICK_TIME: 30000
      MODE: "survival"
      #VIEW_DISTANCE: 8
      #LEVEL_TYPE: "BIOMESOP"
      DEBUG: "true"
      DEBUG_EXEC: "true"
      ENABLE_AUTOPAUSE: "true"
      AUTOPAUSE_TIMEOUT_INIT: 300
      AUTOPAUSE_TIMEOUT_EST: 30
      GUI: "false"
      LOG_TIMESTAMP: "true"
    tty: true
    stdin_open: true
    restart: unless-stopped
    volumes:
      - ./modded-data:/data
```

### Container logs

```
envy-mc-modded-1   | + MAX_TICK_TIME=315360000000
envy-mc-modded-1   | + export MAX_TICK_TIME
envy-mc-modded-1   | + /autopause/autopause-daemon.sh
envy-mc-modded-1   | + isTrue false
envy-mc-modded-1   | + local oldState
envy-mc-modded-1   | ++ shopt -po xtrace
envy-mc-modded-1   | + oldState='set -o xtrace'
[...]

PR #1571, opened 2022-06-17
tried my best to fix `itzg/docker-minecraft-server/issues/1562`

envyniv:
due to my very limited knowledge of bash there may still be issues with my approach.
#1562


itzg (approved):
It looks good. Later I might refactor that into a common function.

PR #1573, opened 2022-06-18
Add ability to specify packwiz gh auth token

endigma:


endigma:
Blocked by https://github.com/packwiz/packwiz-installer-bootstrap/pull/5

itzg:
Why is an auth token needed? The linked PR doesn't provide any explanation either.

endigma:
Github API has an anonymous rate limit, after some queries it returns

```
{
    "message": "API rate limit exceeded for <ip>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
    "documentation_url": "https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"
}
```

The limit is like 60, and I run upwards of 10 minecraft servers off my single dedicated using infrared, so this limit gets hit reasonably quickly especially if a tenant breaks their config and it crashloops. I'd like to run with tokens at the docker level and have them provided to packwiz and the script's own curl commands so this doesn't happen, authenticated rate limit is like 5000.

I don't think this script quite works, some of the `jq` stuff needs to be meddled with below my change, but it's overall blocked by the packwiz `--bootstrap-update-token` opt being broken so I'll fix this up when I can actually test against working packwiz.

itzg:
Thanks for clarifying.

Issue #994, opened 2021-08-06
add TYPE Limbo


DanielRamp:
I had the idea to add the server ``TYPE`` [Limbo](https://github.com/LOOHP/Limbo), because it's a standalone server (.jar) and not just any plugin etc.

_I'm not sure if I understand the code enough to implement it, but just wanted to add my idea here and maybe later implement/contribute (to) it._

## What is Limbo
A [Limbo Minecraft Server](https://github.com/topics/limbo-server) is just a standalone (non forge, bukkit, spigot, fabric, sponge...) mc server which is a for every player independent newly predefined generated world (as far as I understood). It can for example be used as a **afk server** or **if the server is full (queue)**.

If you need an example, there is the official __ of __ here (join in mc): ``mc.loohpjames.com``
Also the popular server ``Hypixel.net`` seems to have a limbo server, used as a afk space.

## Full Links
Spigotmc.org: https://www.spigotmc.org/resources/limbo-standalone-server-lightweight-solution-for-afk-or-waiting-rooms-in-your-server-network.82468/
Github repo: https://github.com/LOOHP/Limbo
His Jenkins build server: http://ci.loohpjames.com/job/Limbo/

itzg:
Speaking of the links, that server type would go well with the bungeecord image I also maintain. This server type sounds like it should be easy to add. I'm out of town visiting family for a few days so I'll get to it after I get back.

itzg:
Limbo is now supported by the latest image

https://github.com/itzg/docker-minecraft-server#running-a-limbo-server

DanielRamp:
Thank you for the implementation. I hope this answer is not "too late". I tried Limbo now twice but with no success (once with docker-compose, once with raw docker command).


The part that is a bit irritating to me, at least until I saw the code is that the file is named purpur-56.jar:
https://github.com/itzg/docker-minecraft-server/blob/f5ea43b3d3699783588eecd6545f3b7b75fddb58/start-deployLimbo#L29

command:
```docker
docker run --rm -it -p 25560:25565 -e EULA=true -e TYPE=LIMBO --name limbo-test itzg/minecraft-server
```
compose:
```docker-compose
version: "3.8"

#...

limbo1:
  image: itzg/minecraft-server:latest
  ports:
    - 25599:25565
  networks:
    - some-network-for-later
  environment:
    - TYPE=LIMBO
    - EULA=TRUE
  volumes:
    - ./limbo1-data:/data
[...]

PR #1604, opened 2022-07-16
Default -> default in limbo world name

gmemstr:
See #994 - Limbo will pass along the world name verbatim, but Minecraft (doesn't appear to) allow uppercase letters in world names.


itzg (approved):
Thanks!

PR #1606, opened 2022-07-16
Fix Oracle Linux zstd issue

NoaHimesaka1873:
OL8 has problem handing zstd without passing --use-compress-program


NoaHimesaka1873:
Hmm. PR is borked. closing for now.

PR #1607, opened 2022-07-16
Fixing zstd uncompression for Oracle Linux

NoaHimesaka1873:
OL has problem handing zstd, so fixing it with simple workaround.


itzg (approved):
Thanks for your super quick updates.

PR #1608, opened 2022-07-16
fix #1544

caycehouse:
This should fix #1544 by picking the directory path closest to $base_dir, rather than the first match.

PR #1614, opened 2022-07-18
Alpine build fix #1612

### LINKED ISSUE #1612: Alpine build fails with awk "Unexpected token" finding base of generic pack
### Describe the problem

Regression introduced by #1608 

### Container definition

n/a

### Container logs

https://github.com/itzg/docker-minecraft-server/runs/7378881866?check_suite_focus=true#step:7:292

caycehouse:
This should hopefully fix the regression introduced in #1608. I tested this with both the java8 and latest image variants.

### DIFF
--- a/scripts/start-setupModpack
+++ b/scripts/start-setupModpack
@@ -235,7 +235,7 @@ function genericPacks() {
       done
 
       # recalculate the actual base directory of content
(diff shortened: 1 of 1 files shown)


itzg (approved):
Thanks. I'm sorry though I should have mentioned I was already addressing it with a broader solution to replace `find`. I'll merge your MR, but heads up that I'll be replacing this whole line soon.

PR #1623, opened 2022-07-22
fix typo in README.md

KhooHaoYit:
The image name should be `itzg/minecraft-server` instead of `itzg:/minecraft-server`


itzg (approved):
Wow, that's a bad spot for a typo. Thanks for fixing!!!

PR #1644, opened 2022-08-03
Install procps-ng on Oracle Linux-based images for autopause

NoaHimesaka1873:
I haven't tested autopause so this issue went missing. Installing `procps-ng` will fix this issue.


itzg (approved):
Thanks!

PR #1647, opened 2022-08-03
Set container name in docker-compose config

kenwi:
I propose naming the container directly in the configuration such that when docker-compose is run, the name is "mc" and not "docker-minecraft-server_mc_1".


itzg:
An end user can and should configure their compose file however they want. 

Also, this file is not meant to serve as an example file -- it only exists because a particular vendor's product required it to exist.

If you provide further technical reasons it would be needed then this can be re-opened.

PR #1707, opened 2022-09-01
Add AUTOSTOP_USE_PKILL option to use pkill instead of kill in autostop feature

DAlperin:
In some nonstandard environments (see fly.io for an example) that the
server itself will be PID 1 is not a given. By allowing the user to
choose to use pkill instead they can ensure the autostop feature will
work even in those environments.

### DIFF
--- a/files/autostop/stop.sh
+++ b/files/autostop/stop.sh
@@ -6,4 +6,8 @@ if isTrue "${DEBUG_AUTOSTOP}"; then
 fi
 
 logAutostopAction "Stopping Java process"
-kill -SIGTERM 1
+if isTrue "${AUTOSTOP_PKILL_USE_SUDO:-false}"; then
+  sudo pkill -f --signal SIGTERM mc-server-runner
+else
+  pkill -f --signal SIGTERM mc-server-runner
+fi


files/autostop/stop.sh
```diff
@@ -6,4 +6,8 @@ if isTrue "${DEBUG_AUTOSTOP}"; then
 fi
 
 logAutostopAction "Stopping Java process"
-kill -SIGTERM 1
+if isTrue "${AUTOSTOP_USE_PKILL}"; then
+  sudo pkill -f --signal SIGTERM mc-server-runner
```
itzg, line 10: I'm wondering if pkill should just be used always since it'll work in the typical case. 

However, is sudo actually needed?

DAlperin: It is in the environment I am currently using. I could put it behind a flag though if you want?

itzg: Yeah, putting that behind a flag and then using the pkill approach unconditionally would be great.

DAlperin:
@itzg it now uses pkill by default, has a flag to use sudo with it, and rewrote commit to reflect new behavior 👍🏼 Let me know what you think

files/autostop/stop.sh
```diff
@@ -6,4 +6,8 @@ if isTrue "${DEBUG_AUTOSTOP}"; then
 fi
 
 logAutostopAction "Stopping Java process"
-kill -SIGTERM 1
+if isTrue "${AUTOSTOP_PKILL_USE_SUDO}"; then
```
itzg, line 9: I'm am gradually working on a documentation tool extract variables and defaults, so it'll be great if this is

```suggestion
if isTrue "${AUTOSTOP_PKILL_USE_SUDO:-false}"; then
```

itzg (commented):
One little tweak left

DAlperin:
Fixed!

itzg (approved):
Thanks!

PR #1792, opened 2022-10-18
if REMOVE_OLD_MODS = true redownload plugins

boyphongsakorn:
if REMOVE_OLD_MODS is True
set installedVersion = 0.0.0

### DIFF
--- a/scripts/start-spiget
+++ b/scripts/start-spiget
@@ -58,8 +58,11 @@ getResourceFromSpiget() {
         exit 2
       fi
 
-      # installedVersion=$(jq -r '.name' $versionfile)
-      installedVersion=0.0.0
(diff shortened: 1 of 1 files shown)


boyphongsakorn:
> Please provide a description on PRs. I can't review this until I know what problem was being solved.

ok, i add it.

boyphongsakorn:
> why are you proposing this change? Container config attempted? Expected results? Opening an issue first for discussion would have been good.

when i environment variables `REMOVE_OLD_MODS` is `True` or manually remove plugin, it remove plugin but not redownload plugins because `version meta data is still` and it see same version in spigot/spiget, it will not redownload

itzg:
Thank you. That is a good reason for the change.

itzg (approved):
Sorry for the delay getting back to this. The changes look great for the problem solved.

Issue #1801, opened 2022-10-25
I thought it should work


boyphongsakorn:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

after i PR this #1792 
I thought it should work, so i didn't try.

```
if [[ ${SPIGET_RESOURCES} ]]; then
  if isTrue "${REMOVE_OLD_MODS:-false}"; then
    removeOldMods /data/plugins
    REMOVE_OLD_MODS=false
  fi

  log "Getting plugins via Spiget"
  IFS=',' read -r -a resources <<<"${SPIGET_RESOURCES}"
  for resource in "${resources[@]}"; do
    getResourceFromSpiget "${resource}"
  done
fi
```
https://github.com/itzg/docker-minecraft-server/blob/5e2553274fc8b733dda6cbf9f27b8c59f0e609dc/scripts/start-spiget#L124-L135

but after removeOldMods, is that set REMOVE_OLD_MODS back to false?

itzg:
Please set `DEBUG=true` and `DEBUG_HELPER=true` on your container and provide those logs. The inclusion here should have caused the version files to removed in the first place:

https://github.com/itzg/docker-minecraft-server/blob/5e2553274fc8b733dda6cbf9f27b8c59f0e609dc/scripts/start-spiget#L11

boyphongsakorn:
```
2022-10-25T13:48:27.897423762Z + removeOldMods /data/plugins
2022-10-25T13:48:27.897436432Z + '[' -d /data/plugins ']'
2022-10-25T13:48:27.897459021Z + log 'Removing old mods including:*.jar excluding:_'
```

🤔 hmmm something wrong?

boyphongsakorn:
ok i forget to remove REMOVE_OLD_MODS_INCLUDE env set 😂

PR #1802, opened 2022-10-25
set REMOVE_OLD_MODS back to false after loop download plugin done

boyphongsakorn:
from this issue https://github.com/itzg/docker-minecraft-server/issues/1801#issue-1422104661
so i think after removeOldMods it set REMOVE_OLD_MODS back to false

https://github.com/itzg/docker-minecraft-server/blob/5e2553274fc8b733dda6cbf9f27b8c59f0e609dc/scripts/start-spiget#L124-L135

and when call function getResourceFromSpiget is not gonna redownload because of it

https://github.com/itzg/docker-minecraft-server/blob/5e2553274fc8b733dda6cbf9f27b8c59f0e609dc/scripts/start-spiget#L61-L75

### DIFF
--- a/scripts/start-spiget
+++ b/scripts/start-spiget
@@ -124,7 +124,6 @@ downloadResourceFromSpiget() {
 if [[ ${SPIGET_RESOURCES} ]]; then
   if isTrue "${REMOVE_OLD_MODS:-false}"; then
     removeOldMods /data/plugins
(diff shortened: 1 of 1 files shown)


itzg (changes requested):
Need container logs from https://github.com/itzg/docker-minecraft-server/issues/1801#issuecomment-1290525114

PR #1816, opened 2022-11-10
Add option to skip generic pack checksum

ChipWolf:
Currently, there is an option `SKIP_GENERIC_PACK_UPDATE_CHECK`. The documentation alludes to this reducing time due to the lengthy SHA1 checksum process; however, the option just skips the generic pack being applied entirely.

In the spirit of backward compatibility, I've left that option as-is while adding a new option `SKIP_GENERIC_PACK_CHECKSUM` that simply skips the checksum.

--

As a side note, I'm not seeing where the checksum is actually used now. The function appears to apply the pack regardless of the sum file contents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -815,7 +815,9 @@ GENERIC_PACKS_SUFFIX=.zip
 
 would expand to `https://cdn.example.org/configs-v9.0.1.zip,https://cdn.example.org/mods-v4.3.6.zip`.
 
-If applying large generic packs, the update check can be time-consuming since a SHA1 checksum is compared. To skip the update check set `SKIP_GENERIC_PACK_UPDATE_CHECK` to "true". Conversely, the generic pack(s) can be forced to be applied without comparing the checksum by setting `FORCE_GENERIC_PACK_UPDATE` to "true".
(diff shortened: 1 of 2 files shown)


itzg (approved):
Thanks for adding this. I really appreciate the help.

To answer your question, it is evaluated here

https://github.com/itzg/docker-minecraft-server/blob/61592d5c3e78a238fd1376c26931cadb58f10405/scripts/start-setupModpack#L215

where `checkSum` is implemented here

https://github.com/itzg/docker-minecraft-server/blob/a1e4657e62d9fff68aa2de453ca33066e089c8ed/scripts/start-utils#L290

ChipWolf:
> Thanks for adding this. I really appreciate the help.
> 
> To answer your question, it is evaluated here
> 
> https://github.com/itzg/docker-minecraft-server/blob/61592d5c3e78a238fd1376c26931cadb58f10405/scripts/start-setupModpack#L215
> 
> where `checkSum` is implemented here
> 
> https://github.com/itzg/docker-minecraft-server/blob/a1e4657e62d9fff68aa2de453ca33066e089c8ed/scripts/start-utils#L290

Ah got it, awesome. Thanks again! 🚀

PR #1828, opened 2022-11-15
Add remove old files feature, deprecate remove old files

### LINKED ISSUE #1817: Remove old files
### Enhancement Type

Improve an existing feature

### Describe the enhancement

Currently, `REMOVE_OLD_MODS` only targets the mods and plugins directory.

https://github.com/itzg/docker-minecraft-server/blob/46a09ed33103074f9c956a8cdf08bed564e40738/scripts/start-setupModpack#L20

I propose we rename the variable to `REMOVE_OLD_FILES` and allow the `REMOVE_OLD_FILES_INCLUDE` to contain the path globs the user wishes to remove.

ChipWolf:
This PR mutates the existing `REMOVE_OLD_MODS` feature into a generic `REMOVE_OLD_FILES` utility.

To mitigate disruption, the former variables are deprecated and adapted for feature parity.

> **Warning**
> Needs testing - marked as a draft until the feature is validated

_Closes #1817_

### DIFF
--- a/scripts/start-setupModpack
+++ b/scripts/start-setupModpack
@@ -2,8 +2,11 @@
 
 set -e -o pipefail
 
-: "${REMOVE_OLD_MODS:=false}"
 : "${MODS_FILE:=}"
+: "${REMOVE_OLD_FILES:=false}"
+: "${REMOVE_OLD_FILES_DEPTH:=2} "
+: "${REMOVE_OLD_FILES_INCLUDE:={plugins,mods\}/*.jar,{plugins,mods\}/*-version.json}"
+: "${REMOVE_OLD_MODS:=false}"
 : "${REMOVE_OLD_MODS_DEPTH:=1} "
 : "${REMOVE_OLD_MODS_INCLUDE:=*.jar,*-version.json}"
 sum_file=/data/.generic_pack.sum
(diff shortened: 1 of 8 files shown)


scripts/start-setupModpack
```diff
@@ -2,8 +2,11 @@
 
 set -e -o pipefail
 
-: "${REMOVE_OLD_MODS:=false}"
 : "${MODS_FILE:=}"
+: "${REMOVE_OLD_FILES:=false}"
+: "${REMOVE_OLD_FILES_DEPTH:=2} "
+: "${REMOVE_OLD_FILES_INCLUDE:={plugins,mods\}/*.jar,{plugins,mods\}/*-version.json}"
```
itzg, line 8: These are name matching patterns not path matching. This isn't going to work as is. Instead `plugins` and `mods` need to be passed as the entry points.

ChipWolf: https://github.com/itzg/docker-minecraft-server/blob/d60e67e6f99318360fa765c23020e521a8221d34/scripts/start-utils#L215

https://github.com/itzg/mc-image-helper/blob/master/src/main/java/me/itzg/helpers/find/PathMatcherConverter.java#L11-L15

> [The following rules are used to interpret glob patterns:](https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystem.html#getPathMatcher%28java.lang.String%29)
>
> The { } characters are a group of subpatterns, where the group matches if any subpattern in the group matches. The "," character is used to separate the subpatterns. Groups cannot be nested.

ChipWolf: Oh, understood. I think it'd be useful here if we could glob paths in `--name`.

I can see where in `mc-image-helper` the tree is being walked through instead of `getPathMatcher`.

itzg: Sorry I should clarified in the first place. Yes, just needs an enhancement to mc-image-helper

ChipWolf:
@itzg I saw picomatch is in the runner now, is this progressable? 😊

Issue #1113, opened 2021-11-08
Allow mc-backup's PAUSE_IF_NO_PLAYERS to work properly


itzg:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

rcon port needs to be removed from knockd config

PR #1830, opened 2022-11-17
Add .paused file when server is paused

kylestang:
Implements the suggestion from docker-mc-backup [issue #43](https://github.com/itzg/docker-mc-backup/issues/43#issuecomment-1087984202) to add a `.paused` file when the server is paused so it doesn't get woken up to check the player count. I'll be opening another PR in docker-mc-backup to add a check for the file.

I believe this closes issue #1113 as well.


itzg (approved):
Thanks!

PR #1838, opened 2022-11-23
Add SERVER-START.sh as a possible entry script file name

P4sca1:
Used by https://www.curseforge.com/minecraft/modpacks/create-flavored


itzg (approved):
Thanks! So many ways to combine the words "server" and "start" 😀

BTW test failure seems to be an unrelated flaky download.

PR #1873, opened 2022-12-23
Fixed format as code for MOTD newline escape

torgeros:
There was a wrong indentation in the description on how to create a multiline MOTD. This way, the displayed was actually `\n` instead of `\\n`.


itzg (approved):
Thanks!

Issue #1891, opened 2023-01-10
Type of "AUTO_CURSEFORGE" fails to start 'All of Fabric 6' CurseForge server pack


Ogloppi:
### Describe the problem

_All of Fabric 6_ offers separate server pack downloads, thus I used the provided server pack download: https://www.curseforge.com/minecraft/modpacks/all-of-fabric-6/files/4325309
As seen from the logs, the Docker image can't find `manifest.json` it requires.

However, the server pack in question does not include any `manifest.json` file in its CurseForge download.
You can find a file called `server-setup-config.yaml` inside the downloaded server pack, and further inside this file it includes a link to the actual server modpack: `modpackUrl: https://media.forgecdn.net/files/4325/308/All-of-Fabric-6-1.2.2.zip`

Is the `server-setup-config.yaml --> modpackUrl` the actual link that I'm supposed to provide for `CF_PAGE_URL`?

### Container definition

In `docker-compose.yml`:

```
version: "2.1"
  services:
    minecraft_aof6:
      image: itzg/minecraft-server:java17
      container_name: minecraft_aof6
      volumes:
        - ./Minecraft/AoF6/data:/data
      networks:
        - minecraft
      environment:
        - TYPE=AUTO_CURSEFORGE
        - CF_BASE_DIR=/data
        - CF_PAGE_URL=https://www.curseforge.com/minecraft/modpacks/all-of-fabric-6/files/4325309
        - INIT_MEMORY="3G"
        - MAX_MEMORY="6G"
        - EULA=true
      ports:
        - "12345:25565"
      restart: unless-stopped
```

Start container with `docker-compose up -d minecraft_aof6`

### Container logs

```
[init] 2023-01-10 18:01:30+02:00 Running as uid=1000 gid=1000 with /data as 'drwxrwxr-x 17 1000 1000 4096 Jan 10 18:00 /data'
[init] 2023-01-10 18:01:31+02:00 Resolved version given LATEST into 1.19.3 and major version 1.19
[init] 2023-01-10 18:01:31+02:00 Starting RCON commands
[init] 2023-01-10 18:01:31+02:00 Rcon cmds functionality enabled
[init] 2023-01-10 18:01:31+02:00 Resolving type given AUTO_CURSEFORGE
[mc-image-helper] 18:01:32.478 INFO  : Processing modpack All of Fabric 6 Server - 1.2.2 @ 695890:4325309
[...]

PR #1900, opened 2023-01-13
For AUTO_CURSEFORGE, provide the option to force include mods

itzg:
Brings in https://github.com/itzg/mc-image-helper/pull/96


itzg:
Should fix #1891

Issue #1904, opened 2023-01-14
AUTO_CURSEFORGE doesn't work for Vault Hunters due to ToroHealth and LegendaryTooltips being client-side mods


itzg:
Vault Hunters is still broken due to ToroHealth and LegendaryTooltips being client-side mods but not being labeled as such, so I assume your fix doesn't exclude them.

_Originally posted by @DrWhoCares in https://github.com/itzg/docker-minecraft-server/issues/1893#issuecomment-1382631819_
            
> I've added the line CF_EXCLUDE_MODS: "3556931,4035921" which should be the IDs for the two mods, going based off what's being output from the crash report in the logs, but they're still erroring. The docs simply say, "passing a comma or space delimited list of project IDs via CF_EXCLUDE_MODS"
Am I doing something wrong with the syntax? There's no examples in the readme. I'm using docker-compose specifically. Do I need to do anything like removing the image and re-pulling it or something like that?

(Further comments were posted there)

DrWhoCares:
I tried removing the image and stack entirely and trying again and no dice.
This is the docker-compose.yml file I'm using.
```yaml
version: '3.8'

services:
  mc:
    image: itzg/minecraft-server
    volumes:
    - ./modpacks:/modpacks:ro
    - data:/data
    environment:
      EULA: "true"
      TYPE: AUTO_CURSEFORGE
      CF_PAGE_URL: https://www.curseforge.com/minecraft/modpacks/vault-hunters-1-18-2
      CF_EXCLUDE_MODS: "3556931,4035921"
      CF_FORCE_SYNCHRONIZE: "true"
      OVERRIDE_SERVER_PROPERTIES: "true"
      DIFFICULTY: "hard"
      ALLOW_FLIGHT: "true"
      OPS: "DrWhoCares"
      MAX_PLAYERS: 16
      PVP: "true"
      MEMORY: "16G"
      GUI: "false"
    ports:
      - "25565:25565"
    stdin_open: true
    tty: true
    restart: unless-stopped
volumes:
  data: {}
  ```

**Comment by itzg** [comment]
I'm able to recreate the issue even though I see the exclusions correctly passed into image helper. I'll debug over there...

**Comment by itzg** [comment]
Looking closer at debugs I think I figured it out. Those IDs are the **file** not **project** IDs. In the debugs I see

```
Download/confirm mod torohealth-1.18-forge-2.jar @ 245733:3556931
```

So 245733 was the actual ID for torohealth.

[...]

PR #1906, opened 2023-01-14
AUTO_CURSEFORGE by default use exclude/include from latest in repo

itzg:


itzg:
README change would have helped #1904

PR #1919, opened 2023-01-23
add more rubidium mods

LeaderbotX400:
The following mods can be used in conjunction with rubidium, as such they prevent the server from loading


itzg (approved):
Thanks! I'll merge once the PR tests go through the motions -- nothing actually tests for the file yet.

Issue #1927, opened 2023-01-26
Updating a modpack with AUTO_CURSEFORGE doesn't seem to work properly


DrWhoCares:
### Describe the problem

VaultHunters updated a few days ago with v6 and v6H.
I restarted my server, not knowing this, and it seems to have attempted some sort of update, but it did it poorly it seems. Attempting to join the server gave an error about mod versions being different, and updating the client to use v6H allows things to work, somewhat.

I can join and do most things but this message appears immediately:
![javaw_wEGZ90lNyo](https://user-images.githubusercontent.com/12800223/214770027-55af9f44-a1a2-4a81-9fb6-8cc07c137981.png)
And attempting to enter a vault causes a ton of issues.
So I looked at the files and found out they're still using the v5 files. The log says that it already has v6H installed when booting up the server, but that doesn't seem to be the case.

What can I do to fix this?

### Container definition

```
version: '3.8'

services:
  mc:
    image: itzg/minecraft-server
    volumes:
    - ./modpacks:/modpacks:ro
    - data:/data
    environment:
      EULA: "true"
      TYPE: AUTO_CURSEFORGE
      CF_SLUG: "vault-hunters-1-18-2"
      CF_FILE_ID: "4360914"
      CF_EXCLUDE_MODS: "245733,532127"
      OVERRIDE_SERVER_PROPERTIES: "true"
      RCON_PASSWORD: "***"
      DIFFICULTY: "hard"
      ALLOW_FLIGHT: "true"
      OPS: "DrWhoCares"
      MAX_PLAYERS: 16
      PVP: "true"
      MEMORY: "16G"
      GUI: "false"
      SNOOPER_ENABLED: "false"
      ENABLE_AUTOPAUSE: "true"
      MAX_TICK_TIME: "-1"
    ports:
      - "25565:25565"
    stdin_open: true
    tty: true
    restart: unless-stopped
volumes:
  data: {}
```

### Container logs

```
I'll get the logs later, it's really late and I don't feel like spending the time to figure out where they are
```

itzg:
As for the error screenshot, that is something that needs to be resolved on the client side. That is a client screenshot right? I don't have experience upgrading modpacks on the client side, so you might need to search around for advice. 

Did you have CF_FILE_ID this whole time? If so, then the modpack publishers did a very bad thing by swapping out a existing file.

DrWhoCares:
Ah, you are right. Previously I was using CF_PAGE_URL.
[...]

PR #1929, opened 2023-01-26
For AUTO_CURSEFORGE, auto-set LEVEL from world save provided by modpack

itzg:
Also:
- Will help with #1927 and similar where the latest files in overrides always need to be written...but keeping world data untouched


itzg:
Also discussed at https://discord.com/channels/660567679458869252/660567682751528981/1067297541525348352

PR #1939, opened 2023-01-31
Modrinth per project version type declaration

### LINKED ISSUE #1871: For each specified modrinth project it should be possible to give a version type
### Enhancement Type

Improve an existing feature

### Describe the enhancement

Since minecraft mods in my opinion do not follow very consistent version types (release, beta, alpha), it should be possible to overwrite the env `MODRINTH_ALLOWED_VERSION_TYPE` for each project.

The reason behind this is, that even one mod, that is in alpha state will force the oders to may get downloaded in alpha state aswell, if one tries to only use `MODRINTH_PROJECTS` rather than the mods directory.

One suggestion for a format would be:
```
MODRINTH_PROJECTS: mod1@alpha,mod2@beta,mod3
MODRINTH_ALLOWED_VERSION_TYPE: release
```
where release type after the @ is the new minimum required version for this specific mod.

Another suggestion coming with this would be, that its possible to specify `MODRINTH_PROJECTS` in multiline with having whitespaces, like it is possible in itzg/bungeecord with `PLUGINS`, because the list can get pretty long then. For example:
```
MODRINTH_PROJECTS: |
  mod1@alpha,
  mod2@beta,
  mod3
MODRINTH_ALLOWED_VERSION_TYPE: release
```

JustMangoT:
Sorry for the slow PR

closes #1871

### DIFF
--- a/README.md
+++ b/README.md
@@ -793,7 +793,7 @@ For example, the following will auto-download the [EssentialsX](https://www.spig
                               |
                               +-- project slug
   ```
-  Also, specific version/type can be declared using colon symbol and version id/type after the project slug. The version id can be found at 'Metadata' section. Valid version types are ``RELEASE``, ``BETA`` ``ALPHA``. For instance:
+  Also, specific version/type can be declared using colon symbol and version id/type after the project slug. The version id can be found at 'Metadata' section. Valid version types are `release`, `beta`, `alpha`. For instance:
   ```
     -e MODRINTH_PROJECTS=fabric-api,fabric-api:PbVeub96,fabric-api:beta
   ```
--- a/README.md
+++ b/README.md
@@ -793,9 +793,9 @@ For example, the following will auto-download the [EssentialsX](https://www.spig
                               |
                               +-- project slug
   ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -793,9 +793,9 @@ For example, the following will auto-download the [EssentialsX](https://www.spig
                               |
                               +-- project slug
   ```
-  Also, specific version can be declared using colon symbol and version id after the project slug. For instance:
+  Also, specific version/type can be declared using colon symbol and version id/type after the project slug. The version id can be found at 'Metadata' section. Valid version types are ``RELEASE``, ``BETA`` ``ALPHA``. For instance:
```
itzg, line 796: I can't remember if that specific enum is processed case insensitive. In any case, good to be consistent with lowercase. 

```suggestion
  Also, specific version/type can be declared using colon symbol and version id/type after the project slug. The version id can be found at 'Metadata' section. Valid version types are `release`, `beta`, `alpha`. For instance:
```

itzg (changes requested):
Thanks! Just a small docs change I'd like.

itzg (approved):
Thanks for your help.

PR #1957, opened 2023-02-07
Add source dir override for `/plugins`, `/mods`, and `/config`

rhullah:
Requested [in Discord](https://discord.com/channels/660567679458869252/660567682751528981/1070715555767668747). Adds two environment variables to allow the override of the `/mods` and `/config` source directories.

- `COPY_MODS_SRC`
- `COPY_CONFIG_SRC`

Additionally adds a "setup only" test for this scenario.


rhullah:
I did not add one for `/plugins` source because there wasn't an existing one for the `/plugins` destination.  I can add a `COPY_PLUGINS_SRC` (and `COPY_PLUGINS_DEST`) if you feel it's needed.

itzg:
> I did not add one for `/plugins` source because there wasn't an existing one for the `/plugins` destination.  I can add a `COPY_PLUGINS_SRC` (and `COPY_PLUGINS_DEST`) if you feel it's needed.

Thanks for catching that. To be consistent with mods, it does make sense to add `COPY_PLUGINS_SRC`. 

FWIW, `COPY_MODS_DEST` seems like overkill now that I think about it.

itzg (commented):
Thanks! Looks great and I'll merge when the plugins variable is ready.

rhullah:
> Thanks for catching that. To be consistent with mods, it does make sense to add `COPY_PLUGINS_SRC`.

Would you like both `COPY_PLUGINS_SRC` and `COPY_PLUGINS_DEST`, or just the `SRC`?

rhullah:
I assumed both and added them.  If you only want `SRC`, I can remove `DEST`.

itzg (approved):
Thanks. Adding both was great for completeness.

PR #2013, opened 2023-03-17
Correcting and extending vanillatweaks example

ytterx:
The vanillatweaks example didn´t work, I got the following error:

mc_1  | [mc-image-helper] 12:41:54.892 ERROR : 'vanillatweaks' command failed. Version is 1.25.12
mc_1  | java.lang.NullPointerException: Cannot invoke "me.itzg.helpers.vanillatweaks.model.Type.toString()" because the return value of "me.itzg.helpers.vanillatweaks.model.PackDefinition.getType()" is null

When I checked the tests under tests/setuponlytests/vanillatweaks_file I noticed there were two missing entries in the JSON file.
I corrected this and added the examples for resourcepacks and craftingtweaks as well.

See also https://github.com/itzg/mc-image-helper/issues/163

### DIFF
--- a/examples/vanilla-tweaks/docker-compose.yml
+++ b/examples/vanilla-tweaks/docker-compose.yml
@@ -25,3 +25,19 @@ services:
       VERSION: ${MINECRAFT_VERSION:-LATEST}
       VANILLATWEAKS_SHARECODE: MGr52E
       REMOVE_OLD_VANILLATWEAKS: "TRUE"
+  vanillatweaks_file_datapacks_and_resourcepacks_and_craftingtweaks:
+    # port is set to 25567 to not conflict with vanillatweaks_file example
+    restart: "no"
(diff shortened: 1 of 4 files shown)


examples/vanilla-tweaks/docker-compose.yml
```diff
@@ -25,3 +25,19 @@ services:
       VERSION: ${MINECRAFT_VERSION:-LATEST}
       VANILLATWEAKS_SHARECODE: MGr52E
       REMOVE_OLD_VANILLATWEAKS: "TRUE"
+  vanillatweaks_file_datapacks_and_resourcepacks_and_craftingtweaks:
```
itzg, line 28: I'm not sure if I had noticed before this example file was one big list of services 😀

itzg (approved):
Thanks for updating that!

PR #2025, opened 2023-03-21
Fix autopause not working in java8

### LINKED ISSUE #2017: Autopause not working in java8 tag
Getting an error on startup:

    sudo: unable to execute /usr/local/sbin/knockd: No such file or directory
    [Autopause loop] Failed to start knockd daemon.
    [Autopause loop] Probable cause: Unable to attach to interface "eth0".
    [Autopause loop] Available interfaces within the docker container:
    [Autopause loop]   eth0
    [Autopause loop] Please set the environment variable AUTOPAUSE_KNOCK_INTERFACE to the interface that handles incoming connections.
    [Autopause loop] If unsure which interface to choose, run the ifconfig command in the container.
    [Autopause loop] Autopause failed to initialize. This log entry will be printed every 30 minutes.

Guessing one of the issues is:

https://github.com/itzg/docker-minecraft-server/blob/48414b3e87ccf524f55dd4e22e74826346f167e3/files/autopause/autopause-daemon.sh#L42

Going into the container gives:

    bash-4.4# which knockd
    /usr/local/sbin/knockd
    bash-4.4# which knock
    /usr/local/bin/knock

Metalcape:
Fixes #2017

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,11 +52,6 @@ RUN curl -fsSL ${MC_HELPER_BASE_URL}/mc-image-helper-${MC_HELPER_VERSION}.tgz \
   | tar -C /usr/share -zxf - \
   && ln -s /usr/share/mc-image-helper-${MC_HELPER_VERSION}/bin/mc-image-helper /usr/bin
 
-# Install modified knockd
-ADD https://github.com/Metalcape/knock/releases/download/0.8.1/knock-0.8.1-${TARGETARCH}${TARGETVARIANT}.tar.gz /tmp/knock.tar.gz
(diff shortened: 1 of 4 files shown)


Metalcape:
Actually, I could also rewrite the download commands to use curl. I can do that if it's better to avoid installing wget.

itzg:
Yeah, I was going to point out that I purposely don't install wget in order to reduce the number of "downloading" tools. I was going to suggest using `mc-image-helper get` but it is installed later in the dockerfile. Switching to curl will be great.

itzg (approved):
Thanks!

PR #2050, opened 2023-03-30
added information on the escaping of $ in the CF API Key

camalot:
Updated the readme for the section about the CF_API_KEY to include info about having to escape the `$` characters in the API Key.


itzg (approved):
Thanks!

Issue #2048, opened 2023-03-30
Add the possibility to pass `CF_API_KEY` via seperate file


feathecutie:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

I usually like to store my docker configuration (e.g. via docker-compose) in version control on services like GitHub. However, with the new `CF_API_KEY` system, this would require uploading a secret key in plain text, which I'd really like to avoid.

It would be great if it was possible to store the API key in a seperate file that I can exclude from version control and just provide its location in the container's file system (e.g. via a new env var `CF_API_KEY_FILE`), which is then read at runtime.

Another option besides a new env var would be defining a standard path for the file and first checking if a file at that path exists before falling back to `CF_API_KEY`.

If you're willing to accept such a feature, I'd be happy to try and implement it myself. I haven't taken too close of a look at the codebase yet, so I'm not sure if what I'm suggesting is even possible, but I don't see why it wouldn't be.

feathecutie:
Follow up: After doing more research (which I should have done beforehand), I've realized that both docker-compose and the NixOS module system (which I use for docker configuration) allow using `.env`-files instead of directly specifying environment variables, which would most likely solve the problem I described.

I'll leave this issue open because from a security standpoint, it still seems like a sensible idea, but I'd understand if this is seen as unnecessary

itzg:
In fact, I'll use this issue to track a change I wanted to make to those docs to mention and highly recommend the use of the `.env` file.

While you're at it, @camalot , want to add this info into the same block you just updated? The `.env` file might be a way to avoid escaping `$`'s since it probably is not subject to variable substitution.

itzg:
Resolved by #2054

Issue #1440, opened 2022-03-22
Move documentation to a docs site for navigability and accessibility


ChipWolf:
### Enhancement Type

Not sure

### Describe the enhancement

With the size the readme has become, it would benefit from becoming a micro-site using Mkdocs or similar.
This would be more scalable, accessible, and allow for better navigation between key concepts and features.

itzg:
I like the look of this one and use of markdown: https://squidfunk.github.io/mkdocs-material/

itzg:
@camalot thought I'd bring our conversation over here. 

I'm now thinking a Jekyll based site generator would be ideal since that'll plug into GitHub pages without additional build support 

https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#static-site-generators

itzg:
...and taking a few more minutes to read, I see that using custom site generators are fairly simple:

https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow

camalot:
Looks like this https://github.com/marketplace/actions/deploy-mkdocs can be used to directly deploy an MkDocs site directly to github pages with the material theme.

camalot:
Quick little test: https://camalot.github.io/mkdocs-test/ in this repo https://github.com/camalot/mkdocs-test

itzg:
That looks great @camalot ! I think we should go with that. The only I'm left pondering is how to transition over. 

A few sections at a time leaving major headers as a link over to new docs?

Do the whole thing at once with each major section as a "page" and then have one big link at the top that says where to find docs?

camalot:
Shouldn't take too long to do in initial convert. I have been taking each "Heading2" and making those into "pages" in this test repo. 

Maybe leave original readme there with a link to the docs indicating the transition, and that the doc site may be incomplete, so refer to the readme.

Can probably better organize the "pages" too, by "edition" or something. then you could technically have bedrock docs there too. 

example, I created `docs/deployment` and that has files for helm, aws, docker-compose and they are then referenced in the nav like this:

```yaml
    - "Deployment Templates and Examples":
      - "Helm Charts": "deployment/helmcharts.md"
      - "Amazon Web Services (AWS)": "deployment/aws.md"
      - "Docker Compose": "deployment/docker-compose.md"
```

which renders as
[...]

PR #2054, opened 2023-03-30
added docs to use env file

camalot:
documents #2048

### DIFF
--- a/README.md
+++ b/README.md
@@ -631,22 +631,16 @@ To manage a CurseForge modpack automatically with upgrade support, pinned or lat
 > ```shell
 > docker run ... -e "CF_API_KEY=\$11\$22\$33aaaaaaaaaaaaaaaaaaaaaaaaaa"
 > ```
-> You can also use an [environemnt file](https://docs.docker.com/compose/environment-variables/set-environment-variables/) to load your confguration values. Using an environment file will remove the need to escape your API Key.
->
+> You can use a `.env` file, in the same directory as your `docker-compose.yaml`, and define your API Key there to remove the need to escape the `$` in your API Key. The `.env` file is [loaded automatically by docker compose](https://docs.docker.com/compose/environment-variables/set-environment-variables/#substitute-with-an-env-file). 
+> 
+> *.env*
 > ```
-> TYPE=AUTO_CURSEFORGE
 > CF_API_KEY=$11$22$33aaaaaaaaaaaaaaaaaaaaaaaaaa
 > ```
 >
-> Configure docker-compose
-> ```yaml
-> minecraft:
->   env_file:
->   - ./minecraft-server.env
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -631,6 +631,23 @@ To manage a CurseForge modpack automatically with upgrade support, pinned or lat
...
+> ```
+> TYPE=AUTO_CURSEFORGE
+> CF_API_KEY=$11$22$33aaaaaaaaaaaaaaaaaaaaaaaaaa
+> ```
+>
+> Configure docker-compose
+> ```yaml
+> minecraft:
+>   env_file:
+>   - ./minecraft-server.env
```
itzg, line 645: Actually `.env` is a special file that doesn't even need to be called out specifically:

https://docs.docker.com/compose/environment-variables/set-environment-variables/#substitute-with-an-env-file

I'm leaning towards that one being documented with just the api key and showing all the other variables can be in the compose file like normal.

I can be easily convinced otherwise since `docker run` still needs the env file specifically mentioned.

camalot: that's a good point, i will indicate `.env`.

camalot:
Updated the doc to, hopefully, explain how to use a `.env`. 

Might want to consider changing for a single readme to use the wiki in github. can then break things down by server type, and such, so its not all in 1 huge file. but it does make it easier to find things with CTRL+F if you sort of know what you're looking for. ¯\_(ツ)_/¯

itzg:
> Might want to consider changing for a single readme to use the wiki in github. can then break things down by server type, and such, so its not all in 1 huge file. but it does make it easier to find things with CTRL+F if you sort of know what you're looking for. ¯\_(ツ)_/¯

Indeed! There's been [an open issue](https://github.com/itzg/docker-minecraft-server/issues/1440) forever that I momentarily think about and then get overwhelmed 😀

itzg (approved):
Excellent. Thanks again!

camalot:
> Indeed! There's been [an open issue](https://github.com/itzg/docker-minecraft-server/issues/1440) forever that I momentarily think about and then get overwhelmed 😀

If you have any idea of the structure, and what not, and want to create that wiki branch, i would be more than willing to help moving the docs.

itzg:
I was originally thinking of just using the wiki mechanism. What are your thoughts about that vs a documentation site generator like the I found for the issue: https://squidfunk.github.io/mkdocs-material/ ?

camalot:
I am not familiar with the mkdocs. I will look it over. high level, just looks like markdown files. only thing i can think of from a user standpoint is not having built-in navigation in github. but you could also just link to the docs, and make github specifically for the more technical side of it, like the code, issues, and discussions

itzg:
I have zero experience with that tool also 😀, so am more pondering the overall user experience of the "documentation site" tools vs github wiki.

PR #2061, opened 2023-04-03
Update readme for docs

camalot:
Added a link at the top of the readme to the new documentation site.

Should the rest of the readme be "cleared" out, or just contain some basic info, or stay as it is?


itzg:
It makes me nervous 😀, but does seem like clearing out the rest of the docs is the best way to avoid duplicate effort with updates. I'm thinking "Intro" would be the only section that would remain with perhaps your new link just below that. I wish I could remember where I generated this image that use in the Docker Hub README:  that

itzg:
> Maybe rename README for the time being. so that the full docs are available (in case anything was missed)

Great idea. Would you prefer I do the big switcheroo to make it "official"?

itzg (approved):
Forgot to actually merge this one...got distracted with the discussion I started 😀

PR #2064, opened 2023-04-04
Randomize the RCON password with image creation

### LINKED ISSUE #2002: RCon enabled by default with default password
Is it a great idea to have RCon enabled by default with the default password of 'minecraft'? Someone used that default configuration to join my server and grant themselves server op. Not a huge deal and I've dealt with it now but perhaps a more secure config by default would be a good idea?

camalot:
- generate a random password for RCON_PASSWORD during the build
- pass the value to RCON_PASSWORD as a build-arg
- update the tests to account for the random password
- update the test runner to pass in environment variables to the execution of `verify.sh`

Closes #2002


.github/workflows/build-multiarch.yml
```diff
@@ -12,7 +12,22 @@ on:
...
+      rcon_password: ${{ steps.passwdgen.outputs.password }}
+    steps:
+      # the setup for the rcon password is done outside of the matrix, so each 
+      # build uses the same generated password in all the image variants.
+      - name: Create RCON Password
+        uses: licenseware/generate-password-and-hash@v1
+        id: passwdgen
+        with: 
+          length: 20
   build:
```
itzg, line 28: From a CI perspective, I like the idea of this setup stage. However, I was expecting a simpler approach where the build job just included this step inline, which has the nice side effect of "randomizing" the passwords a bit more, one per image variant.

tests/setuponlytests/test.sh
```diff
@@ -46,7 +46,10 @@ setupOnlyMinecraftTest(){
     outputContainerLog "$logs"
     result=1
   elif [ -f verify.sh ]; then
-    if ! docker run --rm --entrypoint bash -v "${PWD}/data":/data -v "${PWD}/verify.sh":/verify "${IMAGE_TO_TEST:-itzg/minecraft-server}" -e /verify; then
+    # if there is an environment.sh, run it, and push the output to ./.tests.env. then load that as an env-file for the docker run
+    env_file=$([ -f environment.sh ] && sh ./environment.sh > ./.tests.env && echo "--env-file=./.tests.env" || echo "")
+    [ ! -z "${env_file// }" ] && echo "Using env-file argument: $env_file"
+    if ! docker run --rm --entrypoint bash $env_file -v "${PWD}/data":/data -v "${PWD}/verify.sh":/verify "${IMAGE_TO_TEST:-itzg/minecraft-server}" -e /verify; then
       endTime=$(date +%s)
```
itzg, line 53: This all seems quite complicated and as previously implemented it was necessary. Looking closer at this I realized I should have been using `docker compose run` for the `verify` also. So, I went with an idea there and confirmed it works:

https://github.com/itzg/docker-minecraft-server/commit/5df07696f839cc87198b38a20993f7f234d43705#diff-f8a6ba079e5e45526e46f87838c9d21b0caff96739fc36d21e3b41ad767f77faR49

If you're good with this, I can just push my changes into your PR branch.

tests/setuponlytests/defaults/environment.sh
itzg, line 1: This file can be removed if going with my proposed changes.

itzg:
...and here's the build testing my proposed changes

https://github.com/itzg/docker-minecraft-server/actions/runs/4632194211/jobs/8195981091

.github/workflows/build-multiarch.yml
camalot: One per image variant is fine too, i just thought to do it once and reuse that, but either way works, since the password doesnt really matter except for the tests.

tests/setuponlytests/test.sh
camalot: Oh, so when you use `compose run` the variables are available? that's great. way less complicated. 👍 yes, lets go with that for sure.

itzg:
Thank you so much for all the effort on this one, but I had an epiphany and realized how I could randomize the default at startup and still allow rcon-cli and mc-backup to work, via https://github.com/itzg/docker-minecraft-server/pull/2071 and https://github.com/itzg/docker-mc-backup/pull/134

Issue #1622, opened 2022-07-22
Update location of Loli catserver retrieval


itzg:
Looks like this server isn't a thing anymore. The GitHub page, as well as the JAR download seem to be gone.
See <https://github.com/Loli-Server/LoliServer>

__Originally posted by @Scrumplex in https://github.com/itzg/docker-minecraft-server/pull/1483#issuecomment-1192558011__

itzg:
It moved

https://github.com/itzg/docker-minecraft-server/pull/1483#issuecomment-1192905724

khankul:
I believe this issue can be closed now.

PR #2079, opened 2023-04-13
Remove LoliServer from README.md

khankul:
LoliServer is dead per #1622


itzg:
> LoliServer is dead per #1622

Not that I was aware. Last comment I noted was that it **moved** 

https://github.com/itzg/docker-minecraft-server/issues/1622#issuecomment-1193202607

khankul:
There's a latest build link in scripts/start-deployLoliserver, but it returns 404, and its builds seem to be no longer available. So I assumed since the project was moved and renamed, the old version of it is dead.

itzg:
So the link needs to change to pull from https://github.com/Luohuayu/CatServer/releases

khankul:
But, there already is an option to install CatServer.
Also, judging by https://github.com/Luohuayu/CatServer/blob/1.16.5/docs/README_EN.md,
"It will now continue to be maintained as a fork of LoliServer"
so the devs treat it as a fork rather than renamed project.

itzg:
Oh, sorry, I didn't even make the connection. These forks of forks of forks are very confusing.

itzg (commented):
Can you also remove the code for it
- delete https://github.com/itzg/docker-minecraft-server/blob/master/scripts/start-deployLoliserver
- remove the entry from https://github.com/itzg/docker-minecraft-server/blob/master/scripts/start-configuration

itzg (approved):

PR #2087, opened 2023-04-19
Base Java 19 image on focal

### LINKED ISSUE #2086: ERROR: could not locate path that contains java
### Describe the problem

Hi.

I'm trying to run a server on Java 19 but I'm getting an error.
All other versions of Java 8, 11, 17 work fine.

### Container definition

```
```

### Container logs

```
[init] ERROR: your Docker provider has an annoying flaw where it
[init] tries to set PATH even though the container establishes
[init] a very specific value.
[init] ERROR: could not locate path that contains java
```

itzg:
Resolves #2086

and is consistent with all the other Debian based images.

### DIFF
--- a/.github/workflows/build-multiarch.yml
+++ b/.github/workflows/build-multiarch.yml
@@ -35,7 +35,7 @@ jobs:
         include:
         # JAVA 19:
           - variant: java19
-            baseImage: eclipse-temurin:19-jre-jammy
+            baseImage: eclipse-temurin:19-jre-focal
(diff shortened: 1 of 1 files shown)


rhullah:
I don't want to throw a wrench in the works or change something just to change it but I was curious, is there anything specific in the Ubuntu images needed verses using the more bare Alpine images?

rhullah:
Looking closer, I see you have Alphine images for v8 and v17.  As an alternative you could just add one for v19.  More to support though, right?  😏

PR #2108, opened 2023-04-30
Download Canyon from GitHub for final build

Stealthii:
Canyon is on an indefinite hiatus, and a final build has been produced by @KoboDev for the meantime. This changes the default value for `CANYON_BUILD` to "final", and in the instance this is provided, we download this build from their GitHub provided build rather than the CI integration.

The CI server previously provided by Velocity is now 410 Gone, so individual build numbers are not available at present. A future PR may address this by them being made available by another provider, but for now I think this change is enough to keep Canyon servers functional for the meantime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,9 +541,10 @@ A [Catserver](http://catserver.moe/) type server can be used with
 
 > **NOTE** only `VERSION=b1.7.3` is supported. Since that version pre-dates the health check mechanism used by this image, that will need to be disabled by setting `DISABLE_HEALTHCHECK=true`.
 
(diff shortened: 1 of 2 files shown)


itzg (approved):
Thanks for the update.

KoboDev:
If I do find the other builds, I will provide them but for right now.
Its what I can give.
Build 6 being the last build before Tux left
and Build 26 & 27/latest being the last ones I have made for canyon.
When my brain is feeling the most healthiest I will continue pushing small patches.

KoboDev:
https://jenkins.glass-launcher.net/job/Canyon/
new canyon builds are hosted here, along with #32

PR #2109, opened 2023-04-30
Update Canyon CI link to Glass Launcher

Stealthii:
The team over at Glass have committed to providing updated builds for Canyon after the sunset of Velocity Jenkins.

This PR is a followup to #2108, which restores `lastSuccessfulBuild` as the default, downloading specific releases from the GitHub archive, and offering builds from the Glass CI server where not explicitly handled.


itzg (approved):
Cool. Thanks again.

PR #2158, opened 2023-05-29
WIP: Create captain-defenition for caprover users

dacadla:
captain-defenition is a file that caprover (a VERY good PaaS that you should check out) uses to be able to deploy it. You still need to set environment variables amnually though.

### DIFF
--- a/captain-defenition
+++ b/captain-defenition
@@ -1,0 +1,4 @@
+ {
+  "schemaVersion": 2,
+  "dockerfilePath": "./Dockerfile"
+ }


itzg (approved):
The addition of this file will enable this style of deployment?

https://caprover.com/docs/deployment-methods.html#automatic-deploy-using-github-bitbucket-and-etc

dacadla:
I believe so.

captain-definition
```diff
@@ -0,0 +1,4 @@
+ {
+  "schemaVersion": 2,
+  "dockerfilePath": "./Dockerfile"
```
itzg, line 3: I thought about it more and seems like it should use this

https://caprover.com/docs/captain-definition-file.html#use-image-name

to reference the pre-built images in which case just paste the four lines of definition file into the caprover UI.

itzg (commented):
I'm now feeling hesitant including this change as a top level file. At most I would like to see this as a docs item in https://github.com/itzg/docker-minecraft-server/tree/master/docs/misc/deployment

BTW with the dockerfilePath approach are you certain it even works since the image build assumes BuildKit/buildx support to inject platform build args? I don't see how caprover allows build args otherwise.

itzg:
> I believe so.

I was expecting a more confident answer. How are you presently using this definition file?

githubsaturn:
No, this file is not needed at all. In fact, This app is already available as a one click app in CapRover. See the latest PR making some minor changes:
https://github.com/caprover/one-click-apps/pull/935

`captain-definition` file is not needed since this docker image is already built and available on docker hub

itzg:
Thanks @githubsaturn and very good to know about the one-click apps repo. 

@dacadla want to add that 👆to the deployment docs I mentioned? Otherwise, I'll close out this PR.

PR #2162, opened 2023-05-31
Moving Java 19 to 20 and adding Alpine

shotah:
Bringing Alpine along for 1.19 and 1.20 MC releases

PR #2166, opened 2023-05-31
build: use multi-arch platforms in verify-pr

itzg:
Follow up to https://github.com/itzg/docker-minecraft-server/pull/2162

cc @shotah


itzg:
and that's probably why it was hard coded to amd64

https://github.com/itzg/docker-minecraft-server/actions/runs/5135698947/jobs/9241476137?pr=2166#step:4:121

PR #2182, opened 2023-06-05
Update the Vault Hunters example

rhullah:
Update the Vault Hunters example to:
- use the latest Update version
- match their MOTD
- use recommended memory size
- exclude client mod
- enabled needed server.properties settings
- comment about needed non-distributable mod(s)


itzg (approved):
Thanks!

PR #2186, opened 2023-06-06
Changed rcon password references from "minecraft" to randomly generated

### LINKED ISSUE #2180: docs(rcon): default password is not "minecraft"
### Describe the problem

The documentation states:

>The default password is "minecraft" but change the password before deploying into production by setting RCON_PASSWORD.

It looks like this was changed in https://github.com/itzg/docker-minecraft-server/pull/2071.

### Container definition

N/A

### Container logs

N/A

NexusNovaz:
I was able to find two files which referenced "minecraft" as the default password. I have changed this to say randomly generated. 

Should fix #2180


itzg (approved):
Thanks!

PR #2247, opened 2023-06-23
added support magma version tags

Two-Faces:
There is a problem with constant updates, which can theoretically break the server startup. I checked the launch of the desired version and it works fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -514,7 +514,7 @@ A [Magma](https://magmafoundation.org/) server, which is a combination of Forge 
     -e MAGMA_VERSION=9f3a3c25
 
 > **NOTE** there are limited base versions supported, so you will also need to  set `VERSION`, such as "1.12.2", "1.16.5", etc. 
-> To switch the MAGMA version, you need to take the name of the desired tag.
+> To switch the MAGMA version, you need to take the name of the desired tag from the magmafoundation git repository, [for example, to version 1.18](https://git.magmafoundation.org/magmafoundation/Magma-1-18-x/-/tags).
 
 
 ### Running a Mohist server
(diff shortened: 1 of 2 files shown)


Two-Faces:
The parameter, in theory, is not required and can be left empty in order to always take the latest version.

README.md
```diff
@@ -511,8 +511,10 @@ Extra variables:
 A [Magma](https://magmafoundation.org/) server, which is a combination of Forge and PaperMC, can be used with
 
     -e TYPE=MAGMA
+    -e MAGMA_VERSION=9f3a3c25
 
-> **NOTE** there are limited base versions supported, so you will also need to  set `VERSION`, such as "1.12.2", "1.16.5", etc.
+> **NOTE** there are limited base versions supported, so you will also need to  set `VERSION`, such as "1.12.2", "1.16.5", etc. 
+> To switch the MAGMA version, you need to take the name of the desired tag.
```
itzg, line 517: Can you rephrase "you need to take the name of the desired tag" and/or explain how one obtains that? As written I have no idea what you're talking about, but looking at the example I'm having to guess you mean the git revision of Magma.

scripts/start-deployMagma
```diff
@@ -6,7 +6,7 @@ isDebugging && set -x
 
 resolveVersion
 
-if ! downloadUrl=$(get --json-path '$.link' "https://api.magmafoundation.org/api/v2/${VERSION}/latest"); then
+if ! downloadUrl=$(get --json-path '$.link' "https://api.magmafoundation.org/api/v2/${VERSION}/latest/${MAGMA_VERSION}"); then
```
itzg, line 9: I know this script doesn't follow the pattern yet, but please do like most of the other scripts and pre-declare this variable and its default of being empty, such as

https://github.com/itzg/docker-minecraft-server/blob/a6c3943136f6c32e5a1e69488379917e330656c2/scripts/start-deployAutoCF#L7C6-L7C6

itzg (commented):
Thanks! Just a couple of small items.

itzg (approved):
Thanks for the quick updates

PR #2261, opened 2023-06-29
Allow skipping autopause/autostop using a file

Steveplays28:
I've added a check to the autopause/autostop daemons to see if the `/data/.skip-pause` (`/data/.skip-stop` for autostop) file is present, and if so, prevent autopausing/autostopping.

I've also added notes about this to the docs.

### DIFF
--- a/files/auto/autopause-daemon.sh
+++ b/files/auto/autopause-daemon.sh
@@ -62,7 +62,12 @@ do
     # Server startup
     if mc_server_listening ; then
       TIME_THRESH=$(($(current_uptime)+$AUTOPAUSE_TIMEOUT_INIT))
-      logAutopause "MC Server listening for connections - pausing in $AUTOPAUSE_TIMEOUT_INIT seconds"
+
+      if -e /data/.skip-pause ; then
+        logAutopause "`/data/.skip-pause` file is present - skipping pausing"
+      else
+        logAutopause "MC Server listening for connections - pausing in $AUTOPAUSE_TIMEOUT_INIT seconds"
+
       STATE=K
     fi
     ;;
@@ -71,6 +76,12 @@ do
     if java_clients_connected ; then
       logAutopause "Client connected - waiting for disconnect"
       STATE=E
+    else
+      TIME_THRESH=$(($(current_uptime)+$AUTOPAUSE_TIMEOUT_KN))
+      if -e /data/.skip-pause ; then
+        logAutopause "`/data/.skip-pause` file is present - skipping pausing"
+        STATE=E
+      fi
     else
       if [[ $(current_uptime) -ge $TIME_THRESH ]] ; then
         logAutopause "No client connected since startup / knocked - pausing"
@@ -81,10 +92,16 @@ do
     ;;
   XE)
     # Established
-    if ! java_clients_connected ; then
+    if -e /data/.skip-pause ; then
       TIME_THRESH=$(($(current_uptime)+$AUTOPAUSE_TIMEOUT_EST))
-      logAutopause "All clients disconnected - pausing in $AUTOPAUSE_TIMEOUT_EST seconds"
-      STATE=I
+      logAutopause "`/data/.skip-pause` file is present - skipping pausing"
+      STATE=E
+    else
+      if ! java_clients_connected ; then
+        TIME_THRESH=$(($(current_uptime)+$AUTOPAUSE_TIMEOUT_EST))
+        logAutopause "All clients disconnected - pausing in $AUTOPAUSE_TIMEOUT_EST seconds"
+        STATE=I
+      fi
     fi
     ;;
   XI)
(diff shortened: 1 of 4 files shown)


files/auto/autopause-daemon.sh
```diff
@@ -71,6 +76,12 @@ do
     if java_clients_connected ; then
       logAutopause "Client connected - waiting for disconnect"
       STATE=E
+    else
+      TIME_THRESH=$(($(current_uptime)+$AUTOPAUSE_TIMEOUT_KN))
+      if -e /data/.skip-pause ; then
+        logAutopause "`/data/.skip-pause` file is present - skipping pausing"
+        STATE=E
+      fi
```
itzg, line 84: I'm thinking this should just be a simple `elif` on the file existence and if present, leave the state as-is and not reset the overall `TIME_THRESH`. BTW a state of "E" means "established connections".

files/auto/autopause-daemon.sh
```diff
@@ -81,17 +92,29 @@ do
...
-      logAutopause "All clients disconnected - pausing in $AUTOPAUSE_TIMEOUT_EST seconds"
-      STATE=I
+      logAutopause "`/data/.skip-pause` file is present - skipping pausing"
+      STATE=E
+    else
+      if ! java_clients_connected ; then
+        TIME_THRESH=$(($(current_uptime)+$AUTOPAUSE_TIMEOUT_EST))
+        logAutopause "All clients disconnected - pausing in $AUTOPAUSE_TIMEOUT_EST seconds"
+        STATE=I
+      fi
```
itzg, line 104: I'm not sure why this logic needs to change at all. The state transition still makes sense to me even if pause is being skipped.

files/auto/autopause-daemon.sh
```diff
@@ -81,17 +92,29 @@ do
...
     # Idle
     if java_clients_connected ; then
       logAutopause "Client reconnected - waiting for disconnect"
       STATE=E
+    else
+      if -e /data/.skip-pause ; then
+        TIME_THRESH=$(($(current_uptime)+$AUTOPAUSE_TIMEOUT_EST))
+        logAutopause "`/data/.skip-pause` file is present - skipping pausing"
+        STATE=E
+      fi
```
itzg, line 117: Same comment here about a simple `elif`

itzg (commented):
Thanks! Theme of my comments is "keep it simple" 😀.

files/auto/autopause-daemon.sh
Steveplays28: Alright, I'll revert the changes I made to the `Established` section.

files/auto/autopause-daemon.sh
Steveplays28: Replaced it with an `elif`.

files/auto/autopause-daemon.sh
Steveplays28: Alright, I replaced this with an `elif` as well and removed the timer reset.

Steveplays28:
I implemented your suggestions, and added some brackets and `fi` statements I missed.
Let me know if anything else needs changing!

files/auto/autostop-daemon.sh
```diff
@@ -47,17 +56,27 @@ do
...
       TIME_THRESH=$(($(current_uptime)+$AUTOSTOP_TIMEOUT_EST))
-      logAutostop "All clients disconnected - stopping in $AUTOSTOP_TIMEOUT_EST seconds"
-      STATE=I
+      logAutostop "`/data/.skip-stop` file is present - skipping stopping"
+      STATE=E
+    else
+      if ! java_clients_connected ; then
+        TIME_THRESH=$(($(current_uptime)+$AUTOSTOP_TIMEOUT_EST))
+        logAutostop "All clients disconnected - stopping in $AUTOSTOP_TIMEOUT_EST seconds"
+        STATE=I
```
itzg, line 67: I don't believe this change is needed

files/auto/autopause-daemon.sh
```diff
@@ -92,6 +101,10 @@ do
     if java_clients_connected ; then
       logAutopause "Client reconnected - waiting for disconnect"
       STATE=E
+    elif [ -e /data/.skip-pause ] ; then
+      TIME_THRESH=$(($(current_uptime)+$AUTOPAUSE_TIMEOUT_EST))
```
itzg, line 105: Resetting the threshold doesn't seem necessary, but doesn't seem bad either.

itzg (commented):
Overall, I was wondering if you have been able to test the changes?

Steveplays28:
At the moment I haven't tested this yet. I can try cloning it onto my homeserver and trying it out though

files/auto/autopause-daemon.sh
Steveplays28: Alright, I'll keep it as is then.

files/auto/autostop-daemon.sh
Steveplays28: Reverted this change.

Steveplays28:
After the latest commit it now looks like this:
image

The log message is being sent every 10 seconds or so.

itzg (approved):

itzg:
Test failures were upstream Morinth 500 errors, so this is good to merge.

Steveplays28:
Alright, thanks for the reviews and merge!

PR #2320, opened 2023-08-06
Fix game name matching in setupForgeApiMods

pikajude:
Not sure how long this has been the case, but the response from the `version-types` endpoint of the CF API looks like this:

``` json5
{
  "data": [
    {
      "id": 73242,
      "gameId": 432,
      "name": "Minecraft 1.17",
      "slug": "minecraft-1-17"
    },
    {
      "id": 628,
      "gameId": 432,
      "name": "Minecraft 1.12",
      "slug": "minecraft-1-12"
    },
    // ...
}
```

### DIFF
--- a/scripts/start-setupForgeApiMods
+++ b/scripts/start-setupForgeApiMods
@@ -67,7 +67,7 @@ retrieveVersionTypeNumber(){
     exit 2
   fi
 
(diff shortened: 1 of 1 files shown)


itzg:
Actually, I'm about to rip out this whole "forge API" feature since it is now part of the CurseForge helper sub-commands:

https://github.com/itzg/docker-minecraft-server/pull/2322

itzg:
...thank you though for the fix, if that code was still active

PR #2328, opened 2023-08-10
Add Ngrok example

Atlas16A:
Added Example Compose for using ngrok with server to bypass networks that block port forwarding.

### DIFF
--- a/examples/docker-compose-ngrok.yml
+++ b/examples/docker-compose-ngrok.yml
@@ -18,7 +18,5 @@ services:
     image: shkoliar/ngrok:latest
     ports:
       - 4551:4551
-    links:
-      - mc
     environment:
       - PARAMS=tcp -region=us -authtoken=YourAuthTokenHere mc:25565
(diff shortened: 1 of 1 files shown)


examples/docker-compose-ngrok.yml
```diff
@@ -0,0 +1,24 @@
...
+    volumes:
+      # attach the relative directory 'data' to the container's /data path
+      - ./data:/data
+
+  ngrok:
+    image: shkoliar/ngrok:latest
+    ports:
+      - 4551:4551
+    links:
+      - mc
```
itzg, line 22: Did you happen to try without this? Docker compose network includes the name resolving of the other services by default, so `mc` should have already been resolvable.

itzg (approved):
Thanks! I'm good to merge this but was curious about my question first.

examples/docker-compose-ngrok.yml
Atlas16A: Id does appear to work without it, however I left it in as that's how the creator of the image suggested usage is setup, I assume its there for redundancy sake. Up to you though if ya wanna remove it.

itzg: Yeah, it would be great to remove it.

itzg (approved):

PR #2434, opened 2023-10-13
MOTD no longer needs to be pre-normalized

### LINKED ISSUE #2433: MOTD Sanitization Breaks Color Coding (using section mark symbol)
The code around [line 25 of `scripts/start-setupServerProperties`](https://github.com/itzg/docker-minecraft-server/blob/e5f818bdf1448930baef16cb914180002f93fb6d/scripts/start-setupServerProperties#L25) sanitizes the MOTD text, but that causes it to be entirely broken if it contains color or other speciak formatting. All section marks (`§`, the character used by Minecraft as an escape character for processing formatting) are replaced with `Â§` which seems to be something to do with UTF-8 vs UTF-16 encoding. However, this causes text to look very wrong in game, as seen in the image below. Here is an example of what I mean:

This MOTD is supposed to make rainbow text that says "Testing 123":
> `§c§lT§6§le§e§ls§a§lt§3§li§9§ln§5§lg §6§l1§e§l2§a§l3`

However, it gets converted to the following:
> `Â§cÂ§lTÂ§6Â§leÂ§eÂ§lsÂ§aÂ§ltÂ§3Â§liÂ§9Â§lnÂ§5Â§lg Â§6Â§l1Â§eÂ§l2Â§aÂ§l3`

In game, it appears like this:
image

itzg:
Resolves #2433


aaronjamt:
Looks like all the tests failed, but they appear to have failed due to networking issues (they are all failed HTTP requests, most of them failed with code 529). Can you re-run them and see if it works a second time round?

itzg:
I might need to disable the spiget test for now. It seems like the plugin chosen for testing is no longer provided by their CDN.

PR #2432, opened 2023-10-12
Fix example file name typo

Titaniumtown:
The example file for Spigot is spelt `spiget`. This PR fixes that.

PR #2446, opened 2023-10-19
Rename docker-compose-spiget.yml to docker-compose-spigot.yml

isErenG:
renamed spiget to spigot


itzg:
This is so weird. You're the second person this week, see #2432

It would be great if you could put a big comment at the top of the file that points users to https://docker-minecraft-server.readthedocs.io/en/latest/mods-and-plugins/spiget/

Or maybe this example file is more confusion than it helps and just needs to be deleted. What do you and @Titaniumtown think?

Titaniumtown:
Hilarious! Yea maybe this file can be deleted and the whole "SPIGET" stuff can be a commented out section in the spigot docker compose with a short blurb explaining it's functionality.

isErenG:
After submitting this pull request, I started diving into Spiget and how its an actual thing. That's so funny and confusing that there is two separate Spiget and Spigot. 

I guess it should stay the same way as it is technically correct and there are no typos. Clarifying things further could cause more confusion.

PR #2531, opened 2023-12-06
Add Discord Release Notification

rhullah:
Add an on-publish step to send a Discord notification when a release is published.

Additionally:
- removed old workflow names
- format the args a little nicer
- remove the redundant expression syntax on `if:`

I assume this will go to the `#ci` channel in Discord. Felt like `#releases` would make more sense, but I'm assuming that's just official Minecraft releases and integrated via Discord.

### DIFF
--- a/.github/workflows/discord.yml
+++ b/.github/workflows/discord.yml
@@ -10,7 +10,7 @@ on:
       - completed
 
 env:
- DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
+  DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
 
 jobs:
   discord:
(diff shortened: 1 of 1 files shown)


itzg (approved):
Looks great, thanks! Yes, in hindsight the releases channel would have been for set aside for image releases and I suppose is could be arranged as such. Let's keep it all in the ci channel until it bugs us and can split things out.

itzg:
That's a little annoying that Discord created embeds for the linked PRs. I wonder if surrounding them in angle brackets would suppress that?

https://discord.com/channels/660567679458869252/927361063572365352/1181809421060423770

Not a big deal though.

rhullah:
Man...you just said everything I was typing up! 😆 Yeah, only options I can think of would be to either tweak the release body to surround the links in `<>`, or see if that Action has any sort of option to to it itself. Do you have any thoughts?

rhullah:
Hmm, also noticed the HTML comment rendered in the Discord message too. Which means that'd need to be stripped out. I'm playing with something.

itzg:
Yeah, I'm actually hoping that GitHub action or an alternate one has an option to escape links in the body or disable embeds when posting. I vaguely remember that's an API option.

PR #2576, opened 2024-01-06
Allow Remote Console to be Enabled

Kaiede:
## Purpose 

This enables the work that was started in the mc-server-runner pull request to add a remote SSH console. 
https://github.com/itzg/mc-server-runner/pull/56

- Updates mc-server-runner to 1.10.0
- Adds `ENABLE_SSH` setting 

## Validation Performed

Built the docker container locally on an ARM64 system (M1 Mac mini with an Ubuntu VM running docker). Started a (Fabric) server with ENABLE_SSH set and the 2222 port exposed. Was able to login from a standard ssh client and issue commands to it. Docker logs were checked and confirmed that the output was sent to both locations and I could audit SSH connections. 

Discovered that the default password of 'minecraft' is never usable with the container due to the configuration scripts generating a scrambled password when RCON_PASSWORD is not defined.

Checked to ensure that the server still works when ENABLE_SSH is not defined, and logs behave as expected.


itzg (approved):
Awesome! Will merge monetarily.

PR #2591, opened 2024-01-14
Rename docker-compose-spiget.yml to docker-compose-spigot.yml

Foskcru:
All in name of PR


itzg:
It is named spiget with "e" on purpose. See #2446

Issue #2569, opened 2024-01-01
Option to use lazymc


savdbroek:
### Enhancement Type

A completely new feature

### Describe the enhancement

It would be great if this docker image could have lazymc as an option.

https://github.com/timvisee/lazymc

itzg:
Conceptually that's a good idea and was also brought up here

https://github.com/itzg/docker-minecraft-server/issues/2115#issuecomment-1542810309

It would be difficult to retrofit that into the existing design, but will park it here in case I can think about it more.

nfroeschl:
I have something similar running with [lazy_container](https://github.com/vmorganp/Lazytainer).

`version: "3"
services:
  lazytainer:
    container_name: lazytainer
    image: ghcr.io/vmorganp/lazytainer:master
    environment:
      - VERBOSE=false
    ports:
      - 25565:25565
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
    labels:
      - lazytainer.group.minecraft.sleepMethod=stop
      - lazytainer.group.minecraft.ports=25565
      - lazytainer.group.minecraft.minPacketThreshold=2
      - lazytainer.group.minecraft.inactiveTimeout=600 # 10 minutes, to allow the server to bootstrap. You can probably make this lower later if you want.
    restart: unless-stopped
    network_mode: bridge
  mc:
    image: itzg/minecraft-server
    environment:
      - EULA=TRUE
      - TYPE=PURPUR
      - MEMORY=4G
      - TZ=Europe/Berlin
      - MOTD=From_9_till_22
      - OVERRIDE_SERVER_PROPERTIES=TRUE
    volumes:
      - /opt/container_volumes/minecraft/data:/data
    labels:
      - lazytainer.group=minecraft
    depends_on:
      - lazytainer
    network_mode: service:lazytainer
    tty: true
    stdin_open: true
    restart: unless-stopped
networks: {}`

itzg:
@nfroeschl that would be a great section for https://github.com/itzg/docker-minecraft-server/blob/master/docs/misc/examples.md if you want to PR that.

joesturge:
[...]

PR #2598, opened 2024-01-17
Added Lazytainer Example

nfroeschl:
According to the request from [Option to use lazymc](https://github.com/itzg/docker-minecraft-server/issues/2569#issuecomment-1873373665) I created an example docker_compose.yml and some description.


itzg (approved):
Thanks!

PR #2620, opened 2024-01-31
Update the doc's Copyright year

Zyztem:
noticed when i was browsing files, that the docs copyright year was still 2023.


itzg (approved):
Thanks!

Issue #2645, opened 2024-02-08
ERROR : 'install-curseforge' command failed. Version is 1.37.10 java.lang.IllegalArgumentException: MALFORMED


NotJohnAfterall:
### Describe the problem

Hi, i have a problem with installing modpack on my server, i have installed already one without any big issues, but here idk what to do. Bellow is the error from log and docker-compose file. Thanks in advance, i am sorry if this is somewhere already answered, but i couldnt find it. :)



### Container definition

version: "3.8"                                                                                                                                                                                                                                                                                                                                                                                                                                                                            services:                                                                                                                                                                                                                                      mc:                                                                                                                                                                                                                                            image: itzg/minecraft-server:java8-jdk                                                                                                                                                                                                       tty: true                                                                                                                                                                                                                                    stdin_open: true                                                                                                                                                                                                                             ports:                                                                                                                                                                                                                                         - "25665:25665"                                                                                                                                                                                                                            environment:                                                                                                                                                                                                                                   EULA: "TRUE"                                                                                                                                                                                                                                 TYPE: "AUTO_CURSEFORGE"                                                                                                                                                                                                                      CF_API_KEY: "<API key is normally in the config, but i dont want to put it here>"                                                                                                                                                                CF_PAGE_URL: "https://www.curseforge.com/minecraft/modpacks/age-of-engineering/files/2503175"                                                                                                                                                DEBUG: "true"                                                                                                                                                                                                                                #ENABLE_AUTOPAUSE: "true"                                                                                                                                                                                                                    #AUTOPAUSE_TIMEOUT_EST: 300                                                                                                                                                                                                                  INIT_MEMORY: 2G                                                                                                                                                                                                                              MAX_MEMORY: 4G                                                                                                                                                                                                                              # CF_EXCLUDE_MODS: |                                                                                                                                                                                                                                                                                                                                                                                                                                                                     volumes:                                                                                                                                                                                                                                       # attach the relative directory 'data' to the container's /data path                                                                                                                                                                         - ./data:/data

### Container logs

[...]

Issue #1636, opened 2022-07-29
Autopause doesn't start


greyhammer:
### Describe the problem

Running Paper MC in docker with docker compose using latest version 1.19.1

I have configured the various items to enable Autopause, but the server never pauses and I never see anything related to Autopause past `[init] Autopause functionality enabled` in the logs

Nothing like `[Autopause loop] MC Server listening for connections - stopping in 600 seconds,` 

The server just runs forever

### Server Properties
```
#Minecraft server properties
#Fri Jul 29 14:37:12 UTC 2022
enable-jmx-monitoring=false
level-seed=1777181425785
rcon.port=25575
enable-command-block=true
gamemode=survival
enable-query=false
generator-settings={}
enforce-secure-profile=false
level-name=world
motd=Minecraft Server
query.port=25568
texture-pack=
pvp=true
generate-structures=true
max-chained-neighbor-updates=1000000
difficulty=easy
network-compression-threshold=256
max-tick-time=-1
require-resource-pack=false
max-players=20
use-native-transport=true
online-mode=true
enable-status=true
allow-flight=false
broadcast-rcon-to-ops=true
view-distance=10
max-build-height=256
server-ip=
resource-pack-prompt=
allow-nether=true
server-port=25568
enable-rcon=true
sync-chunk-writes=true
op-permission-level=4
server-name=Dedicated Server
prevent-proxy-connections=false
hide-online-players=false
resource-pack=
entity-broadcast-range-percentage=100
simulation-distance=10
player-idle-timeout=0
rcon.password=minecraft
force-gamemode=false
debug=false
rate-limit=0
hardcore=false
white-list=false
broadcast-console-to-ops=true
spawn-npcs=true
previews-chat=false
spawn-animals=true
snooper-enabled=true
function-permission-level=2
level-type=default
text-filtering-config=
spawn-monsters=true
enforce-whitelist=false
resource-pack-sha1=
spawn-protection=16
max-world-size=29999984
```

### Container definition

```
  minecraft:
    container_name: minecraft
    image: itzg/minecraft-server
    ports:
      - "25568:25568"
    restart: unless-stopped
    volumes:
      - "$DOCKER_DIR/minecraft/:/data"
    environment:
      TYPE: PAPER
      EULA: "true"
      MEMORY: "8G"
      SPIGET_RESOURCES: "274,18494"
      ENABLE_AUTOPAUSE: "TRUE"
      AUTOPAUSE_TIMEOUT_EST: "600"
[...]

PR #2646, opened 2024-02-08
auto-cf: support modpack zips using IMPLODE compression method

itzg:
Resolves https://github.com/itzg/docker-minecraft-server/issues/2645

Observed with https://www.curseforge.com/minecraft/modpacks/age-of-engineering/files/2503175

Also
- adds and uses by default the PlayerDB API for resolving whitelist and ops. Pulls in https://github.com/itzg/mc-image-helper/pull/393

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ RUN easy-add --var os=${TARGETOS} --var arch=${TARGETARCH}${TARGETVARIANT} \
   --var version=${MC_SERVER_RUNNER_VERSION} --var app=mc-server-runner --file {{.app}} \
   --from ${GITHUB_BASEURL}/itzg/{{.app}}/releases/download/{{.version}}/{{.app}}_{{.version}}_{{.os}}_{{.arch}}.tar.gz
 
-ARG MC_HELPER_VERSION=1.37.11
+ARG MC_HELPER_VERSION=1.37.12
 ARG MC_HELPER_BASE_URL=${GITHUB_BASEURL}/itzg/mc-image-helper/releases/download/${MC_HELPER_VERSION}
 # used for cache busting local copy of mc-image-helper
 ARG MC_HELPER_REV=1
--- a/docs/configuration/server-properties.md
+++ b/docs/configuration/server-properties.md
@@ -128,6 +128,11 @@ To change the behavior when the whitelist file already exists, set the variable 
 
 To [enforce the whitelist changes immediately](https://minecraft.wiki/w/Server.properties#enforce-whitelist) when whitelist commands are used , set `ENFORCE_WHITELIST` to "true". If managing the whitelist file manually, `ENABLE_WHITELIST` can be set to "true" to set the `white-list` property.
 
+!!! tip "Changing user API provider"
+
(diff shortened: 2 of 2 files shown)


NotJohnAfterall:
Thanks, i updated the image but now it throws an timed out error to the api server: 

[init] Running as uid=1000 gid=1000 with /data as 'drwxr-xr-x 5 1000 1000 4096 Feb  8 14:15 /data'                                                                                                                                           [mc-image-helper] 21:26:56.551 INFO  : Processing modpack 'Age of Engineering-1.1.2.zip' (age-of-engineering) @ 263897:2503175                                                                                                               
[mc-image-helper] 21:27:26.673 WARN  : [2eb28e2b-1, L:/<here is some ip 1> - R:api.curseforge.com/<here is some ip 2>] The connection observed an error                                                                                              io.netty.handler.timeout.ReadTimeoutException: null                                                                                                                                                                                          [mc-image-helper] 21:27:26.674 WARN  : [b856eba3-1, L:/<here is some ip 1> - R:api.curseforge.com/<here is some ip 2>] The connection observed an error                                                                                              io.netty.handler.timeout.ReadTimeoutException: null                                                                                                                                                                                          [mc-image-helper] 21:27:26.679 ERROR : Operator called default onErrorDropped                                                                                                                                                                io.netty.handler.timeout.ReadTimeoutException: null                                                                                                                                                                                          [mc-image-helper] 21:27:26.684 ERROR : 'install-curseforge' command failed. Version is 1.37.11                                                                                                                                               io.netty.handler.timeout.ReadTimeoutException: null                                                                                                                                                                                          [init] ERROR failed to auto-install CurseForge modpack

idk if this is just a some other server issue, but sadly it doesnt work again :)

itzg:
I'm thinking CurseForge was having some problems on their end. I noticed a lot of the image variants failed to build with similar error messages.

NotJohnAfterall:
Yeah i thought same thing, i will wait till tomorrow and see if anything will change.

itzg:
This build seems to be running better now

https://github.com/itzg/docker-minecraft-server/actions/runs/7835950318

NotJohnAfterall:
Ok, so i tried to remove docker image and run the server again and finally it works. Thank you so much for quick responses and very fast fix of the issue. I didn't expect this to be resolved so quickly, so thank you!

NotJohnAfterall:
If I may ask, I can't set auto-pause, I found this thread  #1636 , but I don't know if I'm completely stupid, but I can't find how to fix it. I have the same problem as the guy who described it there but I really don't know where the problem is. Sorry for asking such a stupid question.

itzg:
@NotJohnAfterall did you happen to manually change the server port in server.properties? I noticed in your original issue description you're using port 25665 rather than the default 25565. If that's the case then set this to match https://docker-minecraft-server.readthedocs.io/en/latest/configuration/server-properties/#server-port

In case this helps

https://raw.githubusercontent.com/itzg/docker-minecraft-server/master/notes/container-network-namespaces.drawio.png

NotJohnAfterall:
I've changed the port in server server.properties to 25665 and changed in docker-compose.yml server ports to be 25665:25565 (before i had server.properties to 25665 and docker-compose.yml server ports to be 25665:25665) , based on the image you sent. And it still doesnt work. The reason why i change ports is that, i have multiple servers running (this is second one :) )

itzg:
> The reason why i change ports is that, i have multiple servers running (this is second one :) )

That's the point I'm making with the image. You don't need to change the ports except for the one value that is the host port in the mapping. Leave all the other ports at 25565.

NotJohnAfterall:
Oh, i see. I did that and now it works. Thank you so much. Have a awesome day :)

PR #2679, opened 2024-02-20
feat: force include just-enough-resources-jer for create-arcane-engineering modpack

nohehf:
When creating a server for [create-arcane-engineering modpack](https://www.curseforge.com/minecraft/modpacks/create-arcane-engineering) I had to force include [just-enough-resources-jer](https://www.curseforge.com/minecraft/mc-mods/just-enough-resources-jer).
Here is a PR to add this to the defaults if you wish to, so that it's easier for people that might encounter the same issue :)

### DIFF
--- a/files/cf-exclude-include.json
+++ b/files/cf-exclude-include.json
@@ -51,14 +51,13 @@
   ],
   "modpacks": {
     "all-of-fabric-6": {
-      "forceIncludes": [
-        "revelationary"
-      ]
+      "forceIncludes": ["revelationary"]
     },
     "valhelsia-5": {
-      "excludes": [
-        "modernfix"
-      ]
+      "excludes": ["modernfix"]
+    },
+    "create-arcane-engineering": {
+      "forceIncludes": ["just-enough-resources-jer"]
     }
   }
 }


files/cf-exclude-include.json
```diff
@@ -51,14 +51,13 @@
...
     "all-of-fabric-6": {
-      "forceIncludes": [
-        "revelationary"
-      ]
+      "forceIncludes": ["revelationary"]
     },
     "valhelsia-5": {
-      "excludes": [
-        "modernfix"
-      ]
```
itzg, line 61: It would be helpful if you could revert these extra diffs.

itzg (changes requested):
Thanks! Extra diffs would be nice to revert, but let me know if that requires too much fighting the IDE/editor.

files/cf-exclude-include.json
nohehf: Yes my bad, we probably use a different formater. Pushing a fix. 
Thanks for the awesome work on this dockerfile btw, loving it

nohehf: Also, I wonder why is `just-enough-resources-jer` in the `globalExcludes` at first ?

itzg: It looks like I needed to exclude it for MC Eternal to startup correctly

https://github.com/itzg/docker-minecraft-server/pull/2143

I may have confused it with the "too many items" series of mods by putting it in the global excludes. Is it actually valid to run that mod on the server side?

itzg:
I just tried out the following and it started up fine. I'll plan on following up to this PR with an exclusion scoped to MC Eternal

```yaml
services: 
  mc:
    image: itzg/minecraft-server:java8-multiarch
    environment:
      EULA: true
      TYPE: FORGE
      VERSION: "1.12.2"
      CURSEFORGE_FILES: just-enough-resources-jer
      CF_API_KEY: ${CF_API_KEY}
    volumes:
      - mc:/data
      
volumes: 
  mc: {}
      
```

itzg (approved):

files/cf-exclude-include.json
nohehf: Yes this mod is valid on the server side, and even needed in some cases it seems

PR #2721, opened 2024-03-16
Fallback to existing server file when getbukkit retrieval fails

### LINKED ISSUE #2720: Spigot server is downloaded on start even when server.jar is present
### Describe the problem

I have a working spigot server with a `server.jar` and `spigot_server-1.20.4.jar` present in the data directory.

Even though `FORCE_REDOWNLOAD` is not set, there is an attempt to download the JAR file each time on server start. Since `getbukkit.org` currently is down, this attempt fails.

As a workaround, I have hosted `spigot_server-1.20.4.jar` on my own server and set `SPIGOT_DOWNLOAD_URL` to it. The suggested fix would be to not re-download the server if `server.jar` is present.

### Container definition

```
docker run -v ./test:/data \
    -e TYPE=SPIGOT -e VERSION=1.20.4 \
    -p 25599:25565 -e EULA=TRUE --name mc itzg/minecraft-server 
```

### Container logs

```
[init] Downloading Spigot from https://download.getbukkit.org/spigot/spigot-1.20.4.jar ...
```

itzg:
Resolves #2720

### DIFF
--- a/examples/docker-compose.yml
+++ b/examples/docker-compose.yml
@@ -1,6 +1,3 @@
-version: '3'
-# Other docker-compose examples in /examples
-
 services:
   minecraft:
     image: itzg/minecraft-server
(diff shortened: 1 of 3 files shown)


itzg:
Actually getbukkit seems to be completely shut down. I recommend you switch to Paper. 

image

itzg:
The fix I just merged will only be useful until a new version is needed.

PR #2749, opened 2024-03-31
Removed write to /var/log/knocked-source during autopause resume

itzg:
See if it helps #2746


torincb:
@itzg This removes the knock source from being logged - is it possible to restore this feature? Thanks!

itzg:
It never actually worked...as far as I noticed. Was it working for you?

torincb:
@itzg Well, not exactly &ndash; I only started using this project a couple weeks ago! But I have that same problem of unknown knocks to the server, so I dug and found the relevant change reverted here.

itzg:
It was always an empty string and writing to that /var file seemed to be a potential problem for some people. So, you're not missing out anything by this being "removed".

PR #2770, opened 2024-04-08
Update cf-exclude-include.json

rvandernoort:
add free-cam client mod to exclusion list as this was a problem with one of the modpacks I was using


itzg (approved):
Thanks

PR #2781, opened 2024-04-17
Add git-lfs package

BlownIntoSpace:
Also created a dummy gitconfig file at /etc/gitconfig so it does not need to be manually set at runtime


BlownIntoSpace:
I wasn't sure of the best place to generate the gitconfig file, but doing it alongside the git installation made the most sense as it only needs to be made at build time.

PR #2816, opened 2024-05-06
Remove version from docker compose since it's been deprecated

MASACR99:
Hi, I used one of your composes as a base and saw a warning that the version flag has been deprecated, I don't know if you'll want them removed from your examples but I went ahead and did that on my fork, so might as well create a pull request.

Cheers and thanks for your amazing work


itzg (approved):
Thanks! I was wondering about doing that myself, so this is a big help.

itzg:
Just making note for anyone confused about which "version" is meant here. It is the docker compose specification version itself:

https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-obsolete

Issue #2828, opened 2024-05-16
When proxy-protocol set to true, autostop no longer working.


ajarmoszuk:
### Describe the problem

When proxy-protocol is set to true inside config/paper-global.yml the autostop functionality is no longer working, once set to false we get "[Autostop loop] MC Server listening for connections - stopping in 600 seconds" inside the logs, otherwise it does not appear and does not seem to be working as designed.

Server version is Paper 1.20.4. Using newest available docker image. I've also tested with the newest Paper version and the same issue appears.

### Container definition

```
container = docker_client.containers.run(
        image='itzg/minecraft-server:latest',
        name=actual_server_name,
        detach=True,
        network="restricted-network",
        ports={'25565': f'{server_port}/tcp'},
        environment={
        'EULA': 'true',
        'TYPE': 'CUSTOM',
        'CUSTOM_SERVER': 'server.jar',
        'USE_AIKAR_FLAGS': 'true',
        'ENABLE_ROLLING_LOGS': 'true',
        'CREATE_CONSOLE_IN_PIPE': 'true',
        'MAX_PLAYERS': '10',
        'MEMORY': '1G',
        'INIT_MEMORY': '512M',
        'MAX_MEMORY': '1G',
        'ONLINE_MODE': 'false',
        'ENABLE_AUTOSTOP': 'TRUE',
        'AUTOSTOP_TIMEOUT_EST': '1800',
        'AUTOSTOP_TIMEOUT_INIT': '600',
        },
        volumes={
            f'/home/minecraft/servers/{actual_server_name}': {'bind': '/data', 'mode': 'rw'},
            '/home/minecraft/custom_files/server.jar': {'bind': '/data/server.jar', 'mode': 'ro'},
        },
        cpu_period=100000,  # Set the CPU period (microseconds)
        cpu_quota=150000, # 150% CPU Time
        mem_limit='2G',  # Memory limit
        memswap_limit='2G',  # Swap limit
    )
```

### Container logs

```
[init] Running as uid=1000 gid=1000 with /data as 'drwxrwxr-x 12 1000 1000 4096 May 16 23:13 /data'
[init] Autostop functionality enabled
[init] Resolving type given CUSTOM
[init] **********************************************************************
[init] WARNING: Some mods and modpacks may require Java 8.
[init]          Please use itzg/minecraft-server:java8
[init] **********************************************************************
[mc-image-helper] 23:30:08.902 INFO  : Created/updated 2 properties in /data/server.properties
[...]

PR #2832, opened 2024-05-18
Add mc_health_extra_args to autopause-fcns.sh functions

ajarmoszuk:
This adds ability that was non-working in issue #2828, allowing for use of undocumented environment variable MC_HEALTH_EXTRA_ARGS setting.

This functionality was originally in bin/mc-health but not in autopause-fcns.sh which is used to autostop and autopause, and which was non-working when for example proxy protocol acceptance was enabled in Paper.

### DIFF
--- a/files/auto/autopause-fcns.sh
+++ b/files/auto/autopause-fcns.sh
@@ -17,13 +17,13 @@ rcon_client_exists() {
 }
 
 mc_server_listening() {
-  mc-monitor status --host localhost --port "$SERVER_PORT" --timeout 10s >& /dev/null
+  mc-monitor status "${MC_HEALTH_EXTRA_ARGS[@]}" --host localhost --port "$SERVER_PORT" --timeout 10s >& /dev/null
 }
 
 java_clients_connections() {
   local connections
   if java_running ; then
-    if ! connections=$(mc-monitor status --host localhost --port "$SERVER_PORT" --show-player-count); then
+    if ! connections=$(mc-monitor status "${MC_HEALTH_EXTRA_ARGS[@]}" --host localhost --port "$SERVER_PORT" --show-player-count); then
       # consider it a non-zero player count if the ping fails
       # otherwise a laggy server with players connected could get paused
       connections=1


files/auto/autopause-fcns.sh
```diff
@@ -17,13 +17,13 @@ rcon_client_exists() {
 }
 
 mc_server_listening() {
-  mc-monitor status --host localhost --port "$SERVER_PORT" --timeout 10s >& /dev/null
+  mc-monitor status "${MC_HEALTH_EXTRA_ARGS[@]}" --host localhost --port "$SERVER_PORT" --timeout 10s >& /dev/null
```
itzg, line 20: The idea of this change is great, but I don't the think the syntax or usage is correct. Have you tested this?

`MC_HELATH_EXTRA_ARGS` is being used as an array type since normally it is declared in `/data/.mc-health.env` here

https://github.com/itzg/docker-minecraft-server/blob/7419a668431e5e13832122d7295d66f10883ae86/scripts/start-finalExec#L226-L229

however, the file `.mc-health.env` file doesn't seem to be getting source'd in the autopause scripts. So that needs to be added, such as [this](https://github.com/itzg/docker-minecraft-server/blob/dca29dcbb5d357811e5114b23f16e59126069a54/bin/mc-health#L5-L7)

Setting the env var `MC_HEALTH_EXTRA_ARGS` is not correct since env variables only have string values, not array values. So the array expansion `${MC_HEALTH_EXTRA_ARGS[@]}` will not be correct. 

I would suggest introducing a new env var named something like `USES_PROXY_PROTOCOL` that is "true" or "false". In the `start-finalExec` location above add a conditional on `isTrue "$USES_PROXY_PROTOCOL"` and add `--use-proxy` to `MC_HEALTH_EXTRA_ARGS` there.

ajarmoszuk: Yes, before making this pull request I've tested it with the --use-proxy option but that was about it.

ajarmoszuk: If you wish I can make the changes you mentioned, but it will probably be in about 1.5 weeks time from now as I'm working on other projects at the moment.

itzg: Yes, making those changes would be good. mc-health can also make use of that option then. I don't mind waiting for you to make that change.

BTW, `MC_HEALTH_EXTRA_ARGS` was purposely undocumented since it was only meant to communicate internally the start-time discovery for legacy ping protocol. So, it'll be great to document the new variable since it actually should be visible to users.

PR #2846, opened 2024-05-25
Added a missing quotation mark

JuanJesusAlejoSillero:
Added a missing quotation mark I spotted while reading the documentation.

Thanks for the great work!


itzg (approved):
Thanks!

PR #2859, opened 2024-05-27
rad2 compose example

noah1019:
I created a compose example for Roguelike Adventures and Dungeons with the proper excludes earlier for my server then reformatted it like the auto-cf example available in case anyone wants to use it for their own server. This is for the most recent version of RAD 2 1.11, it may require more excludes in the future.


itzg (approved):
Thanks!

PR #2867, opened 2024-06-01
Enhance CurseForge server packs that use variables.txt

rubbersteak:
The Curseforge server type has not been using the specified memory options. The startup logs (screenshot below) and server memory usage indicate that only 4GB is allocated regardless of the values set in the environment variables. Min 8GB, Max 16GB, and Aikar's flags have been enabled but the java args remain at -Xms4G -Xmx4G.

image

The issue is that the launch script used by /data/FeedTheBeast/start.sh uses the variables set in /data/FeedTheBeast/variables.txt to determine java launch options. The docker scripts set the desired launch options to variables but never write them to this file.

The proposed change replaces the current JAVA_ARGS variable in variables.txt with the ones defined within the script. Logs from the revised build show the desired result below.

image

### DIFF
--- a/scripts/start-finalExec
+++ b/scripts/start-finalExec
@@ -285,6 +285,9 @@ EOF
   if isTrue "${EXEC_DIRECTLY:-false}"; then
     "${finalArgs[@]}"
   else
+    JVM_ARGS="${JVM_XX_OPTS} ${JVM_OPTS} $expandedDOpts"
+    JVM_ARGS=${JVM_ARGS//$'\n'/}
+    sed -i "s~JAVA_ARGS=.*~JAVA_ARGS=\"${JVM_ARGS}\"~" "${FTB_DIR}/variables.txt"
     exec mc-server-runner "${mcServerRunnerArgs[@]}" "${finalArgs[@]}"
   fi
 elif [[ $SERVER =~ run.sh ]]; then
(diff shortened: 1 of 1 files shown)


scripts/start-finalExec
```diff
@@ -285,6 +285,9 @@ EOF
   if isTrue "${EXEC_DIRECTLY:-false}"; then
     "${finalArgs[@]}"
   else
+    JVM_ARGS="${JVM_XX_OPTS} ${JVM_OPTS} $expandedDOpts"
+    JVM_ARGS=${JVM_ARGS//$'\n'/}
+    sed -i "s~JAVA_ARGS=.*~JAVA_ARGS=\"${JVM_ARGS}\"~" "${FTB_DIR}/variables.txt"
```
itzg, line 290: Please add a conditional on variables.txt file existing. One big reason I deprecated support for server modpacks is that there was no guaranteed consistency. Some embed the equivalent in their start script.

itzg (commented):
Thanks for enhancing this. If you add the requested conditional I'm good to merge.

rubbersteak:
Conditional has been added. I have no idea how the other modpacks are structured and thought the server start sequence would be standard. Thanks for your work on this docker image.

itzg (approved):
Thanks for the update. Yes, you'd assume/expect consistency...but nope 😄

PR #2874, opened 2024-06-05
Add zume to cf-exclude-include.json

wirytiox:
i added zume, it was producing a crash on industrial-village modpack, the creator said that it was client only.

### DIFF
--- a/files/cf-exclude-include.json
+++ b/files/cf-exclude-include.json
@@ -1,5 +1,6 @@
 {
   "globalExcludes": [
+    "zume",
     "ambientsounds",
     "armor-toughness-bar",
     "biomeinfo",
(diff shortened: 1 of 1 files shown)


files/cf-exclude-include.json
```diff
@@ -1,5 +1,6 @@
 {
   "globalExcludes": [
+    "zume",
```
itzg, line 3: Sorry to be picky, but can you add it in the alphabetical spot for it?

itzg (commented):
Thanks for the update and research. It's good to go after the requested tweak.

files/cf-exclude-include.json
wirytiox: is it okay now? maybe we should add that as a comment to the file

PR #2899, opened 2024-06-13
Add true-darkness to CF global excludes

wirytiox:
added true-darkness because it is a client only mod. https://www.curseforge.com/minecraft/mc-mods/true-darkness


itzg (approved):
Thanks

PR #2915, opened 2024-06-19
Update compose file reference

emmanuel-ferdman:
# PR Summary
PR #2334 moved and renamed the compose file reference. This PR adjusts sources to changes.


itzg (approved):
Thanks! I'm afraid there's probably several links I have broken now with my gradual refactoring of examples. I'll be sure to watch out for that.

PR #2932, opened 2024-06-27
Update cf-exclude-include.json with Apple Skin, etc

LeaderbotX400:
## Added

- Apple Skin
- Better Third Person
- Better F3
- Better Advancements
- Falling Leaves
- Searchables
- Visuality


itzg (approved):
Thanks!

PR #2939, opened 2024-07-01
add check for absolute directories to world copy script

hpf3:
noticed that the "copy existing world" ability does not check if a level path is absolute, and appends it to /data/, added simple regex ( ^\/.*$ ) check to see if LEVEL is absolute


itzg (approved):
Thanks!

PR #2964, opened 2024-07-06
Updated mod exclusion list

KrosTheProto:
Added additional mods that cause dedicated server failure


itzg (approved):
Thanks!

files/cf-exclude-include.json
```diff
@@ -4,24 +4,29 @@
...
     "defensive-measures",
     "ding",
     "drippy-loading-screen",
     "dynamiclights-reforged",
+    "effective_fg",
     "embeddium",
     "embeddium-extension",
     "embeddium-extras",
     "enchantment-descriptions",
+    "entity_collision_fps_fix",
```
itzg, line 29: For future reference, these slugs need to match exactly as they are in CurseForge system

image

itzg (commented):
Turned it out there were a few I had to fix. I'm not sure why the PR test didn't run in this particular case -- it would have caught these.

https://github.com/itzg/docker-minecraft-server/pull/2965/files

PR #2967, opened 2024-07-07
More Client-side mods to exclude

KrosTheProto:
Client Mod exclusion 2 creeper boogaloo

Just a handful more I had listed out from previous modded servers I've made. 
probably last addition for a while from me.


itzg (approved):
Thanks again. Related to our Discord convo, I simply tested the changes locally where I could enable the API key.

Issue #2968, opened 2024-07-07
"Unable to resolve slug into ID (no matches)"


Cole-Mckay:
### Describe the problem

"foamfix-for-minecraft"s slug has appeared to have changed to "foamfix-optimization-mod" which is forcing the container to exit. When I try to exclude foamfix-for-minecraft nothing happens. 

![image](https://github.com/itzg/docker-minecraft-server/assets/113075534/0bf91e78-fb67-4444-bb36-5b379824c2d7)

![image](https://github.com/itzg/docker-minecraft-server/assets/113075534/444643df-7bf1-4ff7-8607-456e049c0f96)

### Container definition

```

services:
  mc:
    image: itzg/minecraft-server
    tty: true
    stdin_open: true
    ports:
      - "25565:25565"
    environment:
      VERSION: "1.20.1"

      SERVER_NAME: "Helium - Better MC"
      ICON: https://static.wikia.nocookie.net/minecraft_gamepedia/images/5/57/Helium.png/revision/latest?cb=20180620192736
      OVERRIDE_ICON: "TRUE"

      TYPE: "AUTO_CURSEFORGE"
      CF_API_KEY: ''
      CF_PAGE_URL: https://www.curseforge.com/minecraft/modpacks/better-mc-forge-bmc4/files/5418297
      CF_EXCLUDE_MODS: "foamfix-for-minecraft"
      MEMORY: "6GB"
      EULA: "TRUE"
    volumes:
      # attach the relative directory 'data' to the container's /data path
      - ./data:/data

```

### Container logs

```
[init] Running as uid=1000 gid=1000 with /data as 'drwxr-xr-x 5 1000 1000 4096 Jul  7 00:39 /data'
[mc-image-helper] 01:29:52.840 INFO  : Processing modpack 'Better MC [FORGE] 1.20.1 v28' (better-mc-forge-bmc4) @ 876781:5418297
[mc-image-helper] 01:29:53.015 ERROR : 'install-curseforge' command failed. Version is 1.39.3
me.itzg.helpers.errors.GenericException: Unable to resolve slug into ID (no matches): foamfix-for-minecraft
        at me.itzg.helpers.curseforge.CurseForgeApiClient.lambda$slugToId$7(CurseForgeApiClient.java:178)
        at java.base/java.util.Optional.orElseThrow(Unknown Source)
[...]

PR #2971, opened 2024-07-07
Remove JEI from exclusion list to fix ATM9 launch issue

AkashiSN:
https://github.com/itzg/docker-minecraft-server/issues/2968#issuecomment-2212320299
I confirmed that ATM9 launches normally locally after removing JEI from the exclusion list.


itzg (approved):
Thanks!

PR #2973, opened 2024-07-08
Add use-proxy support to health extra args

paradox460:
This fixes #2828, and resolves the concerns in #2838.

Specifically, it adds an env of `USES_PROXY_PROTOCOL`, and when present, passes the `--use-proxy` argument to mc-health

PR #2977, opened 2024-07-08
Revert "Add use-proxy support to health extra args"

itzg:
Reverts itzg/docker-minecraft-server#2973

PR wasn't tested and later confirmed not to work as intended.


itzg:
Follow up PR fixed this.

PR #2978, opened 2024-07-08
Update the autopause server listener to be proxy aware

paradox460:
I missed this in my previous PR, #2973. This makes it so the autostop monitor is proxy aware as well


itzg (approved):
Thanks for the follow up

PR #2979, opened 2024-07-10
Optimize startup on versionLessThan function

felipepasc:
Hi, how are you? I've been using the lib for a few months now, but I notice a slowdown when trying to start, and when analyzing and testing I realized that most of it is because of the versionLessThan function, so I redid it in bash script, did some tests and everything worked fine.

### DIFF
--- a/scripts/start-utils
+++ b/scripts/start-utils
@@ -176,6 +176,16 @@ function compare_version() {
         return 1
     fi
 
+    # Handle version channels ('a', 'b', or numeric)
+    if [[ $left_version == a* || $left_version == b* ]]; then
+        left_version=${left_version:1}
+    fi
+    
+    if [[ $right_version == a* || $right_version == b* ]]; then
+        right_version=${right_version:1}
+    fi
+
+
     local left_version_channel=${left_version:0:1}
     if [[ $left_version_channel =~ [0-9] ]]; then
         left_version_channel='r'
--- a/scripts/start-utils
+++ b/scripts/start-utils
@@ -226,6 +226,7 @@ function compare_version() {
 
     # Compare the versions using sort -V
     local result
+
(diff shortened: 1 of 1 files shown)


felipepasc:
Wtf error?

scripts/start-utils
```diff
@@ -161,9 +161,120 @@ function normalizeMemSize() {
...
+            ;;
+    esac
+
+    return $result
+}
+
 function versionLessThan() {
   # Use if-else since strict mode might be enabled
-  if mc-image-helper compare-versions "${VERSION}" lt "${1?}"; then
+  if compare_version "1.8" "lt" "${1?}"; then
```
itzg, line 277: Hardcoded "1.8"?

itzg (commented):
Ha, this used to be implemented in bash. The new implementation was fine until the log4j mess and then like you found it slowed down with all those version checks. 

I have a review comment. What did you mean by your last comment?

scripts/start-utils
felipepasc: sorry, i resolve this

felipepasc:
I had gotten a weird Docker error, so I just added a space to run it again and it worked, I changed it to the version variable to check

itzg (approved):
Looks great now. Ready for me to merge?

BTW, what kind of startup time improvements did you notice from the change?

felipepasc:
From my tests on my dedicated device, before the change, the startup message took around 17 seconds to arrive, after the change it dropped to 7 seconds

felipepasc:
This number may not be the same for everyone, some will feel the changes more and others less depending on the CPU and especially the speed of the ra memory.

felipepasc:
But everything is ok now I believe

PR #3008, opened 2024-07-21
Interpret backslash escapes for RCON_CMDS

### LINKED ISSUE #3007: Alternate separator for RCON_CMDS_
### Enhancement Type

Improve an existing feature

### Describe the enhancement

I'm trying to start the container as a Podman Quadlet. This means that it's just a normal systemd unit file, and there's no YAML involved, so I can't use `|-` quite as nicely as with Docker Compose for instance.

Since the scripts seem to use `while read -r` to go through the commands, I'm currently not able to pass multiple commands to run.

For example, this does not seem to work:

```
RCON_CMDS_STARTUP="gamerule disableElytraMovementCheck false\ngamerule playersSleepingPercentage 0"
```

systemd does seem to support newlines in such a way, but it's possibly the Podman integration here that interferes and this just goes in to the container verbatim:

```
# env | grep CMDS
RCON_CMDS_STARTUP="gamerule disableElytraMovementCheck false\ngamerule playersSleepingPercentage 0"
```

And then it fails like so:

```
[Rcon loop] Unknown or incomplete command, see below for error"gamerule disableElytraMovementCheck false\ngamerule playersSleepingPercentage 0"<--[HERE]
```

I'd be happy if we can turn the literal `\n` into an actual newline, so the current implementation works, or as the title suggest if there would be a way to separate multiple commands with another known separator that'd be nice too. :)

quulah:
Uses `echo -e` to interpret any included backslash escapes within `RCON_CMDS_`. Such as the literal `\n` newlines.

Fixes #3007

### DIFF
--- a/bin/rcon-cmds-daemon.sh
+++ b/bin/rcon-cmds-daemon.sh
@@ -44,7 +44,7 @@ do
       if [[ "$RCON_CMDS_STARTUP" ]]; then
         while read -r cmd; do
           run_command "$cmd"
-        done <<< "$RCON_CMDS_STARTUP"
+        done <<< "$(echo -e "$RCON_CMDS_STARTUP")"
       fi
       if 
         [[ -z "$RCON_CMDS_ON_CONNECT" ]] && 
@@ -66,7 +66,7 @@ do
         logRcon "First Clients has Connected, running first connect cmds"
         while read -r cmd; do
           run_command "$cmd"
-        done <<< "$RCON_CMDS_FIRST_CONNECT"
+        done <<< "$(echo -e "$RCON_CMDS_FIRST_CONNECT")"
     fi
 
     # When a client joins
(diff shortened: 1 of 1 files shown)


quulah:
`printf` might be more portable, it seems `echo` has varying implementations across shells and OS's. But this being a container and we have the `/bin/bash` shebang, I guess we can rely `-e` on being there.

itzg:
> `printf` might be more portable. It seems `echo` has varying implementations across shells and OS's. But this being a container and we have the `/bin/bash` shebang, I guess we can rely `-e` on being there.

Yeah, I had seen similar comments about portability, but yes, as you said, the container gives us a guaranteed bash implementation.

itzg (approved):
Thanks for the quick turn around! I'll merge this after the PR test finishes.

PR #3026, opened 2024-08-01
Added hide-online-players server property

EmilyxFox:
I think I've done what I need to add this server property.
`hide-online-players` was added in 1.18 21w44a and prevents sending of player list on status requests


itzg (approved):
Thanks!

PR #3032, opened 2024-08-05
Add just-zoom (jar: justzoom) to global exclude

EmilyxFox:
Here are the URLs just to double check the IDs

https://www.curseforge.com/minecraft/mc-mods/just-zoom
https://modrinth.com/mod/just-zoom

This mod prevents the server from starting since it's dependency Konkrete was added to the global exclude. It seems to be a wrongly declared client side mod.

### DIFF
--- a/files/modrinth-exclude-include.json
+++ b/files/modrinth-exclude-include.json
@@ -45,7 +45,7 @@
     "iris-flywheel",
     "ItemBorders",
     "ItemLocks",
-    "just-zoom",
+    "justzoom",
     "language-reload",
     "lazy-language-loader",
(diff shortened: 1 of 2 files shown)


files/modrinth-exclude-include.json
```diff
@@ -45,6 +45,7 @@
     "iris-flywheel",
     "ItemBorders",
     "ItemLocks",
+    "just-zoom",
```
itzg, line 48: Mentioned in discord also, but filename doesn't have dash

image

itzg (approved):
Thanks for the addition and thanks for helping out in discord.

PR #3034, opened 2024-08-05
New server properties

EmilyxFox:
I've sorted the [property-definitions.json](https://github.com/itzg/docker-minecraft-server/blob/master/files/property-definitions.json) alphabetically like the server.properties is sorted. I then went ahead and added:
```diff
+ resource-pack-id
+ resource-pack-prompt
+ max-chained-neighbor-updates
+ log-ips
+ region-file-compression
+ bug-report-link
```
I decided not to add `test-rainbow-chat` since that was only available for a few 1.19 snapshots. I can add that one if you want me to.


itzg (approved):
Thanks! I was also thinking it was time to sort those. Agreed with the logic to leave out `test-rainbow-chat`.

PR #3057, opened 2024-09-05
Simple Voice Chat examples

EmilyxFox:
I thought these examples might be useful for some people. Let me know if you like the format, comments, and configuration of these. I was also generally thinking of cleaning up the examples. Do you have any thoughts about that?


itzg (approved):
Thanks. The structure and style looks great. 

For the examples in general I'd like to get them organized more into subdirectories.

PR #3061, opened 2024-09-08
Bump mc-monitor version

md5nake:


itzg (approved):
Ah yes, no additional changes are needed since `SKIP_READINESS_CHECK` env var will already get picked up.

PR #3100, opened 2024-10-12
refactoring a little bit some test script files

IronMine:


itzg:
I'll review the changes more thoroughly, but so far I like what I see. Was there a particular problem you were solving? Or just wanted to contribute some code cleanup?

IronMine:
Thank you for the feedback! I’m eager to contribute more to this project. My main goal was to learn how it works, and I noticed some areas that could use a bit of code cleanup. I thought it would be a good way to start my first PR. I wasn’t addressing a specific issue; I just wanted to make a contribution and get involved.

Looking forward to your thoughts!

scripts/start-utils
```diff
@@ -85,11 +81,7 @@ function isFalse() {
 }
 
 function isDebugging() {
-  if isTrue "${DEBUG:-false}"; then
-    return 0
-  else
-    return 1
-  fi
+  $(isTrue "${DEBUG:-false}")
```
itzg, line 84: Should this just be the following, otherwise it'll try to execute the output of the isTrue operation
```suggestion
  isTrue "${DEBUG:-false}"
```

scripts/start-utils
```diff
@@ -162,106 +152,74 @@ function normalizeMemSize() {
...
-    fi
+  case "$comparison" in
+  lt) ((left_version_channel < right_version_channel)) && return 0 ;;
+  gt) ((left_version_channel > right_version_channel)) && return 0 ;;
+  le) ((left_version_channel < right_version_channel)) || ((left_version_channel == right_version_channel)) && return 0 ;;
+  ge) ((left_version_channel > right_version_channel)) || ((left_version_channel == right_version_channel)) && return 0 ;;
+  eq) [[ "$left_version" == "$right_version" ]] && return 0 ;;
+  *)
+    echo "Unsupported comparison operator: $comparison"
+    return 1
```
itzg, line 196: It looks like you have turned a lexical comparison into a numerical comparison. 

I'd prefer you didn't touch any of `compare_version` other than whitespace changes, since that was a contribution from someone else and the logic all seems very brittle. Now if you have done very exhaustive testing of this, then I would feel better about more changes.

scripts/start-utils
```diff
@@ -270,14 +228,10 @@ function versionLessThan() {
...
-    eval "$oldState"
-    return 0
-  else
-    eval "$oldState"
-    return 1
-  fi
+  eval "$oldState"
+
+  # Verify strict mode because it might be enabled
+  return $(compare_version "${VERSION}" "lt" "${1?}")
```
itzg, line 234: Shouldn't this be the following to avoid returning the output of the `compare_version`?

```suggestion
  compare_version "${VERSION}" "lt" "${1?}"
```

tests/setuponlytests/test.sh
```diff
@@ -44,39 +44,37 @@ setupOnlyMinecraftTest(){
   # false positive since it's used in delta calculations below
   # shellcheck disable=SC2034
   start=$(date +%s)
+  status=PASSED
```
itzg, line 47: I'm not able to see where `status` is set to "FAILED"

IronMine: You are right it was not

scripts/start-utils
IronMine: i revert and only simplify a return var that could be simplified

itzg (approved):
Thanks!

PR #3106, opened 2024-10-14
Fix mc-send-to-console running with uid=0(root)

GiriNeko:
I found an error in the conditional statement` if [ "$(id -u)" != 0 ]; then`. 
This condition incorrectly allows the ‘minecraft’ user to access the pipe `/tmp/minecraft-console-in` when the UID is 0 (i.e., when running as the root user), resulting in permission errors. 
This pull request corrects the erroneous check.

### DIFF
--- a/bin/mc-send-to-console
+++ b/bin/mc-send-to-console
@@ -17,7 +17,7 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
   exit 1
 fi
 
-if [ "$(id -u)" != 0 ]; then
+if [ $UID != 0 ]; then
   if [[ $(getDistro) == alpine ]]; then
     exec su-exec minecraft bash -c "echo '$*' > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
   else
--- a/bin/mc-send-to-console
+++ b/bin/mc-send-to-console
@@ -17,7 +17,7 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
   exit 1
 fi
 
-if [ "$(id -u)" = 0 ]; then
+if [ "$(id -u)" != 0 ]; then
   if [[ $(getDistro) == alpine ]]; then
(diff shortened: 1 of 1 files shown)


bin/mc-send-to-console
```diff
@@ -17,7 +17,7 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
   exit 1
 fi
 
-if [ "$(id -u)" = 0 ]; then
+if [ "$(id -u)" != 0 ]; then
   if [[ $(getDistro) == alpine ]]; then
     exec su-exec minecraft bash -c "echo '$*' > '${CONSOLE_IN_NAMED_PIPE:-/tmp/minecraft-console-in}'"
```
itzg, line 22: ...but you can't sudo to an arbitrary user from a non-root user

GiriNeko: oh you're right, i need thinkabout non-root user running.

GiriNeko:
i found the pipe `/tmp/minecraft-console-in` will grant by the environment `$UID` and `$GID`
the orignal if statement only can take the running user's id, so i changed.

bin/mc-send-to-console
```diff
@@ -17,7 +17,7 @@ if [ ! -p "${CONSOLE_IN_NAMED_PIPE}" ]; then
   exit 1
 fi
 
-if [ "$(id -u)" = 0 ]; then
+if [ $UID != 0 ]; then
```
itzg, line 20: Now understand what you were trying to solve. I think this is what will work for this scenario:

```suggestion
if [ "$(id -u)" = 0 -a $UID != 0 ]; then
```

This says "I'm starting out root (which is the default case) and I want to run the server as non-root user (the default case with UID=1000)".

GiriNeko: yeah, 
by default the server running as non-root user, and the command `mc-send-to-console` will execute by root, so its no problem.
but the pipe path `/tmp/minecraft-console-in` will be grant to root when the server running as root, so that if statement still to sudo execute an non-root user, and tooks error.

itzg (approved):
Thanks for fixing that.

PR #3116, opened 2024-10-20
port configuration for local modrinth modpack setup

joaogmbd:
missing port configuration on docker compose for local modrinth modpack setup


itzg (approved):
Thanks! Yeah when I'm spot testing  some configs I don't bother with port mappings 😄

Issue #3123, opened 2024-10-26
Error in /start-deploy-FTBA when setting up new docker on Raspberry Pi5


KroneBonn:
### Describe the problem

I have tried to setup a container with this image today on a new Raspberry Pi 5, but am left with this error:
`/start-deployFTBA: line 48: =/arm: No such file or directory`
I have tried several combinations of modpacks from FTB and java version but ultimately run in the same error.
As I am new to docker it may be my fault, but looking into the GitHub commits it appears that the line throwing the error is rather new and the curl-call was handled differently before. I am not an expert in shell skripting, but this might be the problem. Would be happy for help.

### Container definition

```
docker run -d --name mc-ftb -e EULA=TRUE   -e TYPE=FTBA -e FTB_MODPACK_ID=119  -p 25565:25565 -v /home/samuel/Docker/Minecraft/FTB_DW_21/  itzg/minecraft-server:java17
```

but also with different modpack/java version:

```
docker run -d --name mc-ftb -e EULA=TRUE   -e TYPE=FTBA -e FTB_MODPACK_ID=20   -p 25565:25565 -v /home/samuel/Docker/Minecraft/FTB_IE_S/   itzg/minecraft-server:java8
```

### Container logs

```
[init] Running as uid=1000 gid=1000 with /data as 'drwxr-xr-x 2 1000 1000 4096 Oct 24 12:20 /data'

[init] Downloading FTB installer

[init] Downloading ARM installer

/start-deployFTBA: line 48: =/arm: No such file or directory
```

itzg:
Yes your suspicion is mostly likely correct about that line probably being a new issue. I'll work on fixing that.

itzg:
Or @IronMine if you can get to it before me. 

I'm guess it's a regression from #3108

IronMine:
> Or @IronMine if you can get to it before me. 
> 
> I'm guess it's a regression from #3108

I am on it !

KroneBonn:
Thanks for the speedy replies! Very much appreciated!

IronMine:
> Thanks for the speedy replies! Very much appreciated!

i made a mistake i am sorry ! it's normal to be fast

itzg:
Fix will be in this build https://github.com/itzg/docker-minecraft-server/actions/runs/11532283059

PR #3124, opened 2024-10-26
fix start-deployFTBA

IronMine:
fixes : #3123

### DIFF
--- a/scripts/start-deployFTBA
+++ b/scripts/start-deployFTBA
@@ -45,7 +45,7 @@ if isTrue "$FTB_FORCE_REINSTALL" || ! [ -f "${ftbInstallMarker}" ] || [ "$(cat "
     log "Downloading FTB installer"
     if [ "$(uname -m)" == "aarch64" ]; then
       log "Downloading ARM installer"
-      $arm=/arm
+      $arm="/arm"
     else
       log "Downloading x86 installer"
     fi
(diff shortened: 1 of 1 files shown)


IronMine:
let me check just to be sure !

scripts/start-deployFTBA
```diff
@@ -45,7 +45,7 @@ if isTrue "$FTB_FORCE_REINSTALL" || ! [ -f "${ftbInstallMarker}" ] || [ "$(cat "
     log "Downloading FTB installer"
     if [ "$(uname -m)" == "aarch64" ]; then
       log "Downloading ARM installer"
-      $arm=/arm
+      $arm="/arm"
```
itzg, line 48: And remove that leading $ I think

(Some languages include that there so now I'm getting myself confused 😀)

IronMine: yes because i was doing some php ealier this day :/

IronMine:
really sorry for that i should have tested this ! But this makes me want to add a lot more tests ^^

itzg:
> really sorry for that i should have tested this ! But this makes me want to add a lot more tests ^^
> 
> 

No problem at all. Take a look at the setuponly tests if you're interested...however, it's a balancing act since too many of them ran fairly slow.

itzg (approved):

PR #3135, opened 2024-11-03
add ftb-evolution example file

EmilyxFox:
I'm not sure if I included too much information or too little. I could add a line about it applying to other modpacks as well, but it might be good enough to just be able to send this as an example when someone shows up with a similar issue.


itzg (approved):
Just the right amount of info. Thanks.

PR #3143, opened 2024-11-09
Create ATM10 example

rhullah:
I'm already maintaining it locally, might as well share it.

### DIFF
--- a/examples/atm10/docker-compose.yml
+++ b/examples/atm10/docker-compose.yml
@@ -1,0 +1,31 @@
+services:
+  mc:
+    image: itzg/minecraft-server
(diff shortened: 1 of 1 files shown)


itzg (approved):
Awesome, thanks. Will be nice to point people to the example since ATM10 seems to come up a lot.

rhullah:
Yeah, was thinking the same thing.

Also, while doing this it was the first time I needed to exclude the `overrides/shaderpacks` (due to an invalid character thing) but was wondering if excluding the `overrides/shaderpacks` should be a default thing? Would a server ever need the shaderpacks folder?

PR #3176, opened 2024-12-04
Fixed new FTB installer args and manifest

itzg:
Fix regression from https://github.com/itzg/docker-minecraft-server/pull/3174

Applies info from https://github.com/FTBTeam/FTB-Modpack-Issues/issues/6312#issuecomment-2516899020

### DIFF
--- a/scripts/start-deployFTBA
+++ b/scripts/start-deployFTBA
@@ -11,8 +11,30 @@ ftbInstallMarker=".ftb-installed"
 isDebugging && set -x
 set -e
 
-function getVersion() {
-  jq -r --arg name "$1" '.targets|unique[] | select(.name == $name) | .version' version.json
+#
+#{
+#  "id": 119,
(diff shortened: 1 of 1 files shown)


framerunner:
Currently closes the server, this is what it says in the logs:
```
[init] Running as uid=1000 gid=1000 with /data as 'drwxr-x--- 18 1000 1000 4096 Nov 15 19:48 /data'
[init] FTB modpack ID 126, version ID 100001 is ready to go
jq: error: Could not open file .manifest.json: No such file or directory
```

itzg:
Ah, I'm sorry, I broke the legacy installer support with that and I'll see if I can fix that. For now you can set FTB_FORCE_REINSTALL to "true".

itzg:
@framerunner this build should fix the issue https://github.com/itzg/docker-minecraft-server/actions/runs/12180940206

PR #3212, opened 2024-12-27
Add an Example of Setting Up GregTech: New Horizons

elizabethlfransen:
GregTech: New Horizons requires a little bit of extra setup because it's a generic pack.

This PR adds an example of how to configure it

### DIFF
--- a/examples/gtnh/docker-compose.yaml
+++ b/examples/gtnh/docker-compose.yaml
@@ -1,0 +1,23 @@
+services:
+  mc:
+    # make sure this java version matches with pack java version
+    image: itzg/minecraft-server:java21
+    tty: true
+    stdin_open: true
+    ports:
+      - "25565:25565"
+    environment:
+      EULA: "TRUE"
+      TYPE: CUSTOM
+      GENERIC_PACKS: GT_New_Horizons_2.7.1_Server_Java_17-21
+      GENERIC_PACKS_SUFFIX: .zip
+      GENERIC_PACKS_PREFIX: https://downloads.gtnewhorizons.com/ServerPacks/
+      # if this isn't true, then the container tries to download the modpack every run
+      : "true"
+      # Make sure that this matches what is in your pack's startserver bash file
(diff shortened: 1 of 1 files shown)


elizabethlfransen:
I would have liked to use [java options](https://docker-minecraft-server.readthedocs.io/en/latest/configuration/jvm-options/)  and [CUSTOM_SERVER](https://docker-minecraft-server.readthedocs.io/en/latest/types-and-platforms/server-types/others/#custom) but unfortunately CUSTOM_SERVER requires the jar file to exist before the generic pack is downloaded

elizabethlfransen:
I'm not sure if putting `@java9args.txt` in CUSTOM_JAR_EXEC will work, but I was able to run the server and connect to it.

itzg (approved):
Ah yes, that is quite particular. Thanks for providing that!

itzg:
> I would have liked to use [java options](https://docker-minecraft-server.readthedocs.io/en/latest/configuration/jvm-options/)  and [CUSTOM_SERVER](https://docker-minecraft-server.readthedocs.io/en/latest/types-and-platforms/server-types/others/#custom) but unfortunately CUSTOM_SERVER requires the jar file to exist before the generic pack is downloaded

Yeah I'll have to remember why I didn't relax that in the first place. It seems like it would be reasonable to relax it at least when GENERIC_PACK is used.

itzg:
@elizabethlfransen (or @SgtMate ), there's been [some discussion in Discord](https://discord.com/channels/660567679458869252/660569641550217327/1329543683103522847) about someone running into sporadic SEGSEGV java crash with GTNH and dropping down to java20 image seems to have fixed it. Since you posted the example with java21 I assume the answer will be no, but just wondering if you saw that type of crash with your use of this composition?

itzg:
Agreed, let's keep the example at java21 and yes it's easy for the image tag as people need on their particular systems. So many JVM bugs tend to be a combo of "strange" code and some particular hardware/OS/kernel quirk.

SgtMate:
Hey I tried it with the java21 version as well and did not run into this kind of issue.

PR #3213, opened 2024-12-28
fix vault hunters example

prabhjot98:
removed client side mode item-highlighter
made CF_API_KEY property more consistent with other examples 
set image tag to java17 for compatibility


itzg (approved):
Thanks! Really appreciate the consistency tweak.

Issue #3211, opened 2024-12-26
Unable to patch default config file `spigot.yml`


Krymonota:
### Describe the problem

Currently, it is not possible to patch default configuration files that are supposed to be located outside the `/data/config` directory, such as the `spigot.yml` file.

The `spigot.yml` file is downloaded as part of the default configuration files:
https://github.com/itzg/docker-minecraft-server/blob/05cbbfc9e3c15635d76368bf4e887fbcdc3a6dc1/scripts/start-deployPaper#L46-L48

Unfortunately, all default configuration files (so even if I add some more by providing the `DOWNLOAD_DEFAULT_CONFIGS` environment variable with additional files) are put in the `/data/config` directory:
https://github.com/itzg/docker-minecraft-server/blob/05cbbfc9e3c15635d76368bf4e887fbcdc3a6dc1/scripts/start-setupEnvVariables#L30-L38

However, only the `paper-global.yml` and `paper-world-defaults.yml` files belong to the `/data/config` directory. The `spigot.yml` file and all other default configuration files that you may want to patch must be put in the `/data` directory, which is not the case at the moment.

### Container definition

```sh
docker run -d -e DEBUG=true -e EULA=true -e TYPE=PAPER -e VERSION=1.21.1 -e PATCH_DEFINITIONS=/patches -v ./patches:/patches itzg/minecraft-server
```

File `patches/spigot.yml.patch.json`:
```json
{
  "file": "/data/spigot.yml",
  "ops": [
    {
      "$set": {
        "path": "$.settings.bungeecord",
        "value": true
      }
    }
  ]
}
```
This patch does not work because `/data/spigot.yml` does not exist.

File `patches/spigot2.yml.patch.json`:
```json
{
  "file": "/data/config/spigot.yml",
  "ops": [
    {
      "$set": {
        "path": "$.settings.bungeecord",
        "value": true
      }
    }
  ]
}
```
This patch works because the default `spigot.yml` file is put in `/data/config`, but it remains there and is not used by the server. Instead, the server generates a new file at `/data/spigot.yml`, as can be seen here (check `settings.bungeecord`):

```yml
root@c97b65306078 ~
❯ cat spigot.yml
# This is the main configuration file for Spigot.
# As you can see, there's tons to configure. Some options may impact gameplay, so use
# with caution, and make sure you know what each option does before configuring.
# For a reference for any variable inside this file, check out the Spigot wiki at
[...]

PR #3217, opened 2024-12-31
Download `paper.yml` and `bukkit.yml`

Krymonota:
Thanks a lot for resolving #3211 🙂

I've made some further improvements:
- Only attempt download `paper-global.yml` and `paper-world-defaults.yml` if the version is 1.19 or newer, as [these configuration files were introduced in 1.19](https://forums.papermc.io/threads/paper-1-19.344/).
- Download `paper.yml` if the version is older than 1.19.
- Always download `bukkit.yml`, which is also present in https://github.com/dayyeeet/minecraft-default-configs.

### DIFF
--- a/scripts/start-deployPaper
+++ b/scripts/start-deployPaper
@@ -42,10 +42,17 @@ else
   applyResultsFile ${resultsFile}
 fi
 
-# Download default configs to allow for consistent patching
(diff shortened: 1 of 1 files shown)


itzg (approved):
Awesome follow up. Thanks!

PR #3238, opened 2025-01-06
Fixed compose error due to missing key in GTNH example

SgtMate:
Docker compose errored out due to a missing key, fixed it by adding the key that should go there based on the comment above.


itzg (approved):
Thanks!

PR #3272, opened 2025-01-22
Add configurator tool to documentation

Krymonota:
I would say I have completed adding support for most of the image variables to my configurator tool https://setupmc.com/java-server/. I thought of mentioning the configurator in the documentation as it might be helpful to others - especially those new to the world of Docker. The “Using Docker Compose” section seemed like the most appropriate place in the documentation, but let me know if you would rather put it somewhere else.


itzg (approved):
Yeah, that seems like a great place for it.

Issue #3144, opened 2024-11-09
Support `*.jar.disabled` mods


rhullah:
### Enhancement Type

A completely new feature

### Describe the enhancement

Some launchers allow you to disable installed mods as a way to prevent it from being used and not having to fully remove it. I usually use this if I'm testing things and likely want to return to using the mod after tests. If I was to do this with the image now, it would just redownload the mod that was renamed `.jar.disabled`.

I was wondering the possibility of supporting this file naming as an alternative to the mod's expected filename? That way the system recognizes that the mood is there and doesn't redownload it.

itzg:
That's a great idea. I'll research the code around there but off hand I'm thinking that's quite doable.

itzg:
### Implementation notes

Manifest integrity check:

https://github.com/itzg/mc-image-helper/blob/ede488ce5d8e3b90334eb4414228ef8957def448/src/main/java/me/itzg/helpers/files/Manifests.java#L125

To handle "download to directory" mode's file existence check:

https://github.com/itzg/mc-image-helper/blob/0161ae3dd56d221695033f9dc1c999b8019dbc6c/src/main/java/me/itzg/helpers/files/ReactiveFileUtils.java#L33

Specific file download mode, which should probably be enhanced to use same as `me.itzg.helpers.http.OutputToDirectoryFetchBuilder#skipExisting(reactor.netty.http.client.HttpClientResponse, java.nio.file.Path)`

https://github.com/itzg/mc-image-helper/blob/df3c8e91de06345db06b9a4a87ecf7854af07aaf/src/main/java/me/itzg/helpers/http/SpecificFileFetchBuilder.java#L59

rhullah:
Excellent, thanks for looking into it!

azunaVT:
I'm not sure my need was the same here, I needed to disable a mod when downloading a generic pack (GTNH in my case), I've done a POC to allow disabling certain mods when downloading generic packs:
https://github.com/azunaVT/docker-minecraft-server/pull/1

@itzg is this something we'd be interested in having in this repo / image? If we could, that'd be great because I wouldn't have to rely on a locally built image or maintain a fork for my specific need 😄 I seems to work fine with manifest generation / checking. Otherwise, feel free to comment on the POC and let me know if there's any improvements / changes we'd need to get my changes into this repo!

itzg:
[...]

PR #3281, opened 2025-01-26
Feat: Disabling Mods in Generic Packs

azunaVT:
Adding the ability to disable mods for generic packs in `start-setupModpack` through a `DISABLE_MODS` environment variable.

Example of usage for something like running GTNH 2.7.2 with CrucibleMC, where `lwjgl3ify` and `archaicfix` need to be disabled for the pack to work:
```
  gtnh:
    image: itzg/minecraft-server
    container_name: gtnh
    tty: true
    stdin_open: true
    restart: unless-stopped
    environment:
      TYPE: CRUCIBLE
      VERSION: 1.7.10
      CRUCIBLE_RELEASE: staging-8ae5051
      GENERIC_PACKS: GT_New_Horizons_2.7.2_Server_Java_17-21
      GENERIC_PACKS_SUFFIX: .zip
      GENERIC_PACKS_PREFIX: https://downloads.gtnewhorizons.com/ServerPacks/
      DISABLE_MODS: |
        lwjgl3ify-2.1.5.jar
        archaicfix-0.7.4.jar
      ENABLE_QUERY: true
      EULA: TRUE
      MEMORY: 8G
      JVM_OPTS: -Dfml.readTimeout=180 @crucible-gtnh-java9args.txt
    ports:
      - '25565:25565'
    volumes:
      - ./gtnh/server:/data
      - ./gtnh/crucible-gtnh-java9args.txt:/data/crucible-gtnh-java9args.txt
      - ./gtnh/mods.txt:/data/mods.txt

### DIFF
--- a/docs/mods-and-plugins/index.md
+++ b/docs/mods-and-plugins/index.md
@@ -78,6 +78,20 @@ If applying large generic packs, the update can be time-consuming. To skip the u
 
 The most time-consuming portion of the generic pack update is generating and comparing the SHA1 checksum. To skip the checksum generation, set `SKIP_GENERIC_PACK_CHECKSUM` to "true".
 
+To disable specific mods, which can be useful for conflicts between multiple generic packs, you can use the `GENERIC_PACKS_DISABLE_MODS` variable to specify mods to disable.
+
+Disabling mods with docker run:
+```shell
+docker run -d -e GENERIC_PACKS_DISABLE_MODS=mod1.jar,mod2.jar ...
+```
+
+Disabling mods within docker compose files:
+```yaml
+      GENERIC_PACKS_DISABLE_MODS: |
+        mod1.jar
+        mod2.jar
+```
+
 ## Mods/plugins list
 
 You may also download or copy over individual mods/plugins using the `MODS` or `PLUGINS` environment variables. Both are a comma or newline delimited list of
--- a/docs/mods-and-plugins/index.md
+++ b/docs/mods-and-plugins/index.md
@@ -82,7 +82,7 @@ To disable specific mods, which can be useful for conflicts between multiple gen
 
 Disabling mods with docker run:
 ```shell
-docker run -d -e GENERIC_PACKS_DISABLE_MODS=mod1.jar,mod2.jar ...
+docker run -d -e GENERIC_PACKS_DISABLE_MODS="mod1.jar mod2.jar" ...
 ```
 
 Disabling mods within docker compose files:
--- a/scripts/start-setupModpack
+++ b/scripts/start-setupModpack
@@ -6,6 +6,7 @@ set -e -o pipefail
 : "${MODS:=}"
 : "${MODS_OUT_DIR:=/data/mods}"
 : "${MODS_FILE:=}"
+: "${DISABLED_MODS:=}"
 : "${PLUGINS:=}"
 : "${PLUGINS_OUT_DIR:=/data/plugins}"
 : "${PLUGINS_FILE:=}"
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: scripts/start-setupModpack [hunk 1]
```diff
@@ -192,6 +194,12 @@ function handleGenericPacks() {
         extract "${pack}" "${base_dir}"
       done
 
+      # Disable mods
+      for mod in ${DISABLE_MODS}; do
+	log Disabling $mod
+	find "${base_dir}" -name "$mod" -exec mv {} {}.DISABLED -v \;
```
itzg, line 200: Please indent this within the for-block

itzg, line 200: Let's use a suffix of ".disabled", lowercase to be consistent with other launcher/solutions.

scripts/start-setupModpack
```diff
@@ -155,6 +156,7 @@ function handleGenericPacks() {
   : "${GENERIC_PACKS:=${GENERIC_PACK}}"
   : "${GENERIC_PACKS_PREFIX:=}"
   : "${GENERIC_PACKS_SUFFIX:=}"
+  : "${DISABLE_MODS:=${DISABLE_MODS}}"
```
itzg, line 159: While the concept is applicable broadly, this processing is only done for generic packs, so it will be good to rename the variable with a `GENERIC_PACKS_` prefix to scope it. So, something like `GENERIC_PACKS_DISABLE_MODS`.

itzg (commented):
Thanks. Now I see what you meant with the comment https://github.com/itzg/docker-minecraft-server/issues/3144#issuecomment-2614529777

Please also add a brief description of the variable to the docs at https://github.com/itzg/docker-minecraft-server/blob/master/docs/mods-and-plugins/index.md?plain=1#L61

scripts/start-setupModpack
itzg: Also, I think you mistyped the init logic here -- you currently have it defaulting to itself. It should have just been:

```
 : "${DISABLE_MODS:=}
```

azunaVT: 100% a copy-past error for the init logic. I've renamed it to `GENERIC_PACKS_DISABLE_MODS`, I was actually thinking the same while I was renaming from `DISABLED` to `DISABLE`.

scripts/start-setupModpack
azunaVT: 100%, that's my vim editor not fully setup on this machine.

scripts/start-setupModpack
```diff
@@ -6,6 +6,7 @@ set -e -o pipefail
 : "${MODS:=}"
 : "${MODS_OUT_DIR:=/data/mods}"
 : "${MODS_FILE:=}"
+: "${GENERIC_PACKS_DISABLE_MODS:=}"
```
itzg, line 9: Seems to be an extra declaration to line 159

docs/mods-and-plugins/index.md
```diff
@@ -78,6 +78,20 @@ If applying large generic packs, the update can be time-consuming. To skip the u
 
 The most time-consuming portion of the generic pack update is generating and comparing the SHA1 checksum. To skip the checksum generation, set `SKIP_GENERIC_PACK_CHECKSUM` to "true".
 
+To disable specific mods, which can be useful for conflicts between multiple generic packs, you can use the `GENERIC_PACKS_DISABLE_MODS` variable to specify mods to disable.
+
+Disabling mods with docker run:
+```shell
+docker run -d -e GENERIC_PACKS_DISABLE_MODS=mod1.jar,mod2.jar ...
```
itzg, line 85: Does the comma delimiter work with the for-loop implementation?

azunaVT: Good catch, this needs to be space separated actually.

scripts/start-setupModpack
itzg: Sorry, I should have clarified to remove this one. The one on line 159 was correct since it was with the other variables for generic packs.

itzg (commented):
Super close 😀

scripts/start-setupModpack
azunaVT: Ah my bad, I can revert that!

itzg (approved):
Thanks!

PR #3289, opened 2025-02-01
Remove deprecated flag from mc-image-helper command in start-setupRbac

augustabt:
The `--append-only` flag was removed from `mc-image-helper manage-users` in commit [5fbfd0e](https://github.com/itzg/mc-image-helper/commit/5fbfd0e31b50f8eab9b115f9681e7a5bd5807b09#diff-ebfa578788a4287084a9d0b5b0c25783b3dea399fd06bbaa9d1f9af88119019cL62). If `APPEND_WHITELIST` or `OVERRIDE_WHITELIST` are set to false the container fails to start. As far as I can tell it is safe to simply remove the section since if neither variable is set to true the script defaults to `--existing=MERGE` which is the desired behavior.

### DIFF
--- a/scripts/start-setupRbac
+++ b/scripts/start-setupRbac
@@ -77,12 +77,17 @@ if [[ -v WHITELIST_FILE ]]; then
 fi
 if [[ -v WHITELIST ]]; then
   args=()
-  if isTrue "${APPEND_WHITELIST:-false}" || isFalse "${OVERRIDE_WHITELIST:-true}"; then
-    args+=(--append-only)
-  fi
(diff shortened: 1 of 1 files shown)


itzg (commented):
Thanks for working on a fix for this. I think accidentally forgot to do the same as APPEND_OPS and OVERRIDE_OPS here:

https://github.com/itzg/docker-minecraft-server/blob/26d97c21cc08571940d4016e75b479546fd96124/scripts/start-setupRbac#L50-L57

Can you try out a change similar to that?

The removal is good too since those are long now legacy variables.

itzg (approved):
Thanks!

PR #3322, opened 2025-02-16
Use /data/tmp for WORLD zip preparation

itzg:


scripts/start-setupWorld
```diff
@@ -26,24 +26,25 @@ if [[ "$WORLD" ]] && ( isTrue "${FORCE_WORLD_COPY}" || [ ! -d "$worldDest" ] );
 
   if isURL "$WORLD"; then
     log "Downloading world from $WORLD"
-    if ! get -o /tmp/world.bin "$WORLD"; then
+    mkdir -p /data/tmp
+    if ! get -o /data/tmp/world.bin "$WORLD"; then
```
MaxLevs, line 30: It feels like it will be more convenient to have a single variable for path to world.bin

We can use something like 
```sh
RAW_WORLD_DATA_PATH="/data/tmp/world.bin"
```
and then
```sh
mkdir -p "$(basedir "$RAW_WORLD_DATA_PATH")"
if ! get -o "$RAW_WORLD_DATA_PATH" "$WORLD"; then
# and so on
```

MaxLevs (commented):
👍

Jansvitil (commented):
Pull req 1

scripts/start-setupWorld
itzg: Good point

PR #3336, opened 2025-02-24
1863: use SLP for pre-1.7 servers.

jackguy80:
Tested locally on versions 1.6 and 1.4. No success on <1.4, but that may be a Java version issue


itzg (approved):
Thanks!

PR #3344, opened 2025-02-27
Update auto-curseforge docs wording around version pinning.

gregops:


itzg (approved):
Thanks, yeah I like that wording change. I see now that I leaned too far into CurseForge's vague "file" terminology -- an unfortunate side effect of coding against their API and then documenting it 😀

PR #3347, opened 2025-03-03
Fix log4j2 warning / Remove package scanning

litetex:
This should suppress this warning from showing up during boot:
```
The use of package scanning to locate Log4j plugins is deprecated.
Please remove the `packages` attribute from your configuration file.
See https://logging.apache.org/log4j/2.x/faq.html#package-scanning for details.
```


itzg (approved):
Thanks!

You might have already noticed, but this copy of the file is only used when ENABLE_ROLLING_LOGS is true.

PR #3350, opened 2025-03-06
add sound to excluded mods as it crashes server on load.

EmilyxFox:
This mod confusingly named [Sound​s](https://www.curseforge.com/minecraft/mc-mods/sound) with the slug `sound` is just incorrectly marked and crashes the server on load.


itzg (approved):
Thanks for following up on that.

PR #3360, opened 2025-03-14
Lazymc - Update misc examples with non-deprecated method of managing lazymc

joesturge:
As per the deprecation warning here: https://github.com/joesturge/lazymc-docker-proxy?tab=readme-ov-file#deprecated

This method is about to be removed from lazymc-docker-proxy

### DIFF
--- a/docs/misc/examples.md
+++ b/docs/misc/examples.md
@@ -31,15 +31,33 @@ services:
 With [lazymc-docker-proxy](https://github.com/joesturge/lazymc-docker-proxy) you are able to use [lazymc](https://github.com/timvisee/lazymc) with the minecraft container.
 
 ```yaml
+# Lazymc requires that the minecraft server have a static IP.
+#
+# To ensure that our servers have a static IP we need to create
+# a network for our services to use.
+#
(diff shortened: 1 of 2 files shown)


docs/misc/examples.md
```diff
@@ -31,15 +31,33 @@ services:
 With [lazymc-docker-proxy](https://github.com/joesturge/lazymc-docker-proxy) you are able to use [lazymc](https://github.com/timvisee/lazymc) with the minecraft container.
 
 ```yaml
+# Lazymc requires that the minecraft server have a static IP.
```
itzg, line 34: The way I read their instructions:

> *SERVER_ADDRESS - The address of the Docker Minecraft server to manage, should use the internal Docker network address of the server container, such as mc:25565 or the assigned static IP such as 172.18.0.3:25565.

It sounds like it still allows for docker assigned container IP addresses in which case you use the named form like "mc:25565".

itzg: That makes sense but your docs at lazymc didn't make that readily apparent. Since it's your software I'll trust the example is right and appropriately non-confusing for users 😀

itzg (approved):

PR #3361, opened 2025-03-15
Exclude compass-coords globally from cf and mr, client side mod

harkin:
Excludes this mod https://www.curseforge.com/minecraft/mc-mods/compass-coords (and https://modrinth.com/mod/compass-coords on MR)

Per the mod description "Needed on Client-side only, can't be installed on the server." and sure enough it crashed the server when installed. Came across this while trying to set up a the DawnCraft modpack

### DIFF
--- a/files/cf-exclude-include.json
+++ b/files/cf-exclude-include.json
@@ -24,6 +24,7 @@
     "cherished-worlds",
     "chunk-animator",
     "clickable-advancements",
+    "compass-coords",
     "configured",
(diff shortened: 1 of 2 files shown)


itzg (approved):
Thanks! And thanks for updating both global files since I have a bad habit of only updating one or the other.

itzg:
FYI, verify-build failure is unrelated. Modrinth servers seem to be down right now.

PR #3388, opened 2025-04-02
Add support of CF_API_KEY_FILE environment variable

MaxLevs:
If you try to pass the `CF_API_KEY` with file using `_FILE` suffix, it return an error about it cannot find the key.
This PR is about setting `CF_API_KEY` with `CF_API_KEY_FILE`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ RUN easy-add --var os=${TARGETOS} --var arch=${TARGETARCH}${TARGETVARIANT} \
   --var version=${MC_SERVER_RUNNER_VERSION} --var app=mc-server-runner --file {{.app}} \
   --from ${GITHUB_BASEURL}/itzg/{{.app}}/releases/download/{{.version}}/{{.app}}_{{.version}}_{{.os}}_{{.arch}}.tar.gz
 
-ARG MC_HELPER_VERSION=1.41.5
+ARG MC_HELPER_VERSION=1.41.6
 ARG MC_HELPER_BASE_URL=${GITHUB_BASEURL}/itzg/mc-image-helper/releases/download/${MC_HELPER_VERSION}
 # used for cache busting local copy of mc-image-helper
 ARG MC_HELPER_REV=1


itzg (commented):
Thanks for adding this. 

Can you also add the option in the documentation:

https://github.com/itzg/docker-minecraft-server/blob/master/docs/types-and-platforms/mod-platforms/auto-curseforge.md

itzg:
Also would be great to add the same for CURSEFORGE_FILES support

https://github.com/itzg/docker-minecraft-server/blob/dedf8bc587f451c66a724ed1f878e81cad66b98e/scripts/start-setupModpack#L296

documented in

https://github.com/itzg/docker-minecraft-server/blob/master/docs/mods-and-plugins/curseforge-files.md

MaxLevs:
Okey. I started making changes in master. That's not very good. I will change branch, make some changes and open new PR

itzg:
Being git, you can make changes to your own master branch all you want. But yes, pushing a branch is slightly smoother process.

MaxLevs:
Is there a convinient way to test it locally?

itzg:
https://docker-minecraft-server.readthedocs.io/en/latest/misc/contributing/building/

itzg:
...and now you're making me nervous. You submitted a PR without testing it?

MaxLevs:
Unfortunately, I could only test it undirectly (with a mock script). I have troubles with building an image locally and that's the reason I asked you what u do to test it.

I do not intend merging it before I can test locally. A PR is also a way to get a feedback about something I missed.

itzg:
> I have troubles with building an image locally 

I'm curious what exactly. If you already have docker then you should be able to build an image like

```
docker build -t itzg/minecraft-server .
```

MaxLevs:
> I'm curious what exactly.

image

`/bin/sh^M` looks like a line-endings issue. But have no idea of a reason. I have a windows machine and git configured to handle line enditngs. And I ran unix2dos to fix any possible file but nothing changes.

itzg:
You have git configured the wrong direction then. You need it to keep line endings as is so that they are correct when running within Linux container.

PR #3389, opened 2025-04-02
fixed logic to not compare the VERSION number when VERSION is set to …

jairgs:
…LATEST (default) and skip legacy flag in autopause

### DIFF
--- a/files/auto/autopause-fcns.sh
+++ b/files/auto/autopause-fcns.sh
@@ -25,10 +25,16 @@ use_proxy() {
 }
 
 use_server_list_ping() {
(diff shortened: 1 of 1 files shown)


files/auto/autopause-fcns.sh
```diff
@@ -25,11 +25,17 @@ use_proxy() {
 }
 
 use_server_list_ping() {
+  if [[ "${VERSION^^}" == "LATEST" ]]; then
```
itzg, line 28: For completeness, VERSION can also be set to "SNAPSHOT".

itzg (commented):
Thanks, just a small tweak needed

itzg (approved):
Thanks!

PR #3400, opened 2025-04-05
build: add image info properties

itzg:


itzg:
Note to future self: it would be nice to have the startup (optionally or opt-out-able) query github to tell the user how many days out of date the image's revision is from the repo's most recent git revision.

PR #3404, opened 2025-04-06
add resources key to docker compose jvm options

T0biii:
resources key was missing in example:

https://docs.docker.com/reference/compose-file/deploy/#resources

```yaml
deploy:
      resources:
        limits:
          cpus: '0.50'
          memory: 50M
          pids: 1
        reservations:
          cpus: '0.25'
          memory: 20M`
```


itzg (approved):
Thanks!

PR #3414, opened 2025-04-15
build: Add additional Java variants

lzlrd:


itzg (commented):
I'm sorry, I can't accept this PR. I am very methodical about how many variants are in use at any time and when they're added. Too many result in things like API rate limit overruns. 

You're welcome to build and push all of these variants to your own Docker Hub space though. 

If you can identify one single version that you'd like added and what variant to take away, then I'll consider it.

lzlrd:
Would you be willing to add `java24[-alpine]` (preferably Alpine, but `java24` would do) and java24-graalvm, in that case? Looking to utilise [`-XX:+UseCompactObjectHeaders`](https://openjdk.org/jeps/450) which released with Java 24.

itzg:
Yes but I'll need to remove the java23 variants to keep the list bounded.

lzlrd:
Not an issue for me, hopefully not for anyone else either.

Issue #3436, opened 2025-05-05
Impossible to ignore mods


enzofrnt:
### Describe the problem

Hi,

as you know, my server is crashing due to a client mod `particular` hen I tried to ignore it but server continue crashing after `docker compose down` and `docker compose up -d`.

### Container definition

```
services:
  minecraft:
    image: itzg/minecraft-server
    container_name: mc
    ports:
      - "25565:25565"
      - "24454:24454/udp"
      - "25575:25575"
    volumes:
      - ./data:/data
      - ./mods:/mods
    environment:
      TYPE: AUTO_CURSEFORGE
      EULA: TRUE
      CF_PAGE_URL: https://www.curseforge.com/minecraft/modpacks/holycube-revolution/files/6491613
      MEMORY: 20G
      SERVER_NAME: HolyCube Revolution - Enzo's Server
      RCON_PASSWORD: securepassword
      ENABLE_RCON: true
      MOTD: Les puants %MODPACK_NAME% verion %env:MODPACK_VERSION%
      CF_EXCLUDE_MODS: |
        particular
      CF_FORCE_SYNCHRONIZE: true
    restart: unless-stopped 
```

### Container logs

```bash
mc  | [16:20:38] [main/INFO] [MixinExtras|Service/]: Initializing MixinExtras via com.llamalad7.mixinextras.service.MixinExtrasServiceImpl(version=0.5.0-beta.4).
mc  | [16:20:38] [main/WARN] [mixin/]: Static binding violation: PRIVATE @Overwrite method decoratedId in modernfix-neoforge.mixins.json:perf.tag_id_caching.TagOrElementLocationMixin from mod modernfix cannot reduce visibiliy of PUBLIC target method, visibility will be upgraded.
mc  | [16:20:39] [main/INFO] [minecraft/Bootstrap]: ModernFix reached bootstrap stage (13.69 s after launch)
mc  | [16:20:41] [main/ERROR] [ne.ne.fm.co.as.RuntimeDistCleaner/DISTXFORM]: Attempted to load class net/minecraft/client/Minecraft for invalid dist DEDICATED_SERVER
mc  | [16:20:41] [main/WARN] [mixin/]: Error loading class: net/minecraft/client/Minecraft (java.lang.RuntimeException: Attempted to load class net/minecraft/client/Minecraft for invalid dist DEDICATED_SERVER)
mc  | Exception in thread "main" java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
[...]

PR #3437, opened 2025-05-05
Ignoring `particular`

enzofrnt:
Related to https://github.com/itzg/docker-minecraft-server/issues/3436

### DIFF
--- a/files/cf-exclude-include.json
+++ b/files/cf-exclude-include.json
@@ -91,6 +91,7 @@
     "oculus",
     "oldjavawarning",
     "overloaded-armor-bar",
+    "particular",
     "packmenu",
     "packmodemenu",
     "reauth",


itzg:
I'm not sure why you're submitting a PR when you suspect the exclusion doesn't work. The file is considered equally along with using the CF_EXCLUDE_MODS env var.  So if the var doesn't work the file update wont either.

enzofrnt:
Maybe my configuration is wrong. But the crash is caused by particular.

itzg:
Sure but why are you creating a PR for an issue you haven't resolved yet? Are you guessing/hoping this is the answer?

itzg (approved):
Regardless of my comments, I'm good to add this since https://www.curseforge.com/minecraft/mc-mods/particular does indeed look like it should be flagged as client only.

enzofrnt:
@itzg I just notice that : 
https://www.curseforge.com/minecraft/mc-mods/particular-reforged

I think that I'using the newer version, and maybe ignoring the older one ?

itzg:
Yes, that will be it then. The exclusion can only **exactly** match the identifier. Otherwise for all the code knows it is a completely different mod. 

You can submit a new PR to add that one too.

enzofrnt:
What's strange is that they said that after 1.20.1 the mod is require on server but with it my server is crashing

enzofrnt:
Also, if we check the server files provided with the modpack and see that this particular mod is missing, does that mean they gave false information on their page? That’s strange — I need to test this. Sorry for the confusion.

Sometimes I don’t take the time to do things thoroughly and end up rushing too much.

enzofrnt:
@itzg  So, particular-reforged is required on server before version 1.20.1. Then we cannot ignore particular-reforged everywhere.
But maybe, somewhere in the documentation, we can list by version of modpack the needed exclusion ?

itzg:
The best so far is excluding per modpack. Of course individual users are also able to add exclusions as needed to their particular container. Sounds like this is one that might have to fall into the last case, which is unfortunate.

itzg:
> Sorry for the confusion.
> 
> Sometimes I don’t take the time to do things thoroughly and end up rushing too much.

That's totally fine. I appreciate your enthusiasm and ongoing research on this.

PR #3443, opened 2025-05-12
Add wakes-reforged client mod to the cf excludes

JustSxm:


JustSxm:
After #3436 pointed me to the right direction to make "Holycube Revolution" (https://www.curseforge.com/minecraft/modpacks/holycube-revolution) work, I kept going and noticed that those two files needed to be ignored.

I also noticed smithing-template-viewer needed to be ignored, but it is already in the list ?

Is it because the latest docker image does not contain the the latest changes to that list ?

files/cf-exclude-include.json
itzg, line 1: Your editor seems to have changed every line of the file. Perhaps a whitespace change. Please revert and only make the line addition.

JustSxm: Im not sure why its not working now, I tried to do it without formatting, but its not working

Dragonvi1: What is the new entry. could append it into mine too tomorrow evening

Dragonvi1: I just did it so this can move forward without user having to figure out what his IDE did to the formatting.

PR #3450, opened 2025-05-16
Added forceInclude for particular-reforged since it's MC 1.20.1

Dragonvi1:
Allow particular-reforged in Beyond-Depth since it's a 1.20.1 modpack


itzg (approved):
Great solution

PR #3472, opened 2025-05-30
patch: conditionally add knockd to sudoers file

jasperchess:
If you are hosting this image for unknown users, sudo access to knockd can be used for privilege escalation _within_ the container.

There needs to be a lot of surrounding functionality for this to be a security risk, and for self-hosted minecraft servers this isn't a concern. I'm not 100% certain this is the correct approach, but I'm happy to make changes as necessary.

If you think this _is_ the correct approach I'm happy to update (with the help of [Peri123214](https://github.com/Peri123214)) the docs outlining the risk once this has been patched.

Big thanks to [Peri123214](https://github.com/Peri123214) for finding this.

### DIFF
--- a/files/auto/autopause-daemon.sh
+++ b/files/auto/autopause-daemon.sh
@@ -45,11 +45,7 @@ if isTrue "${DEBUG_AUTOPAUSE}"; then
   knockdArgs+=(-D)
 fi
 
-if isTrue "${SKIP_SUDO}"; then
-  /usr/local/sbin/knockd "${knockdArgs[@]}"
-else
-  sudo /usr/local/sbin/knockd "${knockdArgs[@]}"
-fi
+/usr/local/sbin/knockd "${knockdArgs[@]}"
 
 if [ $? -ne 0 ] ; then
   logAutopause "Failed to start knockd daemon."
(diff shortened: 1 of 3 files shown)


itzg:
...in fact, I think I'll merge this now and we can follow up with the other research outcome.

jasperchess:
> This approach looks fine. What I'm now wondering is with the setcap applied to the executable, maybe it doesn't even need to be sudo'ed
> 
> https://github.com/itzg/docker-minecraft-server/blob/67b0074a9e4f6318dc961881a15f714547dc77ef/build/ubuntu/install-packages.sh#L45
> 
> Can you research that a bit?

Ah good point! Let me look into that. If it is still required to sudo I'll fix the constant appending on every start - good catch!

jasperchess:
@itzg you were correct, with the capabilities set, it isn't necessary to sudo start knockd.

The testing I've done:
1. Start the server `-e ENABLE_AUTOPAUSE=TRUE -e AUTOPAUSE_TIMEOUT_INIT=30` (**without** adding `--add-cap` to the start command)
2. Wait for the autostop functionality to kick in
3. Try and connect
4. ✅ 

I haven't tested every distro (alpine/ubuntu/ol), let me know if more thorough testing is necessary

itzg (approved):
Thanks!

PR #3476, opened 2025-06-01
Complete fix of a privilege escalation exploit in the sudo configuration of Knockd (followup to  #3472)

peri123214:
As @jasperchess said in the original pull request, sudo access to knockd can be used for privilege escalation within the container. He has submitted a hotfix that was intended for the time until I patched the image in its entirety. My branch splits the runtime of the server to two users, one is for the Minecraft JAR (user `minecraft`) and the second one is for the tasks that use knockd to operate (namely Autopause and Autostop, username `service-account`). Service account has the same privileges as the minecraft account had before.

### DIFF
--- a/build/alpine/setup-user.sh
+++ b/build/alpine/setup-user.sh
@@ -3,4 +3,6 @@
 set -e
 
 addgroup -g 1000 minecraft
+addgroup -g 1001 service-group
 adduser -Ss /bin/false -u 1000 -G minecraft -h /home/minecraft minecraft
+adduser -Ss /bin/false -u 1001 -G service-group -H service-account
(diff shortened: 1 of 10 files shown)


itzg:
Did you see that sudo is no longer needed for knockd in #3472 ?

@jasperchess please pair up on this. 

I'm rapidly quite annoyed with knockd and will rip it all out if all this other stuff is needed to accommodate.

itzg:
BTW I didn't intend that PR as a hotfix...or did it break all of autopause?

itzg:
I'm removing sudoers file entirely in https://github.com/itzg/docker-minecraft-server/pull/3477

peri123214:
> BTW I didn't intend that PR as a hotfix...or did it break all of autopause?

The fix that jasperchess proposed would work, but only for servers with autopause/autostop disabled as enabling the feature would also enable the exploit. However the removal of sudoers is also a way of fixing it, so I will close this pull request. You can, however, reopen it if you have any further questions. Thanks for your time and effort in the entire manner!

PR #3478, opened 2025-06-01
Optimize Images

gothickitty93:
Optimized images using ImageOptim


itzg:
What's going on this with PR?

image

gothickitty93:
Wait a minute. it's actually correct. The image sizes are smaller. I misread the filesize as showing an increase, but comparing them with whats on my system, they match.

itzg:
But what's going on with the new top level directory?

image

...and why should I care about the documentation images being optimized? They'll get overwritten whenever I go to change diagrams, etc.

itzg:
For future reference please don't drop a mystery PR without explaining what you're doing, why you're doing it, and getting initial feedback.

PR #3479, opened 2025-06-01
Updated start-configuration

gothickitty93:
logError was missing new type: LEAF. Added.


itzg (approved):
Thanks, I often forget about that list.

Issue #2104, opened 2023-04-27
Add Banner as an option


drakosoldier298:
### Enhancement Type

New server type

### Describe the enhancement

Banner is a brand new fabric plugin based system made by mohist https://github.com/MohistMC/Banner
itd be nice to have an option for it

itzg:
I don't understand what it is. Is it a server type?

There's no releases

https://github.com/MohistMC/Banner/releases

And it's unstable and zero compatibility 

![image](https://user-images.githubusercontent.com/988985/234977328-e6a6137f-2902-4d18-9e1a-9ed75659fce0.png)

I'm going to park this until it's viable.

PR #3484, opened 2025-06-02
Add support for youer and banner

### LINKED ISSUE #3419: Further support of Youer
### Enhancement Type

New server type

### Describe the enhancement

Hi,

MohistMC is currently working on a new server named Youer. I believe it would be great and helpful to start looking into it early, especially because it supports NeoForge.

kingcavespider1:
Should fix #3419  but have not tested it much yet

### DIFF
--- a/docs/types-and-platforms/server-types/hybrids.md
+++ b/docs/types-and-platforms/server-types/hybrids.md
@@ -49,6 +49,34 @@ By default the latest build will be used; however, a specific build number can b
 
     -e VERSION=1.16.5 -e MOHIST_BUILD=374
 
+### Youer
+
+A [Mohist](https://github.com/MohistMC/Youer) server can be used with
+
+    -e TYPE=YOUER
+
+!!! note
+
+    There are limited base versions supported, so you will also need to  set `VERSION`, such as "1.12.2"
+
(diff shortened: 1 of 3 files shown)


kingcavespider1:
should also fix #2104

itzg (approved):
Thanks! Worked out nicely with their downloads API. Let me know when you've been able to fully test and ready for a merge.

kingcavespider1:
Works to download the server jars and correct mods for the platform from modrinth, also tested downloading mods from curseforge and plugins from spiget. Only issue I see is there doesn't seem to be a way to download mods and plugins from modrinth although I may be missing some way to do that, is there a thing like adding datapack: to download a datapack except for plugins?

kingcavespider1:
saw that you wanted to add a "fabric:" prefix for use with connector, would it make sense to have a "plugins:" prefix for hybrid servers, I think that would be through mc-image-helper though because I didn't see the code for that on the docker side

itzg:
> saw that you wanted to add a "fabric:" prefix for use with connector, would it make sense to have a "plugins:" prefix for hybrid servers, I think that would be through mc-image-helper though because I didn't see the code for that on the docker side

Yeah that makes sense to support and yes, that would be in mc-image-helper. I actually thought there was an issue already for it but was probably thinking the fabric: one.

PR #3508, opened 2025-06-22
Bump mc-image-helper version to 1.45.0

fireph:
bumping version to be able to use https://github.com/itzg/mc-image-helper/pull/580

docs PR to update alongside this is https://github.com/itzg/docker-minecraft-server/pull/3506


itzg:
Can you make this change as part of #3506? I prefer the docs and functional changes going together.

PR #3536, opened 2025-07-06
Add fastquit to cf-exclude-include.json

hights111:


itzg:
I'll finish this up in #3555

PR #3551, opened 2025-07-12
Use image helper's hc5 file download approach

itzg:
Pulls in https://github.com/itzg/mc-image-helper/pull/597


itzg:
https://github.com/itzg/mc-image-helper/pull/597#issuecomment-3065701802

PR #3552, opened 2025-07-13
Docs: fix env var for enabling query

gkeeth:
The docs referred to `QUERY_PORT` as the right env var for enabling query in the server properties, but as noted later in the docs that's for configuring the port for query (maps to the `query.port` server property), not for enabling/disabling query.

The correct env var to reference here is `ENABLE_QUERY`, I think. Ref https://github.com/itzg/docker-minecraft-server/blob/master/files/property-definitions.json#L12


itzg (approved):
Thanks, that's indeed what it should have been.

PR #3642, opened 2025-09-04
move USE_AIKAR_FLAGS=TRUE into else block

EmilyxFox:
fixes aikar flags always being enabled along-side meowice flags.. I think..
from [conversation in discord server](https://canary.discord.com/channels/660567679458869252/660567682751528981/1413234096989274203)


itzg (approved):
Thanks!

PR #3660, opened 2025-09-18
Should fix USE_MEOWICE_FLAGS

kingcavespider1:
At some point I need to go thru the updated meowice flags and check if anything should be updated with G1GC and add ZGC ones, but this should be enough to get the flags to work as a minimum

### DIFF
--- a/scripts/start-finalExec
+++ b/scripts/start-finalExec
@@ -166,7 +166,7 @@ if isTrue "${USE_MEOWICE_FLAGS}"; then
   fi
 fi
 
-if isTrue "${USE_AIKAR_FLAGS}"; then
+if isTrue "${USE_AIKAR_FLAGS}" || isTrue "${USE_MEOWICE_FLAGS}"; then
   # From https://mcflags.emc.gs/
   if isTrue "${USE_MEOWICE_FLAGS}"; then
     log "Using MeowIce's flags"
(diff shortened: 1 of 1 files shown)


kingcavespider1:
Guess nanolimbo is failing now

itzg:
> Guess nanolimbo is failing now

I just noticed that too with my latest PR. I have a fix queued up.

itzg:
FYI NanoLimbo fixed in https://github.com/itzg/docker-minecraft-server/pull/3662, so will need a rebase/merge to pick that up

kingcavespider1:
> FYI NanoLimbo fixed in #3662, so will need a rebase/merge to pick that up

Yeah I saw, just merged that, waiting to see if it worked correctly and passes tests

kingcavespider1:
Also, btw I tested graalvm jdk 25 on ol8 and ol9 and both seem to work with vanilla at least. Think 25 is a LTS release but I don't know what it adds over other versions, same goes for ol9.

itzg:
> Also, btw I tested graalvm jdk 25 on ol8 and ol9 and both seem to work with vanilla at least. Think 25 is a LTS release but I don't know what it adds over other versions, same goes for ol9.

I'm waiting for eclipse-temurin to release theirs, but good to know that graalvm is available.

Issue #3665, opened 2025-09-20
java_process_exists: command_not_found


iofq:
### Describe the problem

I'm seeing these logs being spammed on startup:

```
│ [init] Starting RCON commands                                                                                                        │
│ [init] Rcon cmds functionality enabled                                                                                               │
│ [init] Resolving type given FABRIC                                                                                                   │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 12: /auto/autopause-fcns.sh: No such file or directory
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
│ /usr/local/bin/rcon-cmds-daemon.sh: line 28: java_process_exists: command not found                                                  │
[...]

PR #3666, opened 2025-09-20
fix: update autopause-fcns.sh path

iofq:
Closes https://github.com/itzg/docker-minecraft-server/issues/3665

Found 2 other spots that I assume need to be updated.

There also may be a better way to update this path, not sure what you have in mind.


itzg (approved):
Thanks! I thought I had gotten them all but apparently not.

nickdavies:
Thanks for the quick fix!

PR #3711, opened 2025-10-15
add `management-server-allowed-origins` from 25w42a

EmilyxFox:
https://www.minecraft.net/en-us/article/minecraft-snapshot-25w42a


itzg (approved):
Thanks!

PR #3744, opened 2025-11-03
File globbing for Ignore Missing Files

JackHicks02:


JackHicks02:
^Fairly confident that failing check is a bug

itzg:
> ^Fairly confident that failing check is a bug

Seems to be, but it's strange I have never seen that before. I'll double check on my local system later.

itzg (approved):
Changes look great. I'll probably just merge even though RTD had that blip.

Issue #1515, opened 2022-05-11
Provide a clearer way to request JVM to derive memory limit percentages


itzg:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

As observed in #1512 , the unsetting of `MEMORY` with an empty string is an awkward and brittle way of saying "I would like the JVM to derive heap limits from the host/container". 

I would like to instead introduce the new variables:

- `MEMORY_PCT`
- `MAX_MEMORY_PCT`, defaults to `MEMORY_PCT`
- `MIN_MEMORY_PCT`, defaults to `MEMORY_PCT`

That will indicate `MEMORY` and co be ignored, but also add the JVM args, such as `-XX:MaxRAMPercentage=75`

shotah:
@itzg  This is easy to add in start-finalExec, but the real question is how do you want this to be triggered? Do we want to replace MEMORY with MEMORY_PCT? I can't imagine as it is probably widely in use. 🤔 Really this is more in how we'd want to handle memory_pct vs min and max. Thoughts

```
if [ ! "${MEMORY}" ]; then
  MIN_MEMORY_PCT=$([ "$MIN_MEMORY_PCT" == 5 ] && echo "$MIN_MEMORY_PCT" || echo "$MEMORY_PCT")
  MAX_MEMORY_PCT=$([ "$MAX_MEMORY_PCT" == 5 ] && echo "$MAX_MEMORY_PCT" || echo "$MEMORY_PCT")
  JVM_XX_OPTS="${JVM_XX_OPTS}
  -XX:MinRAMPercentage=${MIN_MEMORY_PCT}
  -XX:MaxRAMPercentage=${MAX_MEMORY_PCT}
  "
fi
```

itzg:
To reword what I meant in the description: if any of the "PCT" variables are set, then the MEMORY variables are ignored. Otherwise, MEMORY variables remain working as always. 

I was just expecting the usual fallback variable mechanism like 

```
: "${MIN_MEMORY_PCT:=${MEMORY_PCT:-25}}"
```

and then say 95 for the final fallback for max pct.

SgtMate:
Another idea I had on that topic would be to reuse the existing Variables and just check if the value has % added at the end. If % is used for a variable this would be treated accordingly.

itzg:
Great idea.

SgtMate:
[...]

PR #3747, opened 2025-11-05
Implement jvm mem percentage

SgtMate:
#1515 

Added functions for handling percentages in start-utils.
Went through all scripts srearching for 'mem' to find all places that handle memory related stuff, then I tried adding options for `-XX:InitialRAMPercentage` and `-XX:MaxRAMPercentage`. Where this was not possible (or i did not know any better) I calculated the absolue value based on available memory (not sure if this is a good option).

Happy for any feedback on this!

_Tested with spigot and GTNH type, dont have a curseforge account so this is still open._

Edit: [-XX:MinRAMPercentage](https://www.baeldung.com/java-jvm-parameters-rampercentage#-xxminrampercentage) was not used as i did not see any use for it.

### DIFF
--- a/docs/configuration/jvm-options.md
+++ b/docs/configuration/jvm-options.md
@@ -10,6 +10,12 @@ By default, the image declares an initial and maximum Java memory-heap limit of 
 
 The values of all three are passed directly to the JVM and support format/units as `<size>[g|G|m|M|k|K]`.
 
+To have control over heap size, without relying on absolute memory sizes percentages are also supported using `<size>%`.
+
+!!! info "RAMPercentage parameters"
+    Percentage based heap sizing is enabled using `-XX:InitialRAMPercentage` for `INIT_MEMORY` and `-XX:MaxRAMPercentage` for `MAX_MEMORY`.
+    For details on the function of these parameters look [here](https://www.baeldung.com/java-jvm-parameters-rampercentage).
+
 !!! example "Using docker run"
 
     ```
@@ -37,19 +43,7 @@ The values of all three are passed directly to the JVM and support format/units 
           MAX_MEMORY: 4G
     ```
 
-To let the JVM calculate the heap size from the container declared memory limit, unset `MEMORY` with an empty value, such as `-e MEMORY=""`. By default, the JVM will use 25% of the container memory limit as the heap limit; however, as an example the following would tell the JVM to use 75% of the container limit of 4GB of memory:
-
-!!! example "MaxRAMPercentage using compose file"
-
-    ```
-        environment:
-          MEMORY: ""
-          JVM_XX_OPTS: "-XX:MaxRAMPercentage=75"
-        deploy:
-          resources:
-            limits:
-              memory: 4G
-    ```
+To let the JVM calculate the heap size from the container declared memory limit, unset `MEMORY` with an empty value, such as `-e MEMORY=""`. By default, the JVM will use 25% of the container memory limit as the heap limit.
 
 !!! important
     The settings above only set the Java **heap** limits. Memory resource requests and limits on the overall container should also account for non-heap memory usage. An extra 25% is [a general best practice](https://dzone.com/articles/best-practices-java-memory-arguments-for-container).


scripts/start-deployBukkitSpigot
```diff
@@ -28,7 +28,19 @@ function buildSpigotFromSource {
   mkdir ${tempDir}
   cd ${tempDir}
 
-  jvmOpts="-Xms${INIT_MEMORY:-$MEMORY} -Xmx${MAX_MEMORY:-$MEMORY}"
+  jvmOpts=""
+
+  if isPercentage "${INIT_MEMORY:-$MEMORY}"; then
+    jvmOpts+="-XX:InitialRAMPercentage=$(getPercentageValue "${INIT_MEMORY:-$MEMORY}") "
```
itzg: I just happened to be in these docs a few minutes ago. These need to ultimately go in JVM_XX_OPTS. 

https://docker-minecraft-server.readthedocs.io/en/latest/configuration/jvm-options/#:~:text=If%20needing%20to%20set%20%2DXX%20options%2C%20then%20pass%20those%20with%20the%20JVM_XX_OPTS%20variable%20instead

I haven't looked yet to see where this local var ends up.

scripts/start-utils
```diff
@@ -39,6 +39,23 @@ function get_major_version() {
...
+function isPercentage() {
+  local value=$1
+  
+  [[ $value =~ ^[0-9]+(\.[0-9]+)?\s*%$ ]]
+}
+
+function getPercentageValue() {
+  local value=$1
+  if isPercentage "$value"; then
+    echo "$value" | sed -E 's/([0-9]+(\.[0-9]+)?)\s*%/\1/'
```
itzg: I always have to lookup what bash can do, but I wonder if it has a way to pick out a matching group? I'm a fan of sed, but always nice to have a bash-native approach when possible.

scripts/start-utils
```diff
@@ -239,10 +257,16 @@ function normalizeMemSize() {
   *g)
     scale=1073741824
     ;;
+  *%)
+    # Get system memory
+    mem=$(free -b | awk '/^Mem:/{print $2}')
```
itzg: Nice technique. I'm wondering if that field consistently represents the container's memory limit when configured

https://docs.docker.com/reference/compose-file/services/#mem_limit

itzg (commented):
Thanks. Great overall! Just some questions/thoughts.

scripts/start-deployBukkitSpigot
```diff
@@ -28,7 +28,19 @@ function buildSpigotFromSource {
...
+  fi
+
+  if isPercentage "${MAX_MEMORY:-$MEMORY}"; then
+    jvmOpts+="-XX:MaxRAMPercentage=$(getPercentageValue "${MAX_MEMORY:-$MEMORY}")"
+  else
+    jvmOpts+="-Xmx${MAX_MEMORY:-$MEMORY}"
+  fi
 
   logn ''
   curl -sSL -o ${tempDir}/BuildTools.jar https://hub.spigotmc.org/jenkins/job/BuildTools/lastSuccessfulBuild/artifact/target/BuildTools.jar && \
```
SgtMate: @itzg here its used for building it seems. i dont see a good solution for moving this.

scripts/start-utils
SgtMate: Okey I asked ChatGPT for some input regarding a bash native variant. Ill check if the AI mastermind actually gave me a good hint or if its just hallucinating.

scripts/start-utils
SgtMate: ok it does not. seems like even though its limited via cgroup, free gets the full system memory (might just get the same output as /proc/meminfo). instead you gotta check the cgroups limit in /sys/fs.

SgtMate:
think thats it. not sure what to do with `start-deployBukkitSpigot` this is pretty much a special case.

scripts/start-deployBukkitSpigot
itzg: Agreed, this doesn't need the same memory limits applied. Just let the JVM do its default allocation limit.

itzg:
Thanks. I'll re-review it tomorrow.

scripts/start-deployBukkitSpigot
SgtMate: ok ill remove that then.

Issue #2409, opened 2023-10-05
if there is possible to add catserver version tag to support 1.16.5 and 1.18.2?


IwasawaMasami:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

accroding to cuurent [CatServer release](https://github.com/Luohuayu/CatServer/releases) now support 1.16.5 and 1.18.2.

itzg:
Interesting. So their release names convey the applicable Minecraft version. Looks like Catserver support will need to be completely re-written then. 

Contributions are welcome.

Issue #2204, opened 2023-06-12
Support for an additional java version (java11-graalvm-ce)


PalmierPale:
### Enhancement Type

New server type

### Describe the enhancement

Hello,
I would like to know if it was possible to add a java version in tag for the itzg/docker-minecraft-server image.
I use a Minecraft server software (CatServer) in 1.16.5 and I would have to use a version of java 11 because java 8.
Of all the java versions on offer, I really prefer java graalvm-ce because after several tests I've carried out, it's the one that offers the best cpu / ram performance.
Unfortunately, itzg/docker-minecraft-server images with graalvm-ce only exist in java 8 (outdated) and java 17 (too recent).
Would it be possible to add the itzg/docker-minecraft-server:java11-graalvm-ce image?

Regards

itzg:
The graalvm variants tend to use the most build time, as seen here:

https://github.com/itzg/docker-minecraft-server/actions/runs/5243668852/usage

So I am hesitant to another if not taking away java 8 or 17. 

Can you provide your testing methodology and analysis where you determined the graalvm offered the best performance?

PalmierPale:
I don't really have any precise statistics to provide, the tests I did were more to validate what I read about graalvm.

I can give you the sources I used to research the subject:
- https://github.com/brucethemoose/Minecraft-Performance-Flags-Benchmarks
(and I've just realized that all reddit threads have been made private for some time)

If you want, I can redo the tests with more precision and a more defined methodology to get more reliable data on the performance advantages between native java, openj9 and graalvm.

But for my first observations, if openj9 is slightly better on ram consumption, it's bad on cpu performance, having only 2 players connected and immobile with a hundred mods and a blank server, the duration of a tick is 14ms on openj9 (and I'm not talking about chunk generation which is very slow) when graalvm arrives at 8ms.
After all, as I said above, these are approximate figures based on tests that aren't really very reliable, so it's more my "feeling".

itzg:
Sorry if I'm sounding obstinate, but I need to do some due diligence before adding yet another variant to the long list I already provide...and this is the first time this particular variant has been requested out of millions of image downloads. 

That aside, the article mentions **Enterprise** Edition (EE) has the supposed optimization features.

> Only GraalVM Enterprise Edition comes with the full set of optimizations.

From https://github.com/brucethemoose/Minecraft-Performance-Flags-Benchmarks#graalvm-enterprise-edition

[...]

Issue #1914, opened 2023-01-18
Forge Version is always for MC 12.2 when using Catserver


AlfredQingyao:
### Describe the problem

I want to run a 1.16.5 catserver MC server.
I changed the MC server env to 1.16.5 but the forge Version doesn't change.
I've tried to add env 'FORGE_FORCE_REINSTALL' and 'FORGE_VERSION=36.2.35' to reinstall right Forge version, but useless.

### Container definition

```
Paste run command or compose file here
```
![image](https://user-images.githubusercontent.com/25794476/213180619-f4fd927b-1be7-4f7f-9165-9124c4ca6e06.png)

### Container logs

```
Paste logs here
```
[init] Running as uid=1000 gid=1000 with /data as 'drwxrwx--- 18 1000 100 4096 Jan 18 21:03 /data'
[init] Resolved version given 1.16.5 into 1.16.5 and major version 1.16

[init] Resolving type given Catserver
[init] Using Aikar's flags
[init] Setting initial memory to 4G and max to 4G
[init] Starting the Minecraft server...
Loading libraries, please wait...
2023-01-18 21:05:21,914 main WARN Disabling terminal, you're running in an unsupported environment.
[21:05:22] [main/INFO]: Loading tweak class name net.minecraftforge.fml.common.launcher.FMLServerTweaker
[21:05:22] [main/INFO]: Using primary tweak class name net.minecraftforge.fml.common.launcher.FMLServerTweaker
[21:05:22] [main/INFO]: Calling tweak class net.minecraftforge.fml.common.launcher.FMLServerTweaker
[21:05:22] [main/INFO]: Forge Mod Loader version 14.23.5.2860 for Minecraft 1.12.2 loading

[21:05:22] [main/INFO]: Java is OpenJDK 64-Bit Server VM, version 1.8.0_212, running on Linux:amd64:5.10.28-Unraid, installed at /usr/lib/jvm/java-1.8-openjdk/jre

itzg:
With brief research I see that the releases are mixed 1.12 and 1.16. 

https://github.com/Luohuayu/CatServer/releases

This will need to be an enhancement to handle those more specifically.

SgtMate:
Not sure but is this still a thing. Thought about working on that, but the github page seems down and the api calls in the code hits a wall of 404. 
I see 3 possible solutions: 1. Im stupid and need help opening a weblink. 2. This is dead and we should remove this `type` 3. we keep it as is and nothing happened here.

[...]

Issue #1034, opened 2021-09-09
Support for subsequent end of catserver


yzazzx2021:
Hello, author, thank you for your docker image. Recently, I am preparing to convert a better minecraft forge official service integration package into a module plug-in hybrid terminal. However, the version of its server integration package is 1.16.5, and the existing hybrid end is only supported by Mohist and loliserver. And your docker image only supports the Mohist end, but the compatibility of this end is not as good as loliserver in terms of plug-ins. Therefore, please take the trouble to support loliserver as soon as possible. Thank you.

itzg:
I am finding it very difficult to find any information on Loliserver, so automating the download and installation is impossible at this point. 

Please use the custom server option

https://github.com/itzg/docker-minecraft-server#running-with-a-custom-server-jar

PR #3751, opened 2025-11-06
Removed catserver and updated magma download

SgtMate:
removed catserver and updated magma download as it seems have changed.

_TBD: magma could also be removed as it seems to be in a state of limbo at the moment and is not documented in the wiki._

### DIFF
--- a/scripts/start-deployMagma
+++ b/scripts/start-deployMagma
@@ -9,7 +9,7 @@ isDebugging && set -x
 resolveVersion
 
 # Neo Magma currently supports just 1.21.x
-if [[ ! $VERSION =~ ^1\.21(\.[0-9]+)?$ ]]; then
+if [[ ! $VERSION = 1.21.* ]]; then
 logError "Magma does not support $VERSION (expected 1.21.x)"
 exit 1
 fi


SgtMate:
Catserver related issues
#1914 
#2409
#2204
#1034

scripts/start-deployMagma
```diff
@@ -8,18 +8,14 @@ isDebugging && set -x
...
-if ! downloadUrl=$(get --json-path '$.link' "https://api.magmafoundation.org/api/v2/${VERSION}/latest/${MAGMA_VERSION}"); then
-  logError "Failed to locate latest Magma download for ${VERSION}. Is that version supported?"
-  exit 1
-fi
-
-if [[ $downloadUrl == null ]]; then
-  logError "Magma does not seem to be available for $VERSION"
-  exit 1
+# Neo Magma currently supports just 1.21.x
+if [[ ! $VERSION =~ ^1\.21(\.[0-9]+)?$ ]]; then
```
itzg: Since bash has prefix matching, use this since regex is fun, but aways hard to maintain:


```suggestion
if [[ ! $VERSION = 1.21.* ]]; then
```

itzg (approved):

PR #3752, opened 2025-11-06
Add 'status-effect-bars-reforged' to global excludes

kman3107:
I was trying to start up FTB Stoneblock 4 and this mod `Status Effect Bars Reforged` crashed my server.

I am also having some datapack issues, but that is a separate problem, I'll look into that after making this PR.

EDIT: Here is the link for the mod in question, in case you want to check it out https://www.curseforge.com/minecraft/mc-mods/status-effect-bars-reforged


itzg:
Sorry for the delay merging this.

PR #3753, opened 2025-11-07
forceIncludes particular-reforged for ftb-stoneblock-4

kman3107:
[Particular Reforged](https://www.curseforge.com/minecraft/mc-mods/particular-reforged) must be present in server mods to load some datapack stuff.


itzg (approved):
Thanks!

PR #3762, opened 2025-11-15
tests: use service_started condition

itzg:

PR #3763, opened 2025-11-15
fix(tests): ensure Nginx is fully spun up before starting MC container

AlexTMjugador:
This change implements the behavior described in https://github.com/itzg/docker-minecraft-server/pull/3762#issuecomment-3536417072, which should help making test execution less racy.

### DIFF
--- a/tests/setuponlytests/generic-packs/docker-compose.yml
+++ b/tests/setuponlytests/generic-packs/docker-compose.yml
@@ -3,9 +3,15 @@ services:
     image: nginx
     volumes:
       - ./web:/usr/share/nginx/html
+    healthcheck:
+      test: ["CMD", "curl", "--fail", "http://localhost/configs.zip"]
+      interval: 3s
+      timeout: 5s
+      retries: 3
   mc:
     depends_on:
-      - web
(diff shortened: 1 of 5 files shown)


itzg (approved):
Thanks! That was much easier to understand via a PR.

itzg:
FYI, found a fix for the alpine issue. Need to exclude http3 support due to mismatch on alpine.

AlexTMjugador:
Awesome! I'll keep looking into these mysterious failures for an unrecognized archive type when downloading test files from the web server as well...

itzg:
Again, the only variable that changed was the Oracle Linux upgrade. 

All graalvm builds were fine up to this one

https://github.com/itzg/docker-minecraft-server/actions/runs/19312338293/job/55235490311

And after upgrade the next two have failed 

https://github.com/itzg/docker-minecraft-server/actions/runs/19382359591/job/55463367190

https://github.com/itzg/docker-minecraft-server/actions/runs/19390730903/job/55483795803

I could add graalvm as a variant to include in PR test, but I was avoiding that since it's the slowest one to build.

PR #3799, opened 2025-12-11
Do not overwrite existing NanoLimbo settings.yml

akippnn:
Ran into a likely edge case where any modified changes to the settings.yml is immediately overwritten by the `start-deployNanoLimbo` bash script every time the container starts. This change fixed it for me.

### DIFF
--- a/files/nanolimbo-settings-patch.json
+++ b/files/nanolimbo-settings-patch.json
@@ -1,0 +1,12 @@
+{
+  "file": "/data/settings.yml",
+  "ops": [
+    {
+      "$set": {
+        "path": "$.bind.port",
+        "value": "${SERVER_PORT}",
+        "value-type": "int"
+      }
+    }
+  ]
+}
--- a/scripts/start-deployNanoLimbo
+++ b/scripts/start-deployNanoLimbo
@@ -160,22 +160,7 @@ traffic:
 EOF
 fi
 
-cat <<EOF > /tmp/settings-patch.json
-{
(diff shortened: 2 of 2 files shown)


itzg:
The file is being rewritten every time on purpose in order to ensure container variables are applied, such as SERVER_PORT. 

Can you describe your edge case?

akippnn:
> The file is being rewritten every time on purpose in order to ensure container variables are applied, such as SERVER_PORT.
> 
> Can you describe your edge case?

I modified NanoLimbo's configuration using the `/config` bind mount with `COPY_CONFIG_DEST=/data` to add environment variable substitutions (e.g. `infoForwarding.secret=${CFG_VELOCITY_SECRET}`), but this setup would not work because [start-setupMounts](https://github.com/itzg/docker-minecraft-server/blob/aadf1a8b9f91148b013e320bb5971b1141d39c28/scripts/start-setupMounts) intentionally skips newer files in destination.

The other workaround seems to be making patch definitions. I am mostly abusing `/config` for continuity (since the servers were architected this way since 1.19).

Since SERVER_PORT is the only variable modified, is using `mc-image-helper patch` to place `${SERVER_PORT}` in bind.port good in this case? This would still replace any manual edits for the bind.port but it would be more usable than replacing the entire file altogether.

itzg:
> Since SERVER_PORT is the only variable modified, is using mc-image-helper patch to place ${SERVER_PORT} in bind.port good in this case?

Yeah, that seems like a good solution.

PR #3805, opened 2025-12-14
Fix alignment of annotations on shell command on commands docs page

tibtiq:
#3477 modified the example showing how to send a command when RCON is disabled, but forgot to update the annotation explaining parts of the command so they're misaligned now.


itzg (approved):
Thanks!

PR #3814, opened 2025-12-19
Enable native access for quieter Java 25+ use of mc-image-helper

itzg:
**NOTE**

java8 builds are now pinned at 1.51.x versions of mc-image-helper.

Pulls in https://github.com/itzg/mc-image-helper/pull/680

### DIFF
--- a/.github/workflows/build-multiarch.yml
+++ b/.github/workflows/build-multiarch.yml
@@ -92,19 +92,19 @@ jobs:
             platforms: linux/amd64,linux/arm/v7,linux/arm64
             mcVersion: 1.12.2
             # Pin version for Java 8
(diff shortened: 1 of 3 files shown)


itzg:
It turns out this is only supporting starting with Java 17:

> The command-line option --enable-native-access=ALL-UNNAMED was introduced in Java 17. It is supported in subsequent Java versions as part of an ongoing effort to prepare for restricting the default use of the Java Native Interface (JNI) and the Foreign Function and Memory (FFM) API.

itzg:
I should use env var `$MC_IMAGE_HELPER_OPTS` for adjusting this for mc-image-helper based on java version.

PR #3815, opened 2025-12-19
Revert "Enable native access for quieter Java 25+ use of mc-image-helper"

itzg:
Reverts itzg/docker-minecraft-server#3814


itzg:
Fixed with #3819

PR #3826, opened 2025-12-22
docs: add autoscaling documentation and examples for mc-router integration

Lenart12:


itzg:
I'm so sorry. This PR fell off my radar. Looking now.

PR #3830, opened 2025-12-25
Fixed volume definition in docker-compose

MarcTroll:
I fixed the definition inside the documentation for the definition of the volumes. As far as I know, the volume section must be an enumeration and therefore requires the minus before each entry.


itzg (approved):
Thanks! Got myself confused for a moment, since it is correct further down on the page.

PR #3837, opened 2025-12-27
Decommission EXEC_DIRECTLY

tekikaito:
This removes support for the `EXEC_DIRECTLY` option, which previously allowed users to attach directly to the Minecraft server console.


itzg (approved):
Thanks!

Issue #3851, opened 2026-01-05
Switch from Material for MkDocs to Zensical


itzg:
Material for MkDocs is now in maintenance mode. 

The team is focusing on
https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/

Read the docs already supports it
https://docs.readthedocs.com/platform/stable/intro/doctools.html

EmilyxFox:
I'm taking a look at this 🫡

itzg:
Completed in #3852

PR #3852, opened 2026-01-05
Move docs to zensical

EmilyxFox:
See #3851

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.iml
 /gh-md-toc
 personal-build-and-develop.*
+site/


compose.zensical.yaml
itzg: I'm never quite sure the proper convention of alternate compose file names. So out of ignorance, with a dash, `compose-zensical.yaml` seems better to me. I also gravitate towards ".yml" but I know you prefer ".yaml" 😄 

Or maybe `compose-docs.yml` (or `compose-docs.yaml`)

EmilyxFox: Using `.yaml` is preferred and Compose will pick that over other names first if multiple exist, but that probably doesn't matter here. I thought using the `compose.something.yaml` and was convention as well, but I couldn't find anything about it in the docs, and under [Working with Multiple compose files](https://docs.docker.com/compose/how-tos/multiple-compose-files/) they have one with the name `common-service.yml`. I like the generic name tho I'll change it to that 🫡

EmilyxFox:
I'll add the `site/` dir to `.gitignore`. When using the normal zensical/zensical image it will emit the site even when in watch mode to `site/`, but when mounted like in the docs compose file it doesn't do that so it doesn't really matter

itzg (approved):
Perfect

itzg:
Ready for non-draft? Sorry, I should have asked earlier.

EmilyxFox:
> Ready for non-draft? Sorry, I should have asked earlier.

Yep! If you like the site dir and compose name, I think its ready

PR #3908, opened 2026-02-04
feat: add support for vanilla-like experience in PaperMC server configuration

Lenart12:
This PR adds an optional enviroment flag `PAPER_VANILLA_LIKE: true` which will apply recommended settings for Vanilla-like experience on Paper servers based on this article: https://docs.papermc.io/paper/vanilla/

### DIFF
--- a/docs/types-and-platforms/server-types/paper.md
+++ b/docs/types-and-platforms/server-types/paper.md
@@ -139,7 +139,7 @@ If you have attached a host directory to the `/data` volume, then you can instal
 
 For more technical players or those seeking a gameplay experience closer to vanilla Minecraft, `PAPER_VANILLA_LIKE` environment variable is provided. When set to `true`, it automatically configures Paper to minimize Bukkit/Paper-specific optimizations and changes that differ from vanilla gameplay.
 
-This applies all recommended configuration changes across `server.properties` `paper-global.yml`, `paper-world-defaults.yml`, and `spigot.yml` based on [PaperMC's official vanilla-like experience guide](https://docs.papermc.io/paper/vanilla/).
+This applies all recommended configuration changes across `server.properties`, `paper-global.yml`, `paper-world-defaults.yml`, and `spigot.yml` based on [PaperMC's official vanilla-like experience guide](https://docs.papermc.io/paper/vanilla/).
 
 !!! example
 


Lenart12:
Seems like also 1.20.x config fails to patch. I only tested this working for 1.21.x. I'm not sure what would be the best way to handle all the different versions.

itzg:
Thanks for submitting this PR. Would be great to start with an issue next time so that the idea can be discussed first. 

> Seems like also 1.20.x config fails to patch. I only tested this working for 1.21.x. I'm not sure what would be the best way to handle all the different versions.

Yes, that's the big challenge with maintaining this image.


I haven't reviewed the changes, but I'm worried about how many files were touched. 7 files and over 400 lines added?

Perhaps this is something best left for the end user to configure themselves. And I don't want to have to forever maintain adjustments that Paper publishes.

Lenart12:
Actual implementation only touches 2 files. The rest is docs+tests and a file large file that contains all the patches. So in reality there is little extra code to maintain. 

If this does not get accepted I have also though about an easier patching method. Something that could be enviroment only...
maybe something like:
```yml
environment:
  PATCH_config__paper-global.yml: |
    $.some.path: foo
    $.some.other.path: 3
```
Then everything could be self contained inside docker compose.

itzg:
> Actual implementation only touches 2 files. The rest is docs+tests and a file large file that contains all the patches. So in reality there is little extra code to maintain.

Sorry I jumped to conclusions. I haven't had a chance to jump a computer to look closer. 

Regarding the `PATCH_config__paper-global.yml` suggestion...that seems cool but a bit cryptic. I understand the motivation since there's only so much that can be conveyed in a key-value environment variable. 

It seems like inline configs mounted into the container would be more robust and leverage the existing patch mechanism

https://docs.docker.com/reference/compose-file/configs/#example-3

itzg (commented):
Now that I actually looked at the changes, I'm fine with the overall idea. Please try to incorporate my suggestion though.

Lenart12:
The idea does not seem so cryptic to me. I know about inline configs and I just feel like the docker compose would grow a lot writing json patch schema by hand. I had it envisioned it like this because it would allow for a compact representation of the patch files. The format would be like:

Define an environment variable in the format PATCH_<path> with the slashes replaced with double underscores, e.g. `config/paper-global.yml` becomes `PATCH_config__paper-global.yml` (and x underscores (x > 2) mean x-1 underscores inside the path to keep full compatibility).

The contents of the variable would be simple line by line instructions to generate a 1:1 translation in to a patch file.

For $set:
`$.chunks.delay-chunk-unloads-by = 0s | int` (value type can be optional and defaults to auto)
For $put:
`$.entities.spawning:max-arrow-despawn-invulnerability = disabled`
For $add:
`$.unsupported-settings.oversized-item-component-sanitizer.dont_sanitize[] = minecraft:container`

PR #3920, opened 2026-02-10
Add 'neoculus' to cf-exclude-include.json

kman3107:
I started up a Minecolonies Dimensional Adventure server and it crashed trying to load [NeOculus](https://www.curseforge.com/minecraft/mc-mods/neoculus). I set `CF_EXCLUDE_MODS=neoculus` and the server started fine. I'm assuming this is a client mod.


itzg (approved):
Thanks and thanks for the verification.

Issue #3932, opened 2026-02-15
Deploy of GTNH failes "Server files not found"


TTF5:
### Describe the problem

The deploy script for type=GTNH fails because it cannot locate/download the server files.

It seems the acces rights to the Bucket link used to fetch the versions seems to have changed.
https://github.com/itzg/docker-minecraft-server/blob/67ec9e1b3cced7384d79339d574a66409b15b045/scripts/start-deployGTNH#L11

### Container definition

docker run -d -it -p 25565:25565 -e EULA=TRUE -e TYPE=GTNH -e DEBUG=TRUE itzg/minecraft-server

### Container logs

2026-02-15 13:37:18.404 | ++ mc-image-helper java-release
2026-02-15 13:37:18.905 | + ((  25 >= 17  ))
2026-02-15 13:37:18.905 | + MC_IMAGE_HELPER_OPTS+=' --enable-native-access=ALL-UNNAMED'
2026-02-15 13:37:18.905 | + export MC_IMAGE_HELPER_OPTS
2026-02-15 13:37:18.905 | + export HOME=/data
2026-02-15 13:37:18.905 | + HOME=/data
2026-02-15 13:37:18.906 | ++ id -u
2026-02-15 13:37:18.907 | ++ id -g
2026-02-15 13:37:18.908 | ++ ls -lnd /data
2026-02-15 13:37:18.909 | + log 'Running as uid=1000 gid=1000 with /data as '\''drwxr-x--- 2 1000 1000 4096 Feb 14 20:46 /data'\'''
2026-02-15 13:37:18.909 | + local oldState
2026-02-15 13:37:18.909 | ++ shopt -po xtrace
2026-02-15 13:37:18.909 | + oldState='set -o xtrace'
2026-02-15 13:37:18.909 | + shopt -u -o xtrace
2026-02-15 13:37:18.910 | [init] 2026-02-15 12:37:18+00:00 Running as uid=1000 gid=1000 with /data as 'drwxr-x--- 2 1000 1000 4096 Feb 14 20:46 /data'
2026-02-15 13:37:18.910 | ++ paste -d, -s /etc/image.properties
2026-02-15 13:37:18.911 | + log 'Image info: buildtime=2026-02-14T20:46:17.160Z,version=java25,revision=67ec9e1b3cced7384d79339d574a66409b15b045'
2026-02-15 13:37:18.911 | + local oldState
2026-02-15 13:37:18.911 | ++ shopt -po xtrace
[...]

PR #3937, opened 2026-02-15
Update start-deployGTNH to use hypeserv api since gtnh api is broken atm

Flexusma:
See https://github.com/itzg/docker-minecraft-server/issues/3932 for further information. Not tested, just doing this since no one else apparently did so.


itzg:
Thanks but was just about to push a change using the restify+jq approach.

itzg:
...pushed #3938

Flexusma:
Yeah no worries

PR #3981, opened 2026-03-20
add more examples for new minecraft version schema

EmilyxFox:
I had to query the API just to be sure of how to specify pre-release and release candidates, so I thought I'd just add them as examples to the docs.

Could maybe make sense to change the layout of this page completely and use a table instead or something, but you let me know on that :)


EmilyxFox:
tbh I think formatting this part of the page completely differently might be better. This is just another idea. Let me know what you think 🫡

PR #4022, opened 2026-04-16
fix deadlink for autostop example

pacsui:
Deadlink for autostop pointing to older revision ./examples/docker-compose-autostop.yml


itzg (approved):
Thanks

PR #4031, opened 2026-04-25
Remove Connector & Appleskin from modrinth exclusions

evergales:
- [Sinytra Connector](https://modrinth.com/mod/connector) is required on the server side to load fabric mods on neoforge that add content
- According to its [mod page](https://modrinth.com/mod/appleskin) appleskin is required on the server side to display saturation values accurately


itzg:
What modpack did you test with to confirm that the server doesn't crash with those included?

evergales:
It was my own, heres a zip with the mrpack and compose.yaml included
(with the exclusion list disabled)
[aloft-pack.zip](https://github.com/user-attachments/files/27098489/aloft-pack.zip)

I've ran neoforge servers with both of these mods without any issues
so I'm a bit confused about the crashing part.
My guess as to why Sinytra was excluded in c70f9c3fff3f39da42c891caaf61e6ece2a9666f was because of Sinytra connector loading some other fabric mod that caused a crash on that specific setup
*to be fair, Sinytra was way worse in compatability 2 years ago*

itzg:
Thank for the followup.

Panderine:
Why appleskin only for modrinth and not for curseforge aswell?

itzg:
> Why appleskin only for modrinth and not for curseforge aswell?

Because I forgot to ask for this also in CF exclusions. Half the time I forget one or the other 😄

Issue #1818, opened 2022-11-11
Infer type/version/loader from packwiz


lhvy:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

Currently using packwiz with Quilt for the latest release of Minecraft and the Quilt loader can cause problems because the container sees that the type is set to Quilt, and downloads whichever version of the loader it prefers, which in this case is actually 0.18.x beta, which is incompatible with mods in my packwiz pack. This could be fixed if the version, type and loader version env vars were simply replaced with values read directly from the packwiz pack.

itzg:
I am gradually working towards this type of capability with CurseForge modpacks. So, this enhancement makes sense for packwiz also. 

The unique challenge with the packwiz version is that it'll need to pre-download the definition and interpret just enough to get the versions.

itzg:
[As discussed in Discord](https://discord.com/channels/660567679458869252/660569641550217327/1104122137175920740) this track nicely with TYPE=AUTO_CURSEFORGE where it would be TYPE=PACKWIZ

WilsontheWolf:
Had a similar issue, where despite my pack having the minecraft version set, it downloaded the latest and then it couldn't boot due to mod incompatibilities.

jcotton42:
> The unique challenge with the packwiz version is that it'll need to pre-download the definition and interpret just enough to get the versions.

This seems like it shouldn't be *too* bad? _knocks on wood_

The `pack.toml` pointed to by the URL contains all the pertinent versions under the `[versions]` heading.

jcotton42:
One way that comes to mind is, if the TYPE is packwiz, then set the VERSION, TYPE, and other variables from the manifest, then let the image continue as normal.

e.g. a packwiz manifest with
```toml
[versions]
minecraft = "1.21.1"
neoforge = "21.1.169"
```

Would result in VERSION=1.12.1, TYPE=NEOFORGE, and NEOFORGE_VERSION=21.1.169. The rest of the startup scripts would then continue as before, none the wiser that their settings were coming from packwiz instead of the container env.

jcotton42:
I threw together a proof of concept supporting just NeoForge https://github.com/itzg/docker-minecraft-server/compare/master...jcotton42:docker-minecraft-server:versions-from-packwiz

Running `docker run -e TYPE=packwiz -e PACKWIZ_URL=https://github.com
[...]

PR #4053, opened 2026-05-10
Load env vars from file or archive at startup

### LINKED ISSUE #4037: Inside-out configuration from generic packs/packwiz
### Enhancement Type

Improve an existing feature

### Describe the enhancement

I'd like a mechanism where the container is configured based on the content of the packwiz/generic pack.

For example, an `.env` file present in the generic pack could be used to set up environment variables, including the server `TYPE` etc. or, for packwiz packs, the values in the `pack.toml` could be used to select the server type/versioning.

This would mean when we ship our packs as artifacts, they don't depend on the outer configuration where a version bump would normally mean a breaking change for the pack consumer.

ChipWolf:
Closes #4037

## Summary

Adds four new variables that let a generic pack ship its own container configuration inside-out, so a pack artifact can declare its own `TYPE`, `VERSION`, mod loader, etc. instead of relying on the consumer's outer env.

- `LOAD_ENV_FROM_GENERIC_PACK` — when `true`, each entry in `GENERIC_PACKS` (after `GENERIC_PACKS_PREFIX`/`SUFFIX` expansion) is probed for a top-level `.env` and sourced in pack order. Packs without a `.env` are skipped silently.
- `LOAD_ENV_FROM_FILE` — container path or URL of a shell-style env file.
- `LOAD_ENV_FROM_ARCHIVE` — container path or URL of a zip/tar archive containing an env file.
- `LOAD_ENV_FROM_ARCHIVE_ENTRY` — relative path of the env file inside the archive (default `.env`).

## Why

Per the issue, packs shipped as artifacts shouldn't break their consumers when the recommended `TYPE`/`VERSION` for the pack changes, the pack itself should be the source of truth. The `LOAD_ENV_FROM_GENERIC_PACK` flag is for the original ask (env file lives **inside** the generic pack); the `FROM_FILE`/`FROM_ARCHIVE` primitives cover the orthogonal case of an env source that isn't a generic pack (separate config bundle, packwiz zip used with a different TYPE, etc.).

## Implementation notes

- New helpers `loadEnvFromFile` / `loadEnvFromArchive` / `loadEnvFromGenericPack` in `scripts/start-utils`, reusing the existing `applyResultsFile` (`set -a; source; set +a`) pattern called out in the issue.
- Wired into `scripts/start-configuration` right after `cd /data` and before `DECLARED_TYPE` / `DECLARED_VERSION` are captured, so the loaded `TYPE`/`VERSION` flow through the existing dispatch.
- Load order: `GENERIC_PACK` → `FROM_FILE` → `FROM_ARCHIVE`; later wins, and all override `docker run -e` / compose `environment:`.
- `LOAD_ENV_FROM_GENERIC_PACK` downloads each pack URL into `/data/packs/` using the same `get --skip-up-to-date` call that `start-setupModpack` makes later, so the archive is not fetched twice.
- `extract()` now accepts optional trailing paths-within-archive arguments and forwards them to `unzip`/`tar` (per review feedback), so probing a 100 MB pack for a 200 B `.env` doesn't expand the whole thing.
- Download/scratch dirs live under `/data/.tmp` rather than `/tmp` (per review feedback), matching the existing convention in `start-setupModpack`.
- `EULA` is intentionally **not** settable this way — the EULA check runs earlier in `start-configuration`, before any env is loaded. Documented as such.
- A `!!! warning` admonition flags that the env file is sourced by bash and should only point at trusted sources.

## Tests

Three new setup-only tests under `tests/setuponlytests/`:

- `load-env-from-generic-pack/` — sets `MOTD: from-compose` in compose and `MOTD=from-generic-pack` inside the pack's `.env`; asserts `server.properties` has the pack's value AND that the pack's `config/dummy.yml` made it into `/data` (so the regular unpack step still ran on the same archive).
- `load-env-from-file/` — sets `MOTD: from-compose` in compose and `MOTD=from-env-file` in the loaded file; asserts the env-file value won.
- `load-env-from-archive/` — same idea using a zip with `.env` at its root.

## Out of scope

The packwiz `pack.toml` → TYPE/VERSION inference path is being tracked separately in #1818, per discussion on the issue.

## Test plan

- [ ] `tests/setuponlytests/load-env-from-generic-pack` PASSED verify
- [ ] `tests/setuponlytests/load-env-from-file` PASSED verify
- [ ] `tests/setuponlytests/load-env-from-archive` PASSED verify
- [ ] Existing `setuponly` and `fulltests` still PASS
- [ ] `bash -n scripts/start-configuration scripts/start-utils` clean


scripts/start-utils
```diff
@@ -26,6 +26,71 @@ function applyResultsFile() {
   set +a
 }
 
+function loadEnvFromFile() {
+  local source=${1?Missing required source argument}
+  local downloaded=
+
+  if isURL "$source"; then
+    downloaded=$(mktemp)
```
itzg: Sometimes people have issues with their /tmp within the container. I was going to suggest a subdir of /data, but looking through the code there's like four different spots. 

Could give mktemp a target directory of `/data/.tmp`

scripts/start-utils
```diff
@@ -26,6 +26,71 @@ function applyResultsFile() {
...
+    archive=$downloaded
+  elif [ -f "$source" ]; then
+    archive=$source
+  else
+    logError "Archive not found: $source"
+    return 1
+  fi
+
+  tmpdir=$(mktemp -d)
+  if ! extract "$archive" "$tmpdir"; then
```
itzg: It would be nice to have `extract` utility function accept just the filename to extract. It seems a shame to expand the whole thing just for one file.

itzg (commented):
Thanks for taking care of this. I obviously got myself distracted before starting on this. It ended up looking very clean. 

Just a couple of things to consider, but am good with merging either way.

ChipWolf:
Added `LOAD_ENV_FROM_GENERIC_PACK=true` in [565e2fb](https://github.com/itzg/docker-minecraft-server/pull/4053/commits/565e2fb) for where the `.env` lives **inside** the generic pack and the user doesn't have to repeat the URL.

Behaviour:
- Iterates each `GENERIC_PACKS` entry (after `PREFIX`/`SUFFIX` expansion) in order and probes for a top-level `.env`.
- Each found `.env` is sourced. Later packs override earlier, matching the `cp -R` layering of the regular unpack step, so the same precedence the user already gets on file content carries through to the env they declare.
- Packs without a `.env` are silently skipped, so it composes with packs that just ship content.
- URLs are downloaded into `/data/packs/` using the same `get --skip-up-to-date` call `start-setupModpack` makes later, so nothing is fetched twice. The `extract` `entry` arg from the previous review comment is what makes this cheap — only the `.env` is pulled out for the probe.

The `LOAD_ENV_FROM_FILE` / `LOAD_ENV_FROM_ARCHIVE` primitives stay in for the orthogonal case (env source that isn't a generic pack — a separate config bundle, a packwiz zip used with a different `TYPE`, etc.). Load order is `GENERIC_PACK` → `FROM_FILE` → `FROM_ARCHIVE`.

New test under [`tests/setuponlytests/load-env-from-generic-pack/`](https://github.com/itzg/docker-minecraft-server/tree/master/tests/setuponlytests/load-env-from-generic-pack) covers it end-to-end (both that the pack's `.env` wins over compose `MOTD` AND that the regular unpack still applies `config/dummy.yml` from the same archive).

itzg (approved):
Fantastic! Thanks for implementing this. It sounds like it'll help you out quite a bit.

PR #4079, opened 2026-05-23
Specify .mrpack extension for local mrpack files

roziscoding:
Clarified the requirement for the local mrpack file extension.

Spent quite a while trying to understand this one and project issues make clear others have too 😅


itzg (approved):
Thanks for clarifying that for others.

PR #4080, opened 2026-05-24
fix(exclude): exclude ImmersiveDamageIndicators from server mods

LinkIsGrim:
Mod is clientside only.

### DIFF
--- a/files/modrinth-exclude-include.json
+++ b/files/modrinth-exclude-include.json
@@ -56,7 +56,7 @@
     "gpumemleakfix",
     "Highlighter",
     "ImmediatelyFast",
-    "immersive-damage-indicators",
+    "immersivedamageindicators",
     "indium",
     "inventory-profiles-next",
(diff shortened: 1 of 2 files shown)


files/modrinth-exclude-include.json
```diff
@@ -56,6 +56,7 @@
     "gpumemleakfix",
     "Highlighter",
     "ImmediatelyFast",
+    "immersive-damage-indicators",
```
itzg: Looking at https://modrinth.com/project/tTKSJOdG, the filename doesn't have dashes.

itzg (commented):
Thanks. Just needs a tweak.

files/modrinth-exclude-include.json
LinkIsGrim:
```suggestion
    "immersivedamageindicators",
```

itzg (approved):

Issue #4084, opened 2026-05-30
double backslashing --env-file


lethedata:
### Describe the problem

`MOTD=Line one\nLine two` in the env-file is translating to `motd=Line one\\nLine two`.

### Container definition

Create env-file

```txt
# env-file
EULA=TRUE
MOTD=Line one\nLine two
```

```sh
docker run -it -p 25565:25565 --env-file ./env-file itzg/minecraft-server
```

### Container logs

Stopped after DUMP_SERVER_PROPERTIES.

```
++ mc-image-helper java-release
+ ((  25 >= 17  ))
+ MC_IMAGE_HELPER_OPTS+=' --enable-native-access=ALL-UNNAMED'
+ export MC_IMAGE_HELPER_OPTS
+ export HOME=/data
+ HOME=/data
++ id -u
++ id -g
++ ls -lnd /data
+ log 'Running as uid=1000 gid=1000 with /data as '\''drwxr-x---+ 1 1000 1000 362 May 29 23:43 /data'\'''
+ local oldState
++ shopt -po xtrace
+ oldState='set -o xtrace'
+ shopt -u -o xtrace
[init] 2026-05-30 00:31:15+00:00 Running as uid=1000 gid=1000 with /data as 'drwxr-x---+ 1 1000 1000 362 May 29 23:43 /data'
++ paste -d, -s /etc/image.properties
+ log 'Image info: buildtime=2026-05-27T12:19:03.980Z,version=java25,revision=544382d5d60276f846113991585e69bb4f7077a3'
+ local oldState
++ shopt -po xtrace
+ oldState='set -o xtrace'
+ shopt -u -o xtrace
[init] 2026-05-30 00:31:15+00:00 Image info: buildtime=2026-05-27T12:19:03.980Z,version=java25,revision=544382d5d60276f846113991585e69bb4f7077a3
+ '[' '!' -e /data/eula.txt ']'
+ isTrue false
+ case "${1,,}" in
+ return 1
+ isTrue true
+ case "${1,,}" in
+ return 0
+ set +x
+ isTrue false
+ case "${1,,}" in
+ return 1
+ proxyArgs=()
+ [[ -n '' ]]
+ proxyArgs=()
+ addToProxyArgs http.proxyHost ''
+ [[ -n '' ]]
+ addToProxyArgs http.proxyPort ''
+ [[ -n '' ]]
+ addToProxyArgs http.nonProxyHosts ''
+ [[ -n '' ]]
[...]

PR #4086, opened 2026-05-30
feat: docker env-file handling

lethedata:
This helps fix dockers `--env-files` from being double slashed

### DIFF
--- a/scripts/mc-motd-fix.sh
+++ b/scripts/mc-motd-fix.sh
@@ -1,0 +1,5 @@
+#!/bin/env bash
+
+export MOTD="${MOTD//\\n/$'\n'}"
+exec /image/scripts/start "$@"
+
--- a/scripts/mc-motd-fix.sh
+++ b/scripts/mc-motd-fix.sh
@@ -1,5 +1,0 @@
-#!/bin/env bash
-
-export MOTD="${MOTD//\\n/$'\n'}"
-exec /image/scripts/start "$@"
-
--- a/scripts/start-setupServerProperties
+++ b/scripts/start-setupServerProperties
@@ -30,7 +30,8 @@ function customizeServerProps {
     fi
   elif [[ "${DOCKER_RUN_ENV_FILE:-false,,}"  = "true" ]]; then
      if [[ -v MOTD ]]; then
-       MOTD="$(printf '%b' "$MOTD")"
+       # Replace literal '\n' with a real newline
+       MOTD="${MOTD//\\n/$'\n'}"
      fi
   fi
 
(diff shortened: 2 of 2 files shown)


lethedata:
This fully resolves https://github.com/itzg/docker-minecraft-server/issues/4084. Needs some more tests before full PR

itzg:
Nothing to fix. It was a documentation issue.

itzg:
I can't accept this change. I already went through several revisions of MOTD handling and I don't want to introduce a recession for every other user.

lethedata:
> Nothing to fix. It was a documentation issue.

The `--env-file` is not a documented. Docker does not support multilines in these files and when the single line is pulled in, `\` get doubled.

> I can't accept this change. I already went through several revisions of MOTD handling and I don't want to introduce a recession for every other user.

I can understand that, that's why this is a draft rn. And not ready until further validations are done.

lethedata:
@itzg I've gated it behind a `DOCKER_ENV_FILE` variable for the specific use case which will also prevent regressions and breaking for users.

itzg:
I would prefer to enhance this

https://github.com/itzg/mc-image-helper/blob/master/src/main/java/me/itzg/helpers/properties/SetPropertiesCommand.java

to replace literal `\n` with newline characters.

A parameter can be added to https://github.com/itzg/mc-image-helper/blob/master/src/main/java/me/itzg/helpers/properties/PropertyDefinition.java in order limit which values get that processing.

itzg:
> I would prefer to enhance this
 
Actually it’s not a preference, it’s a requirement. I can make the change if you’re not comfortable with Java.

lethedata:
I'll look into the enhancements there instead and close this one out.

On the current setup, it works and I can't seem to break it so if anyone wants a fork with it it's there.

Changed the env flag to `DOCKER_RUN_ENV_FILE` as this seems limited to just `docker run` as it doesn't show up with compose. Used some pretty messy MOTD to test like below and they seemed to work fine. Just uses bash parameter expansion to swap `\n` into a literal line break.

```txt
MOTD=\u00A74\u00A7lB\u00A7c\u00A7oR\u00A7e\u00A7nO\u00A7a\u00A7mK\u00A7b\u00A7kE\u00A7r \u00A79STRING\n\u00A7d\u00A7l➔ \u00A77Reset (\u00A7r) \u00A7eWorks \u00A7cPerfect!
```

(also sorry for the flood of pushes there. When was syncing the last bit of changes to the current imp I accidentally brought in junk files)

lethedata:
> > I would prefer to enhance this
> 
> Actually it’s not a preference, it’s a requirement. I can make the change if you’re not comfortable with Java.

I'll give it an shot. Not comfortable so it'll take me some time so if you've already got idea's and want to don;t wait on me

itzg:
I appreciate the initiative, but I really don't mind. It'll take me just a few minutes however I can't get to it until later today or tomorrow since I'm busy doing stuff around the house.

Issue #4098, opened 2026-06-09
Improve GTNH Version Selection


SgtMate:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

I think i can improve the parsing and modpack selection using jq and the "new" version.json

SgtMate:
I opended the issue as a reference point for a PR im working on.

SgtMate:
Solved by #4101

PR #4101, opened 2026-06-11
Improve get gtnh download path

SgtMate:
In reference to #4098 this PR aims to improve the download selection for the GTNH modpack. It mainly relies on `jq` features to remove needless iterating and cleanup old flaws. 

> Note: this also removes the modpack version from the `gtnh-upgrade-*` backup folders as this caused (in my opinion) unneeded coupling between  `getGTNHdownloadPath()` and `updateGTNH()`.


itzg (approved):
The new code also looks nice and clean. Thanks for updating this.

PR #4117, opened 2026-06-20
Exclude e4mc and e4mc-retro from modpacks

Reycko:
Both e4mc and e4mc-retro will crash the server if they are used. Some private modpacks may include them as well, this should avoid any issues with whoever accidentally includes them.

### DIFF
--- a/files/cf-exclude-include.json
+++ b/files/cf-exclude-include.json
@@ -49,6 +49,8 @@
     "dynamiclights-reforged",
     "easiervillagertrading",
     "effective-forge",
+	"e4mc",
+	"e4mc-retro",
     "embeddium",
     "embeddium-extension",
     "embeddium-extras",
--- a/files/modrinth-exclude-include.json
+++ b/files/modrinth-exclude-include.json
@@ -36,6 +36,8 @@
     "DisableCustomWorldsAdvice",
     "distraction_free_recipes",
     "drippyloadingscreen",
+	"e4mc",
+	"e4mc-retro",
     "eating-animation",
     "emi_trade",
     "emiffect",


itzg (commented):
Thanks. It would be great if you could add the equivalent slug IDs here

https://github.com/itzg/docker-minecraft-server/blob/master/files/cf-exclude-include.json

Reycko:
> Thanks. It would be great if you could add the equivalent slug IDs here
> 
> [`master`/files/cf-exclude-include.json](https://github.com/itzg/docker-minecraft-server/blob/master/files/cf-exclude-include.json)

Apologizes, I was gonna do it and completely forgot oops
I'll commit it in a second

Reycko:
Apologies for the force pushes, I'm never using GitHub's UI to edit files again..

This should be good though, I also based it on the latest renovate commit.

itzg:
I hadn't heard of e4mc before. Now that I researched it I'm confused because it seems like something that would only work server side. Or is it only client side for people to "open to LAN"?

Reycko:
> I hadn't heard of e4mc before. Now that I researched it I'm confused because it seems like something that would only work server side. Or is it only client side for people to "open to LAN"?

Sorry for the very long reply time
The mod modifies the Open to LAN button to use a reverse proxy and sends you a tunnel link others can use to join, using it server-side seems to cause a crash and is otherwise not recommended

I tested:
e4mc-retro 1.1.0 on Forge 1.12.2: crash
e4mc 6.2.0 on Fabric 1.20.1: crash? (first launch caused an exception on the Netty thread, second launch did not cause any errors, although e4mc warns against using e4mc on dedicated servers)

itzg:
No problem at all. Thanks for the follow up. 

I'm working on a vaguely similar capability and was curious if e4mc was going to be a "competitor" 😁

PR #4133, opened 2026-06-25
Add purpur support for ServerLibraryCleaner

harrisonablack:
Adds support for the ServerLibraryCleaner, implemented in itzg/mc-image-helper#748. Mimics implementation in #4046 by adding a variable, `$CLEAN_SERVER_LIBRARIES` which defaults to true.

Updated documentation to list support for `TYPE=PURPUR`

### DIFF
--- a/docs/configuration/misc-options.md
+++ b/docs/configuration/misc-options.md
@@ -26,7 +26,10 @@ docker run -d --pull=always \
 
 By default, supported server types remove stale server libraries during installation by setting `CLEAN_SERVER_LIBRARIES` to `true`. Set `CLEAN_SERVER_LIBRARIES` to `false` to disable this cleanup if it causes unexpected behavior.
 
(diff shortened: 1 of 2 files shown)


itzg (approved):
Thanks for the progress

itzg:
Build failure is just the sporadic exhaustion of GitHub token.

harrisonablack:
> Build failure is just the sporadic exhaustion of GitHub token.

@itzg I was briefly looking at this, couldn't we get around rate limits by supplying a github token? From what i see currently we pass a github token to the test scripts, but the docker container never actually picks it up.

inside `test/setuponlytest/test.sh`, could we not just add `-e GH_TOKEN` to the `docker compose run` call?

itzg:
@harrisonablack oh, I think you're totally right. I had fixed it at both ends but hadn't realized it was missing the middle hop.

I'll make that change and see how that goes.

PR #4138, opened 2026-06-27
Add example for loading mods from a local directory

samuelstanley0101:
Setting up a Minecraft server through Docker is likely to be many people's first foray into writing Compose files, as it was mine. I had trouble automatically copying mods from a local directory into a container until I realized I had to mount the directory in the container, so I thought an example of this might be helpful for future people new to Compose files.


itzg (approved):
Great idea and contribution.

PR #4141, opened 2026-06-30
Refactor deploy gtnh

SgtMate:
This PR aims to improve code quality and structure of the deployment script for GregTech New Horizon.
It also comes with a small ajustment of the wiki documentation.

### DIFF
--- a/compose-dev.yml
+++ b/compose-dev.yml
@@ -4,6 +4,15 @@ services:
       context: .
     environment:
       EULA: true
+      TYPE: GTNH
+      GTNH_PACK_VERSION: latest-dev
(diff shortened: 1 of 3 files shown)


itzg:
Glancing at the changes they look good. I just wanted to mention that the complexity in this script is why I started shifting over to mc-image-helper where I could implement these kinds of things in Java 😄

SgtMate:
@itzg yes get it. I think I got to do the same in the long run, but this is the first step to bring the script in a more program like functional structure. I see it as a step towards moving complexity over there or enabling others to do so.

itzg (approved):
Thanks for ongoing work on this.

PR #4144, opened 2026-07-01
Added: riscv64 support

Opvolger:
- added java25-noble (has riscv64 support)
- cloned knock and build it for amd64/arm64/arm32/riscv64

My hardware was not fast enough; it could not keep up with generating. (StarFive VisionFive 2 Lite)

But there are now newer CPUs like the SpaceMit K3, which should not have any problems with this.

Test the code with this command:

```bash
docker buildx build --platform linux/amd64,linux/arm64,linux/riscv64 --build-arg BASE_IMAGE=eclipse-temurin:25-jre-noble --build-arg VERSION=latest --tag opvolger/minecraft-server:latest --push .
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,7 @@ jobs:
         # JAVA 25
           - variant: java25
             # NOTE: 25-jre is also Noble (currently), but this variant needs to be 25-jre-noble specifically to include riscv64
+            # Keep this aligned in verify-pr.yml also
             baseImage: eclipse-temurin:25-jre-noble
             platforms: linux/amd64,linux/arm64,linux/riscv64
             mcVersion: latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,7 @@ jobs:
       matrix:
         # NOTE: the "latest" variant is identified in the Docker meta step's 'latest' config
         variant:
+          - java26-noble
           - java25
           - java25-alpine
           - java25-jdk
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -47,6 +48,10 @@ jobs:
             baseImage: eclipse-temurin:25-jre-alpine
             platforms: linux/amd64,linux/arm64
             mcVersion: latest
+          - variant: java25-noble
```
itzg: What are your thoughts about the java25 variant itself being switched over to this 25-jre-noble? Then it would be one less variant built and published.

itzg (commented):
Thanks for the final step of the saga.

.github/workflows/build.yml
Opvolger: The other Docker images do not have a build for RISCV64 yet.

itzg: Understood. 

What I am pondering is that instead of both java25 and java25-noble, where "noble" has no obvious meaning to end users, just have java25 image variant which is based on 25-jre-noble. 

If both are needed, I'd rather call this new one something like java25-riscv64 so it indicates the purpose to end users. Or java25-morearch or java25-expanded, java25-exp

magretopicus: noble is a release version of ubuntu (24.04 LTS: Noble Numbat), maybe java25-ubuntu as name? just like alpine?

itzg: @magretopicus `eclipse-temurin:25-jre` is also based on Ubuntu and Ubuntu Noble specifically. 

That's why I'm nitpicking the image tag since "noble" and "ubuntu" are equally applicable to `java25` and this new one that includes riscv64. 

That's also why I suggested switching java25 over to avoid the confusion entirely. 

[Here's Gemini's explanation of why the difference in architecture support.](https://share.google/aimode/X0ESD1z8griRYctzx)

Opvolger: Oke, will update the pull request and replace it with noble

itzg (approved):

PR #4176, opened 2026-07-17
Use gcompat to ensure mc-image-helper runs on aarch64+alpine

### LINKED ISSUE #4175: The JVM hard-crashes on startup with SIGSEGV
### Summary

The JVM hard-crashes on startup with SIGSEGV inside the bundled `netty-tcnative` native library. Because the crash happens in a C constructor while the `.so` is being loaded, it's uncatchable and takes the whole process down instead of falling back gracefully.

### Environment

- **Image / tag:** `latest`
- **Base:** Alpine Linux (musl libc)
- **Arch:** aarch64 (AWS EC2 Graviton)
- **JRE:** Eclipse Temurin 25.0.3+9 (LTS)
- **Server type:** Paper

### Crash

```
# SIGSEGV (0xb) at pc=0x0000000000023c30, pid=54, tid=65
# JRE version: OpenJDK Runtime Environment Temurin-25.0.3+9 (25.0.3+9) (build 25.0.3+9-LTS)
# Java VM: OpenJDK 64-Bit Server VM Temurin-25.0.3+9 (25.0.3+9-LTS, ... serial gc, linux-aarch64)
# Problematic frame:
# C  [libnetty_tcnative_linux_aarch_64....so+0x2476c]  init_have_lse_atomics+0xc
```

### Root cause

`init_have_lse_atomics` is GCC's outline-atomics (`-moutline-atomics`) runtime probe. It calls glibc's `__getauxval(AT_HWCAP)` to detect ARM LSE atomic support. The `netty-tcnative` `.so` shipped in this image is compiled against **glibc**, but Alpine uses **musl**, where `__getauxval` isn't provided the same way — so the probe dereferences a bad pointer and segfaults during `dlopen`. This is a known netty-tcnative-on-musl/aarch64 failure mode (see grpc-java#12567, netty#12669, grpc-java#10930).

### Container definition

```
```

### Container logs

```
INFO   Resolving type given PAPER
         INFO   #
         INFO   # A fatal error has been detected by the Java Runtime Environment:
         INFO   #
         INFO   #  SIGSEGV (0xb) at pc=0x0000000000023c30, pid=54, tid=65
         INFO   #
         INFO   # JRE version: OpenJDK Runtime Environment Temurin-25.0.3+9 (25.0.3+9) (build 25.0.3+9-LTS)
         INFO   # Java VM: OpenJDK 64-Bit Server VM Temurin-25.0.3+9 (25.0.3+9-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, serial gc, linux-aarch64)
         INFO   # Problematic frame:
         INFO   # C  [libnetty_tcnative_linux_aarch_6417342692647055256540.so+0x2476c]  init_have_lse_atomics+0xc
         INFO   #
         INFO   # Core dump will be written. Default location: Determined by the following: "/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h" (alternatively, falling back to /data/core.54)
```

itzg:
Resolves #4175

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,12 +45,12 @@ jobs:
             baseImage: eclipse-temurin:25-jre-noble
             platforms: linux/amd64,linux/arm64,linux/riscv64
             mcVersion: latest
+            # forked build that include riscv64
+            knockdRepoOrg: Opvolger/knock
           - variant: java25-alpine
             baseImage: eclipse-temurin:25-jre-alpine
             platforms: linux/amd64,linux/arm64
             mcVersion: latest
-            # For GLIBC_2.34 support (and Alpine)
-            knockdRepoOrg: Metalcape/knock
           - variant: java25-jdk
             baseImage: eclipse-temurin:25
             platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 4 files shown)


itzg:
I was able to reproduce the original issue on my MacBook

PR #4191, opened 2026-07-26
Add daily build options to gtnh example

SgtMate:
Added options for daily builds and GitHub token in the docker-compose file for gtnh.


itzg (approved):
Good idea

PR #4197, opened 2026-07-29
docs: add seedloaf sponsor

jasperchess:
Two small formatting things (it was my auto formatter, happy to remove if preferable).

Also added Seedloaf to the sponsor section. $25 feels ridiculously low, but as we grow I hope we can start contributing and paying more too. We love the ecosystem and used quite a few different projects.

Thanks Geoff!


itzg (approved):
I greatly appreciate it and glad to be part of your solution.

PR #4205, opened 2026-08-01
Restore the GraalVM image variants

OowhitecatoO:
You said in #4203 that you would be interested in putting these back if the builds work again, so here is the change with what I have to support that.

The three variants are re-added to the matrix with their original base images, the docs table and its two notes are restored, and the knockd retrieval that #3893 stubbed out is put back. That last one is included because the images published before the variants were dropped still contain the real `knockd` binary — restoring the variants without it would be the first GraalVM images to ship the stub, and would silently take auto-pause away from anyone using it on those tags.

Builds pass on my fork for all three variants, amd64 and arm64: https://github.com/OowhitecatoO/docker-minecraft-server/actions/runs/30679759114. The `tar: Cannot open: Invalid argument` failure behind #3899 did not reappear, including with the knockd retrieval restored: https://github.com/OowhitecatoO/docker-minecraft-server/actions/runs/30649087679.

I did not restore `java8-graalvm-ce`. It is the only one on the community edition from ghcr rather than Oracle GraalVM, and it looked like it had been left behind long before this.

Separately, I have been through what upstream MeowIce has changed since those flags were copied into the image, and will send a small PR for that on its own.

What I can and cannot claim about the build failure: I could not find out why it failed in January, only that it does not fail now. The runs above are on current `master` with the matrix trimmed to the GraalVM variants and the test step skipped, so that the multi-arch build step is reached quickly; nothing else in the build path is modified. If it turns out to have been intermittent, this brings it back.

Beyond the CI builds I also ran the images: two real modded servers, NeoForge 1.21.1 on a locally built `java25-graalvm` and Forge 1.19.2 on `java21-graalvm`, both migrated from the frozen published images with no configuration change. Both reached `Done`, passed the health check and answered RCON. Locally built images report Oracle GraalVM 21.0.12 and 25.0.4, against 21.0.9 and 25.0.1 in the last published ones, so the base image tags have kept moving while nothing was rebuilding them.

Docs changes: the three rows go back into the tag table with the two notes they used to carry — the Oracle GraalVM licence note, and the one about Forge Installer not working on Oracle Linux because of zlib-ng — and the tags come out of the deprecated list.

</details>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,10 +30,13 @@ jobs:
           - java25
           - java25-alpine
           - java25-jdk
+          - java25-graalvm
           - java21
           - java21-alpine
+          - java21-graalvm
           - java21-jdk
           - java17
+          - java17-graalvm
           - java16
           - java11
           - java8
@@ -54,6 +57,10 @@ jobs:
             baseImage: eclipse-temurin:25
             platforms: linux/amd64,linux/arm64
             mcVersion: latest
+          - variant: java25-graalvm
+            baseImage: container-registry.oracle.com/graalvm/jdk:25-ol10
+            platforms: linux/amd64,linux/arm64
+            mcVersion: latest
         # JAVA 21:
           - variant: java21
             baseImage: eclipse-temurin:21-jre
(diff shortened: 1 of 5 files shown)


itzg (commented):
Thanks for another great contribution and the follow up here. 

In https://github.com/itzg/docker-minecraft-server/blob/master/.github/workflows/verify-pr.yml I have selected a subset of variants to test. A subset is used to keep the build "effort" low but hopefully catch major issues early. 

Can you add the Java 25 variant here?

OowhitecatoO:
Added it. It failed on arm64:

```
[linux/arm64] RUN /build/run.sh setup-user
Error while loading /build/ol/setup-user.sh: Exec format error
```

That is a pre-existing bug rather than anything about the variant. `build/ol/setup-user.sh` has no shebang, unlike the alpine and ubuntu ones, and `run.sh` executes it directly. On amd64 the shell falls back to running it after `ENOEXEC`, so it has always worked. Under emulation, buildkit's qemu wrapper tries to load it as a binary and fails. It only shows up on Oracle Linux with arm64 emulation, which nothing was covering.

With the shebang added, all five variants pass including the tests: https://github.com/OowhitecatoO/docker-minecraft-server/actions/runs/30724451300

Two things to flag:

- I set `linux/amd64,linux/arm64` on the new entry rather than amd64 only like java17 and java8, because arm64 is where this kind of problem appears. Happy to trim it if that is more CI time than you want.
- I also added `docker/setup-qemu-action` to verify-pr, since build.yml has it and this one did not. It did not fix the above — I assumed it would and it made no difference — but the two workflows differing seemed worth fixing. Happy to drop that commit if you would rather keep them separate.

itzg:
The latest changes are good. Thanks.

itzg (approved):

Issue #3433, opened 2025-05-01
Autopause & playitgg: not resuming when trying to connect via playitgg


DerRockWolf:
### Describe the problem

I'm using the autopause feature:

```
ENABLE_AUTOPAUSE="true"
AUTOPAUSE_TIMEOUT_EST="900"
AUTOPAUSE_TIMEOUT_INIT="300"
```

Players can connect via playit.gg just fine, but if the server got paused knockd does not seem to notice connection attempts. Connection attempts from my local network are able to wake up the server just fine.

I'm not sure if the issue is with the knockd config or with how playit.gg does things.
Any ideas how I could investigate would be great.

itzg:
Auto-pause and knockd was a contributed feature so I can't offer much advice myself. All of the logic is in

https://github.com/itzg/docker-minecraft-server/blob/master/files/auto/autopause-daemon.sh

My only suggestion is to consider different values for `AUTOPAUSE_KNOCK_INTERFACE`.

DerRockWolf:
I've tried setting `AUTOPAUSE_KNOCK_INTERFACE` to `lo` as this is the only other interface and from what I could find the playit.gg agent forwards requests to `127.0.0.1`.

I've used a patched image that adds `--include-loopback` to the `mc-image-helper network-interfaces` calls in `files/auto/autopause-daemon.sh`: <details><summary>Patch</summary>
<p>

```diff
diff --git a/files/auto/autopause-daemon.sh b/files/auto/autopause-daemon.sh
index 6bb0233..b614abd 100644
--- a/files/auto/autopause-daemon.sh
+++ b/files/auto/autopause-daemon.sh
@@ -10,7 +10,7 @@ fi

 autopause_error_loop() {
   logAutopause "Available interfaces within the docker container:"
-  logAutopause "  $(mc-image-helper network-interfaces)"
+  logAutopause "  $(mc-image-helper network-interfaces --include-loopback)"
   logAutopause "Please set the environment variable AUTOPAUSE_KNOCK_INTERFACE to the interface that handles incoming connections."
   logAutopause "If unsure which interface to choose, run the ifconfig command in the container."
   logAutopause "Autopause failed to initialize. This log entry will be printed every 30 minutes."
@@ -35,7 +35,7 @@ if [[ -z "$AUTOPAUSE_KNOCK_INTERFACE" ]] ; then
   logAutopause "AUTOPAUSE_KNOCK_INTERFACE variable must not be empty!"
[...]

Issue #3315, opened 2025-02-16
Auto-pause for proxied servers


ASTRELION:
### Enhancement Type

Improve an existing feature

### Describe the enhancement

Not sure if this counts as a bug or a feature so using a feature request for now.

I have a Minecraft Paper server running with this container, and it runs behind a Velocity proxy provided by https://github.com/itzg/docker-mc-proxy. I'd like to be able to auto-pause the servers behind the proxy and only unpause them when they're actually attempting to be connected to. Bonus points for also supporting auto-stop.

Currently, if I setup the Minecraft server to autopause, its constantly being woken up (I'm assuming due to the proxy because it's being pinged for info, as noted below).

I'd like this from 2 perspectives:
1. I have a website for my server that displays essentially the same information as what you'd see in a server list, but that means that my server is almost always being pinged because of requests from the website (genuine users, bots, etc).
2. It would be nice to have players able to connect to my proxy and other servers behind the proxy and only wake servers that players are actually on.

I'm not sure if this is all within the scope of the project or even possible; let me know if there's a solution already out there.

itzg:
Your expectations are reasonable, but yes that all is outside the scope of this project.

Perhaps take a look at the two examples here

https://docker-minecraft-server.readthedocs.io/en/latest/misc/examples/#lazymc-put-your-minecraft-server-to-rest-when-idle

Issue #2666, opened 2024-02-17
Auto Pause Not Compatable with `SERVER_PORT` - Still Uses 25565


harrismcc:
### Describe the problem

When using the auto pause feature and `SERVER_PORT` the container crashes with the debug line:
```
+++ mc-monitor status --host localhost --port 25565 --show-player-count
failed to ping localhost:25565 : could not connect to Minecraft server: dial tcp 127.0.0.1:25565: connect: connection refused++ connections=
```

It seems like mc-monitor is still using the default 25565 port instead of the port set by `SERVER_PORT`

### Container definition

env vars:
```
        EULA=true
        TYPE=MODRINTH
        VERSION=1.20.1
        MODRINTH_LOADER=fabric
        MODRINTH_MODPACK=[redacted]
        MODRINTH_EXCLUDE_FILES=e4mc

        # JVM config
        MEMORY=2G
        # Auto Pause
        ENABLE_AUTOPAUSE=true
        AUTOPAUSE_KNOCK_INTERFACE=eth0
        DEBUG_AUTOPAUSE=true
        AUTOPAUSE_TIMEOUT_EST=15
        AUTOPAUSE_TIMEOUT_INIT=15
```

### Container logs

```
+ isTrue true
+ case "${1,,}" in
+ return 0
+ log 'DEBUG: autopause state = S'
+ local oldState
++ shopt -po xtrace
+ oldState='set -o xtrace'
+ shopt -u -o xtrace
+ case X$STATE in
+ rcon_client_exists
++ ps -ax -o comm
++ grep rcon-cli
+ [[ -n '' ]]
+ java_running
++ ps -ax -o stat,comm
++ grep java
++ awk '{ print $1 }'
+ [[ Tl =~ ^S.*$ ]]
+ [[ S == \S ]]
+ sleep 5
+ :
+ isTrue true
+ case "${1,,}" in
+ return 0
+ log 'DEBUG: autopause state = S'
+ local oldState
++ shopt -po xtrace
+ oldState='set -o xtrace'
+ shopt -u -o xtrace
+ case X$STATE in
+ rcon_client_exists
++ ps -ax -o comm
++ grep rcon-cli
+ [[ -n '' ]]
+ java_running
++ ps -ax -o stat,comm
++ grep java
++ awk '{ print $1 }'
+ [[ Sl =~ ^S.*$ ]]
+ java_clients_connected
++ java_clients_connections
++ local connections
++ java_running
[...]

Issue #1046, opened 2021-09-15
Autopause with Geyser


kjburr:
I have a small feature request to add Geyser(https://github.com/GeyserMC/Geyser) functionality with autopause.  Currently if you connect from a bedrock client on a paused server it does not detect the bedrock client at all.  The server also pauses if your connected and playing on the server ie not detecting the bedrock client.

I've tried diving into this myself and went as far as editing the knockd-config.cfg and adding the extra port but that hadn't made a difference.  Maybe it's something that can be added later on.  Thank you

itzg:
When you tried the config change did you account for 19132 being UDP with

```
sequence = 19132:udp 
```

and removing the `tcpflags`?

kjburr:
Yea I did actually.  My config looks like this

[unpauseMCServer-server]
 sequence = 25565,19132:udp
 seq_timeout = 1
 command = /usr/sbin/gosu minecraft:minecraft /autopause/resume.sh
 tcpflags = syn
 
 Sorry I should've mentioned I was trying to allow both 25565 and 19132 to have autopause as I'm running Geyser as a plugin.  Would removing the tcpflags still be required in this case?  Thanks again

itzg:
Since that is the sequence of port knocks you'll instead want to try a whole separate section like

```
[unpauseMCServer-bedrock]
 sequence = 19132:udp 
 command = /usr/sbin/gosu minecraft:minecraft /autopause/resume.sh
```

kjburr:
That worked! The only other thing I'm unsure if it's intended but RCON constantly tries to save when it tries to pause.  If the bedrock client is minimized to the background the server will finally pause kicking the client.  This isn't an issue for the java side I'm guessing because of the differences between how the tcp/udp protocols work.

 [01:46:02 INFO]: kjburr joined the game
[01:46:02 INFO]: kjburr[/172.20.0.1:0] logged in with entity id 3363 at ([world]181.53487, 83.00181, -319.29227)
[Autopause loop] No client connected since startup / knocked - stopping
[01:46:27 INFO]: Thread RCON Client /127.0.0.1 started
[01:46:27 INFO]: [Rcon: Saved the game]
[01:46:27 INFO]: Thread RCON Client /127.0.0.1 shutting down
[2021-09-16T01:46:27+00:00] [Autopause] Pausing Java process
[Autopause loop] Server was knocked - waiting for clients or timeout

[...]

PR #4221, opened 2026-08-07
autopause: warn when the status ping fails instead of silently never pausing

jairgs:
## Problem

`java_clients_connections` treats a failed status ping as one connected player:

```bash
if ! connections=$(mc-monitor status ... --show-player-count); then
  # consider it a non-zero player count if the ping fails
  # otherwise a laggy server with players connected could get paused
  connections=1
fi
```

That fail-safe is correct for a *transient* failure — pausing a laggy server with players on it would be much worse than staying awake. But it assumes the failure is transient. When the ping fails *permanently*, the server can never observe zero players: the state machine parks in `E` ("Client connected - waiting for disconnect") and never pauses, and nothing is logged to say why.

The symptom is a server that silently never sleeps, which is indistinguishable from "someone is always playing" unless you check the JVM's process state by hand.

## How I hit it

A modpack shipping the [Connectivity](https://www.curseforge.com/minecraft/mc-mods/connectivity) mod, which is bundled in many large Forge packs. Its `malformedtraffic` filter blacklists a source after repeated packet sequences that never complete a login — which is exactly what a server-list-ping health check looks like at the packet level, repeated every `AUTOPAUSE_PERIOD` forever. About three minutes into every boot:

```
[Netty Epoll Server IO #2/WARN] [com.connectivity.Connectivity/]:
Ignoring further traffic from: 0:0:0:0:0:0:0:1%0 for repeated malformed traffic.
```

From then on the ping from loopback got no response, so `connections=1` on every cycle. Concretely:

```
07:00:35  server boots
07:03:35  Connectivity blacklists ::1
17:51:32  last player leaves
18:08:00  still running, 21% CPU, 9GB RSS, nobody online
```

`AUTOPAUSE_TIMEOUT_EST` is 180s, so it was 16 minutes overdue. Checking the rotated logs, that same block had fired on **every boot going back to mid-July** — the server had not paused once in three weeks, and there was no indication anywhere. I only found it because I was investigating something unrelated and happened to look at `ps`.

This isn't specific to that mod. Anything that makes the status ping fail persistently produces the same silent non-pausing, and the existing issues for Geyser (#1046), playit.gg (#3433), proxied servers (#3315) and non-default `SERVER_PORT` (#2666) all share the shape: autopause's detection assumption breaks under some networking arrangement, and it's hard to tell from the outside.

## Change

Warn on the first failure and every 30th after that, and log once on recovery.

Two implementation details worth calling out, since both are why this logging didn't already exist:

- **stderr, not stdout.** `logAutopause` writes to stdout, and `java_clients_connections` returns its value via stdout capture, so a naive log line would corrupt the player count. Verified the function still returns a bare integer.
- **File-backed counter.** The function is called through command substitution, so it runs in a subshell and a shell variable would reset on every call — the first version of this patch logged on *every* failure for exactly that reason. The path is overridable via `AUTOPAUSE_PING_FAILURE_STATE`.

## Testing

Harness stubbing `mc-monitor`, `java_running` and `logAutopause`, asserting the stdout contract and the rate limiting:

```
PASS  1st failure: stdout is bare '1'
PASS  1st failure: warns once
PASS  failures 2-29: stdout still '1'
PASS  failures 2-29: stays silent
PASS  30th failure: warns again
PASS  recovery: stdout is player count '2'
PASS  recovery: logs once
PASS  recovery reports count
PASS  steady state: no spam
PASS  steady state: stdout '2'
PASS  java stopped: stdout '0'
```

Happy to adjust the thresholds, wording, or drop the recovery message if you'd rather keep it minimal.

### DIFF
--- a/files/auto/autopause-fcns.sh
+++ b/files/auto/autopause-fcns.sh
@@ -34,8 +34,22 @@ mc_server_listening() {
   mc-monitor status $(use_proxy) --host "${SERVER_HOST:-localhost}" --port "$SERVER_PORT" $(use_server_list_ping) --timeout 10s >&/dev/null
 }
 
+# Tracks consecutive status-ping failures so the warning below can be rate
+# limited instead of repeating every AUTOPAUSE_PERIOD. Held in a file rather
+# than a variable because java_clients_connections is called through command
+# substitution, so it runs in a subshell and any variable it sets is lost.
+autopause_ping_failures_file="${AUTOPAUSE_PING_FAILURE_STATE:-/tmp/.autopause-status-ping-failures}"
+
+autopause_ping_failures() {
+  local count
+  count=$(cat "${autopause_ping_failures_file}" 2>/dev/null)
+  [[ ${count} =~ ^[0-9]+$ ]] || count=0
+  echo "${count}"
+}
+
 java_clients_connections() {
   local connections
+  local failures
   if java_running; then
     if ! connections=$(mc-monitor status \
         --host "${SERVER_HOST:-localhost}" \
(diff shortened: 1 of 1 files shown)


files/auto/autopause-fcns.sh
```diff
@@ -0,0 +1,88 @@
...
+
+mc_server_listening() {
+  mc-monitor status $(use_proxy) --host "${SERVER_HOST:-localhost}" --port "$SERVER_PORT" $(use_server_list_ping) --timeout 10s >&/dev/null
+}
+
+# Tracks consecutive status-ping failures so the warning below can be rate
+# limited instead of repeating every AUTOPAUSE_PERIOD. Held in a file rather
+# than a variable because java_clients_connections is called through command
+# substitution, so it runs in a subshell and any variable it sets is lost.
+autopause_ping_failures_file="${AUTOPAUSE_PING_FAILURE_STATE:-/tmp/.autopause-status-ping-failures}"
```
itzg, line 41: `/tmp` can be problematic on some container systems -- I'm still not sure which ones and why, but I have had to gradually fix issues by moving from `/tmp` to `/data/.tmp`. Feel free to do a `mkdir -p` here to prepare that since only scripts that need it will be creating that.

itzg (commented):
Sorry for the delay on reviewing this. Just the minor change would be good.

itzg:
@jairgs I just wanted to see if you were planning on making the requested changes?

PR #4223, opened 2026-08-09
docs: Add docs for CurseForge regex filename matcher

### LINKED ISSUE #3686: Provide option for regex with CF_FILENAME_MATCHER
[From Discord](https://discord.com/channels/660567679458869252/660569641550217327/1423336730064457819)

It seems the `CF_FILENAME_MATCHER` it's doing a `.contains(...)` (I verified in the code), but how could I restrict that a little more?  With ATM10 TTS, they had version `1.5` then went to `1.5.1` and quickly went to `1.5.1b` (as a hotfix).  I had as a matcher: `-1.5` since their filename was `ATM10 To the Sky-1.5.zip` and had been playing on `1.5` for a while now. But my server restarted (due to a crash) and it upgraded to `1.5.1b`.  Which since it's using contains, I can see why.  Granted, they're using poor versioning, but it would be nice be a little more specific with the specification.  Maybe a regex option or something?

Side note: while it downloaded and is running `1.5.1b` (I checked the logs), the `%MODPACK_VERSION%` variable and value in `.install-curseforge.env` report it as `1.5.1` instead.

Proposal: when the value is a slash-surrounded string then it will be processed as a regex such as 

```
/.*1\.5$/
```

harrisonablack:
Documents support for matching curseforge filenames using regex patterns, added in itzg/mc-image-helper#823, updated in itzg/docker-minecraft-server#4222

### DIFF
--- a/docs/types-and-platforms/mod-platforms/auto-curseforge.md
+++ b/docs/types-and-platforms/mod-platforms/auto-curseforge.md
@@ -127,7 +127,6 @@ The following examples all refer to version 1.0.7 of ATM8:
 ```
 
 ```yaml
-  # Matches filenames containing "1.0.7"
   CF_SLUG: all-the-mods-8
   CF_FILENAME_MATCHER: 1.0.7
 ```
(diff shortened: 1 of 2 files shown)


itzg (approved):
Thanks. I had completely forgotten about the docs.

PR #4231, opened 2026-08-13
Add AutoStopper to related projects

### LINKED ISSUE #4229: Add AutoStopper to the Related Projects page
### Enhancement Type

Improve an existing feature

### Describe the enhancement

Hello! I maintain AutoStopper (https://github.com/Criseda/AutoStopper), a Velocity plugin that manages Docker lifecycle for backend servers on a proxy network. I'd like to propose adding it to the Related Projects page (`docs/misc/related-projects.md`) and would be happy to open the docs PR if it's something you'd consider.

I'm aware of `ENABLE_AUTOSTOP`, that feature stops a container after an idle timeout and requires manual restart afterwards. AutoStopper runs on the proxy and provides the full sleep/wake cycle for proxy-routed backends. It stops mapped containers when empty and starts them again when a player connects, waiting on a real readiness signal before handoff. It manages several backends from one config and detects idleness from actual Velocity sessions rather than the backend's own connection state. So it's a complement for multi-server Velocity setups rather than a replacement.

Proposed entry, matching the existing format:
```md
 [AutoStopper](https://github.com/Criseda/AutoStopper)

A Velocity plugin that stops idle backend servers and wakes them when players connect, with readiness-aware Docker lifecycle management.
```
Happy to adjust wording or drop this if it doesn't fit. Thanks you for considering!

Criseda:
Adds AutoStopper to the Related Projects page, as discussed in #4229.

### [AutoStopper](https://github.com/Criseda/AutoStopper)

A Velocity plugin that stops idle backend servers and wakes them when players connect, with readiness-aware Docker lifecycle management.

- Docs-only change; no behavior or configuration impact.
- Entry follows the existing format of the page.

### DIFF
--- a/docs/misc/related-projects.md
+++ b/docs/misc/related-projects.md
@@ -29,3 +29,7 @@ A tool that is bundled with this image to provide complex, re-usable preparation
 ### [itzg/rcon](https://github.com/itzg/docker-rcon-web-admin)
 
 An image that dockerizes [rcon-web-admin](https://github.com/rcon-web-admin/rcon-web-admin).
+
+### [AutoStopper](https://github.com/Criseda/AutoStopper)
+
+A Velocity plugin that stops idle backend servers and wakes them when players connect, with readiness-aware Docker lifecycle management.


itzg (approved):
Great to see the addition.

PR #4239, opened 2026-08-18
Add missing HARDCORE environment variable to variables docs

eithe:
This PR adds the HARDCORE environment variable to the reference at variables.md, next to the similar DIFFICULTY setting.

I was visiting the docs to check whether hardcore was available as a variable and ended up blindly scrolling past the warning that the page may be outdated ...

This change documents the variable so that others can discover it more easily when looking for the corresponding server setting.


itzg (approved):
Yep even though I don't actively maintain the page, I greatly appreciate the update.

PR #4247, opened 2026-08-27
pin gosu, fail closed on BuildTools download, reject zip slip

lbellows:
## Summary

- `COPY --from=tianon/gosu` used an untagged `latest` image. Pin to `1.19`.
- Require a successful BuildTools download (`curl -fsSL`) before compiling Spigot.
- Refuse world/modpack archives whose entries contain path traversal (`../` or absolute paths) before extract. Applies to zip, tar/gz/bz2, and zstd.

## Test plan

- [x] `bash -n scripts/start-utils scripts/start-deployBukkitSpigot`
- [x] Confirmed `Dockerfile` copies `tianon/gosu:1.19`
- [x] Zip with `../evil.txt` and `/abs.txt` is rejected; clean `world/level.dat` zip extracts
- [x] Tar with `../evil.txt` is listed by `tar -tf` and rejected by the same regex; clean tar is allowed
- Full `tests/setuponlytests` Docker suite was not run here (no Docker daemon)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ ARG KNOCKD_REPO_ORG=Metalcape/knock
 RUN --mount=target=/build,source=build \
     TARGET=${TARGETARCH}${TARGETVARIANT} \
     /build/run.sh install-packages
-COPY --from=tianon/gosu /gosu /usr/local/bin/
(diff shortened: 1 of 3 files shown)


itzg (approved):
Thanks! I created some follow up issues build upon this.

lbellows:
I have some tokens to burn so sending some through :)

PR #4249, opened 2026-08-27
fix: drop extra brace in Forge/NeoForge FORCE_REINSTALL defaults

lbellows:
## Summary

- `: "${FORGE_FORCE_REINSTALL:=false}}"` (and the NeoForge copy) assigned the string `false}` when the variable was unset.
- That value is passed through as `--force-reinstall=false}`.

## Test plan

- `bash -n scripts/start-deployForge scripts/start-deployNeoForge`
- Confirmed the default assignment is now `false` in both scripts

PR #4251, opened 2026-08-27
Do not mark an FTB App pack installed after a failed installer

lbellows:
## Summary

- The FTB installer is piped to `tee` without `pipefail`, so a non-zero installer exit was ignored.
- `.ftb-installed` was still written, and the next start skipped install then failed to find a server jar.

## Test plan

- `bash -n scripts/start-deployFTBA`
- Confirmed `set -o pipefail` is enabled and the installer pipeline is checked before writing the marker


itzg:

itzg (approved):
Thanks

PR #4252, opened 2026-08-27
Compute percentage memory sizes with integer arithmetic

lbellows:
## Summary

- `normalizeMemSize` used `scale=0.01` inside `$(( ))`, which bash cannot evaluate.
- Everyday `MEMORY=80%` does not hit that function (`getPercentageValue` → `-XX:MaxRAMPercentage`). The broken callers are Aikar's 12GB flag split (`USE_AIKAR_FLAGS` + `MAX_MEMORY`) and the deprecated CurseForge `settings-local.sh` conversion.
- A failed Aikar compare does not abort startup; it falls through to the ≤12GB flags. The CurseForge path also multiplied already-byte values by 1048576 instead of converting to MB.

## Test plan

- `bash -n scripts/start-utils scripts/start-finalExec`
- Sourced `normalizeMemSize` locally: `12g`/`8G`/`1024M` convert as expected; `8G < 12g`; invalid input is rejected; `50%` returns an integer instead of a syntax error

### DIFF
--- a/scripts/start-finalExec
+++ b/scripts/start-finalExec
@@ -513,11 +513,15 @@ elif [[ ${TYPE} == "CURSEFORGE" ]]; then
 
 if isPercentage "$INIT_MEMORY" || isPercentage "$MAX_MEMORY"; then
 # Convert to bytes
-NORM_INIT_MEM=$(normalizeMemSize "$INIT_MEMORY")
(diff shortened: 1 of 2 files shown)


itzg:

itzg (approved):
Thanks!

PR #4253, opened 2026-08-27
fix: make REMOVE_OLD_DATAPACKS match zip files by basename

lbellows:
## Summary

- `find -wholename "*.zip"` never matches `/data/<world>/datapacks/foo.zip` because `*` does not cross `/`.
- `REMOVE_OLD_DATAPACKS` was therefore a no-op with the documented default include.
- Depth defaulted to `"1 "` (trailing space) instead of the documented 16.

## Test plan

- `bash -n scripts/start-setupDatapack`
- Confirmed prune uses `-name` and `REMOVE_OLD_DATAPACKS_DEPTH` defaults to `16`

### DIFF
--- a/scripts/start-setupDatapack
+++ b/scripts/start-setupDatapack
@@ -6,7 +6,7 @@ set -e -o pipefail
 : "${VANILLATWEAKS_SHARECODE:=}"
 : "${REMOVE_OLD_DATAPACKS:=false}"
 : "${DATAPACKS_FILE:=}"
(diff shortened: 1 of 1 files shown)


itzg:

lbellows:
Closing this one to save you a review. The `-wholename "*.zip"` claim is wrong on GNU find, and the depth 1→16 change is behavior, not a bug.

itzg:
Thanks

PR #4254, opened 2026-08-27
Stop startup when a MODPACK zip cannot be extracted

lbellows:
## Summary

- `handleModpackZip` logged unzip failures inside `if !`, which bypasses `set -e`.
- The zip was then deleted and setup continued without the mods/plugins.

## Test plan

- `bash -n scripts/start-setupModpack`
- Confirmed both unzip failure branches now `exit 2`


itzg:

itzg (approved):
Thanks

itzg:
@lbellows , I do want to say that I appreciate that you have been keeping each PR to a small, single purpose. I much prefer that rather than a big PR that fixes 20 unrelated things.

lbellows:
of course @itzg , that way you could take what you want and leave what you don't.

PR #4255, opened 2026-08-27
Resolve SpongeVanilla latest version and fail closed on download

lbellows:
## Summary

- jq treated `$SPONGEBRANCH` as a literal path segment (`.buildTypes.$SPONGEBRANCH`) instead of `.buildTypes[$SPONGEBRANCH]`.
- Auto-detect without `SPONGEVERSION` therefore never read `buildTypes.stable`.
- The jar download used `curl -sSL` without `-f`, so a 404 was saved as the server jar.

## Test plan

- `bash -n scripts/start-deploySpongeVanilla`
- Existing setup-only test sets `SPONGEVERSION` and still covers the pinned-version path

### DIFF
--- a/scripts/start-deploySpongeVanilla
+++ b/scripts/start-deploySpongeVanilla
@@ -22,18 +22,28 @@ case "$SPONGEBRANCH" in
 esac
 
 # If not SPONGEVERSION selected, detect last version on selected branch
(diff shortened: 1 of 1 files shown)


itzg:

itzg (approved):
Much better

PR #4257, opened 2026-08-27
Honor cd failure when starting a CurseForge pack

lbellows:
## Summary

- `cd \"\${FTB_DIR}\" || (logError ...; exit 1)` runs `exit` in a subshell.
- A missing `FTB_DIR` did not stop the parent script, which could then exec the server from `/data`.

## Test plan

- `bash -n scripts/start-finalExec`
- Confirmed both CurseForge `cd` sites now use `{ logError ...; exit 1; }`

### DIFF
--- a/scripts/start-finalExec
+++ b/scripts/start-finalExec
@@ -500,7 +500,7 @@ fi
 if [[ ${TYPE} == "CURSEFORGE" && "${SERVER}" ]]; then
   copyFilesForCurseForge
 
-  cd "${FTB_DIR}" || (logError "Can't go into ${FTB_DIR}"; exit 1)
+  cd "${FTB_DIR}" || { logError "Can't go into ${FTB_DIR}"; exit 1; }
   log "Starting CurseForge server in ${FTB_DIR}..."
   if isTrue "${DEBUG_EXEC}"; then
     set -x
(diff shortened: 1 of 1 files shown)


itzg:

scripts/start-finalExec
```diff
@@ -500,7 +500,7 @@ fi
 if [[ ${TYPE} == "CURSEFORGE" && "${SERVER}" ]]; then
   copyFilesForCurseForge
 
-  cd "${FTB_DIR}" || (logError "Can't go into ${FTB_DIR}"; exit 1)
+  cd "${FTB_DIR}" || { logError "Can't go into ${FTB_DIR}"; exit 1; }
```
itzg: I feel like the original code was trying to be too fancy...and was error prone as you found. I'm thinking could either be a `-d` check for the directory or to blend with existing


```suggestion
  if ! cd "${FTB_DIR}"; then
    logError "Can't go into ${FTB_DIR}"
    exit 1
  fi
```

itzg (commented):
I like your change, but feel like the original code wasn't my style anyway.

scripts/start-finalExec
lbellows: right on!

lbellows: Applied — thanks. I also used the same form for the second `cd "${FTB_DIR}"` in the FTB block (line 547), so both sites match the `if ! ...; then` idiom used elsewhere in `scripts/`.

I skipped the `-d` alternative since a `-d` check would miss a permission failure on the `cd` itself.

itzg (approved):
Thanks for the quick update

PR #4258, opened 2026-08-27
Fail closed when CUSTOM_SERVER download returns an HTTP error

lbellows:
## Summary

- `TYPE=CUSTOM` downloaded `CUSTOM_SERVER` with `curl -sSL`, which treats 404 as success.
- The error body was then used as the server jar.
- Require `-f` and log the failure as an error.

## Test plan

- `bash -n scripts/start-deployCustom`
- Confirmed the download uses `curl -fsSL` and `logError`

### DIFF
--- a/scripts/start-deployCustom
+++ b/scripts/start-deployCustom
@@ -16,8 +16,8 @@ if isURL "${CUSTOM_SERVER}"; then
     log "Using previously downloaded jar at ${SERVER}"
   else
    log "Downloading custom server jar from ${CUSTOM_SERVER} ..."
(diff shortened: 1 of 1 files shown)


itzg:

itzg (approved):
Thanks. I'm surprised I missed that one way back when. Typing `-fsSL` is usually second nature for me.

itzg (commented):
...actually, the `-f` only handles following redirects. Why is the thought that'll handle 404's?

itzg:
nm, `-f` is for fail not follow. Now I'm real confused.

lbellows:
it was a late night !

PR #4259, opened 2026-08-27
Prune CurseForge libraries/mods using relative find paths

lbellows:
## Summary

- Two `find` callers used `-path /libraries/*` / `-path /mods/*`, which only match those absolute locations.
- Jars under the pack'\''s own `libraries/` or `mods/` were therefore not pruned.
- The other finder in the same script already used `*/libraries/*`; align the remaining two.

## Test plan

- `bash -n scripts/start-deployCF`
- Confirmed both remaining finds now prune `*/libraries/*` and `*/mods/*`


itzg:

itzg (approved):
Thanks

PR #4260, opened 2026-08-28
fix quadlet example socket access

lethedata:
Adds the group to support v1.46.0+


itzg (approved):
Thanks for keeping that updated.

PR #4261, opened 2026-08-28
Stop figura exclude from matching configurable jars

### LINKED ISSUE #4234: Two Modrinth modpacks fail to start because the default exclude list drops a mod that Minecraft needs on the server. Both are one-line data fixes; no code changes.
`figura` matches `configurable` as a substring

COBBLEVERSE (`cobbleverse` 1.7.42, Fabric / MC 1.21.1) aborts at load:

```
Immediate reason: [HARD_DEP_NO_CANDIDATE neruina 3.3.3 {depends configurable @ [~3.5.2]},
                   ROOT_FORCELOAD_SINGLE neruina 3.3.3]
```

`mods/configurable-3.5.2+1.21.1-fabric.jar` is declared in `modrinth.index.json` with `env = {server: required, client: required}`, and it satisfies Neruina's `[3.5.2, 3.6-)`. It is never downloaded.

The cause is that `FileInclusionCalculator.shouldExcludeFile` matches entries as unanchored substrings of the file path, so the client-only entry `figura` matches `con`**`figura`**`ble-…jar`.

Measured rather than assumed:

- The pack manifest lists 168 files; the boot log has 145 `Downloaded` lines.
- Replaying the exclude list against the 24 missing paths matches **24/24 with no false positives or negatives** — 23 are intended client-only mods, 1 is `configurable`.
- `figura` is the **only** one of the entries that substring-matches that filename.
- There is no log line for the exclusion at any level, so this is silent.

## Proposal

A slash denoted regex option should be utilized like the solution for https://github.com/itzg/docker-minecraft-server/issues/3686

lbellows:
Fixes #4234

Default Modrinth exclude entries are unanchored path substrings, so `figura` also matched `configurable-*.jar` and dropped a server-required mod. The global exclude is now `/(^|/)figura-/`, using the slash-delimited regex matcher from mc-image-helper 1.67.

The cobbleverse `configurable` force-include is removed; it only existed to cover that collision. Helper is bumped 1.66.1 → 1.67.0 so the regex is actually evaluated.

## Test plan

- [x] `/(^|/)figura-/` matches `mods/figura-0.1.5+1.21.jar`
- [x] `/(^|/)figura-/` does not match `mods/configurable-3.5.2+1.21.1-fabric.jar`
- [x] No other `globalExcludes` entry matches that configurable filename

### DIFF
--- a/files/modrinth-exclude-include.json
+++ b/files/modrinth-exclude-include.json
@@ -49,7 +49,7 @@
     "fast-ip-ping",
     "fastquit",
     "feytweaks",
-    "figura",
+    "figura-",
     "ForgeConfigScreens",
     "GeckoLibIrisCompat",
     "gpumemleakfix",
--- a/files/modrinth-exclude-include.json
+++ b/files/modrinth-exclude-include.json
@@ -49,7 +49,7 @@
     "fast-ip-ping",
     "fastquit",
     "feytweaks",
-    "figura-",
+    "/(^|/)figura-/",
     "ForgeConfigScreens",
     "GeckoLibIrisCompat",
     "gpumemleakfix",
@@ -110,9 +110,6 @@
     "cobbleverse": {
       "excludes": [
         "cloth-config"
-      ],
-      "forceIncludes": [
-        "configurable"
       ]
     }
   }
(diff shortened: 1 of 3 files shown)


files/modrinth-exclude-include.json
```diff
@@ -49,7 +49,7 @@
     "fast-ip-ping",
     "fastquit",
     "feytweaks",
-    "figura",
+    "figura-",
```
itzg: I know this technically works, but isn't in the spirit of the issue and introduction of regex matcher. 

Why didn't you use the regex you mentioned in the PR description?

> Slash-delimited regex (/(^|/)figura-/) is available in mc-image-helper 1.67+

lbellows: You're right — the substring was a workaround so it would still match on 1.66.1. That's not what #4234 asked for.

e617884d8d43a2fe47ccff9b74820461472c64a7 switches the exclude to `/(^|/)figura-/`, bumps the helper to 1.67.0 so the regex is actually evaluated, and drops the cobbleverse `configurable` force-include (it only existed to cover this collision). That also overlaps with #4243.

itzg (approved):
Thanks for the follow up

PR #4263, opened 2026-08-28
fix: remove managed files when MODRINTH_PROJECTS is empty

### LINKED ISSUE #3339: set MODRINTH_PROJECTS to an empty string can't removes all previously managed project-files
### Describe the problem

```
services:
  mc:
    image: itzg/minecraft-server
    environment:
      EULA: "true"
      VERSION: 1.21.4
      TYPE: "PAPER"
      MODRINTH_PROJECTS: "tabtps"
    ports:
      - "25565:25565"
    volumes:
      - ./data:/data
```

When `MODRINTH_PROJECTS` set to `""`, can't remove tabtps.

But `CURSEFORGE_FILES` can do it.

https://docker-minecraft-server.readthedocs.io/en/latest/mods-and-plugins/curseforge-files/
> References removed from the declaration are automatically cleaned up and setting CURSEFORGE_FILES to an empty string removes all previously managed project-files.

### Container definition

```
```

### Container logs

```
Paste logs here
```

lbellows:
Fixes #3339

Empty `MODRINTH_PROJECTS` now still invokes `mc-image-helper modrinth` so previously managed mods/plugins are removed, matching `CURSEFORGE_FILES`. Comment the variable out to skip processing.

## Test plan

- [x] `docker run -e MODRINTH_PROJECTS=tabtps` then `-e MODRINTH_PROJECTS=` removes `plugins/tabtps-*.jar`
- [x] Seeded `.modrinth-manifest.json` + dummy plugin, scripts mounted from this branch, empty `MODRINTH_PROJECTS` logs the new message and deletes both the jar and the manifest
- [x] `IMAGE_TO_TEST=itzg/minecraft-server:latest bash tests/setuponlytests/test.sh ./modrinth-empty-projects` — PASSED

### DIFF
--- a/scripts/start-setupModpack
+++ b/scripts/start-setupModpack
@@ -306,14 +306,22 @@ function handleModrinthProjects() {
       loader="${TYPE,,}"
     fi
 
-    mc-image-helper modrinth \
-      --output-directory=/data \
-      --world-directory="${LEVEL:-world}" \
-      --projects="${MODRINTH_PROJECTS}" \
-      --game-version="${VERSION}" \
-      --loader="$loader" \
-      --download-dependencies="$MODRINTH_DOWNLOAD_DEPENDENCIES" \
+    args=(
+      --output-directory=/data
+      --world-directory="${LEVEL:-world}"
+      --game-version="${VERSION}"
+      --loader="$loader"
+      --download-dependencies="$MODRINTH_DOWNLOAD_DEPENDENCIES"
       --allowed-version-type="$MODRINTH_PROJECTS_DEFAULT_VERSION_TYPE"
+    )
+    # An empty list still invokes the helper so previously managed files are
+    # removed, matching CURSEFORGE_FILES. Comment the variable out to skip.
+    if [[ -n "${MODRINTH_PROJECTS}" ]]; then
+      args+=(--projects="${MODRINTH_PROJECTS}")
+    else
+      log "MODRINTH_PROJECTS is empty; removing previously managed project files"
+    fi
+    mc-image-helper modrinth "${args[@]}"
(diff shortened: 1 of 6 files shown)


scripts/start-setupModpack
```diff
@@ -306,14 +306,22 @@ function handleModrinthProjects() {
       loader="${TYPE,,}"
     fi
 
-    mc-image-helper modrinth \
-      --output-directory=/data \
-      --world-directory="${LEVEL:-world}" \
-      --projects="${MODRINTH_PROJECTS}" \
```
itzg, line 312: Are you certain this didn't work before?

itzg (commented):
I'm not sure this PR was needed. I think I simply forgot to close the issue when mc-image-helper was enhanced.

scripts/start-setupModpack
lbellows: You're right — I rechecked and it already worked.

After #3703 this uses `[[ -v MODRINTH_PROJECTS ]]`, so an empty value still calls the helper with `--projects=""`. mc-image-helper has treated that as remove-all since [itzg/mc-image-helper#635](https://github.com/itzg/mc-image-helper/pull/635) (`removesAllWhenEmpty`). Omitting `--projects` vs passing `""` is equivalent, so this hunk isn't needed.

lbellows:
You're right, this isn't needed. I had gone off the still-open #3339 and missed that it was already fixed in [itzg/mc-image-helper#635](https://github.com/itzg/mc-image-helper/pull/635) and wired through here in #3703 (empty string removes previously managed files; commenting the variable out skips processing). Closing this PR and #3339.

